Issue #532, opened 2021-05-11
Add testing for TCE addons


flawedmatrix:
## Feature Request

Hi folks, I'd like to raise an issue here to get a sense of where your position is on adding integration testing for addons that TCE ships. I am interested in contributing some testing for the External DNS addon, but today there is no testing at all for the addons themselves so it would be difficult to start from a blank slate.

Specifically, is TCE planning on providing integration testing for addons?

If so, to what degree will testing be run (i.e. can it be run in CI)? Will these tests be run on TCE running on Docker or on a full TCE deploy on some owned infrastructure? What kind of configuration(s) should be tested for an addon?

PR #753, opened 2021-06-10
External-dns package Addon E2E tests

christianang:

This adds an end-to-end test for the External-DNS addon package. It installs the external-dns addon package on a cluster and tests the functionality of external-dns (e.g queries a record made by external-dns).

This will give us confidence that the external-dns addon we are shipping works and ensures as we make changes to the addon in the future that the addon will continue to work. Ideally, this will also be run in CI in the near future (probably introduced in a follow-up PR).

The test makes some assumptions about the environment:

* A TCE cluster exists and kubectl is targeting the cluster already.
  * For local development the test should ideally work with the `./test-automation` scripts
* The `external-dns.tce.vmware.com` package exists in the cluster. Either the user has to manually create the package or run `tanzu package repository install --default` beforehand.
* The cluster supports the `LoadBalancer` Service.
  *  Docker based clusters don't support the `LoadBalancer` Service without doing some work. One of the easiest ways to do this is to use [MetalLB](https://metallb.universe.tf/).
  * Ideally, there are some local scripts in TCE that can set this up for you so the user doesn't have to do this by hand. I'm thinking this could be done in the `test-automation` scripts either by default or through some optional flag (interested to hear opinions).

Side note: We noticed PR #735, which splits out separate go modules in TCE. We decided to follow this model assuming this is the way forward, but we can change this if necessary.

Related to #532 (does not deal with the issue completely)

1. Ran `./test-automation/run-run-tce-capd-standalone.sh`
2. Installed MetalLB to provide LoadBalancer Service support on this local cluster.
3. Ran the test using `ginkgo`.

This is the first PR to introduce an Addon test. We would welcome feedback on the patterns/utility functions/etc that would be used in other addon tests.

```release-note
NONE
```


navidshaikh (commented):
How about organizing the tests under `test/e2e/addons/packages/external-dns/` ?

Also, are we expecting each test to bring its own go submodule ?

jpmcb (commented):
I sort of see this PR as two separate additions:
- Adding a shared go module testing library at `test/pkg`
- Adds E2E tests for the external-dns package using the new testing library

I'd like to hear others feedback too on this file structure before continuing and maybe wait for #735 so that we can validate all is well with many go mods before diving in

jpmcb (commented):
This is looking great! But yes, we will need to block this on the top level go.mod being refactored to many go mods in https://github.com/vmware-tanzu/tce/pull/735

jpmcb (approved):
Great work on this and thanks much for the patience! :rocket:

mcwumbly (commented):
changes look good to me.

but since this isn't wired up to CI yet, I'd like to execute this once myself before I click "approve"

mcwumbly (approved):
lgtm

caveats: I had to use the latest release (`v.0.6.0`) instead of main to create the cluster.

I would like to see this hooked up into CI sooner than later as I'm concerned it'll just rot without being executed continuously.  We'll be forced to answer some of the questions I asked in line when we get this integrated with CI. Here's an issue to capture that request: https://github.com/vmware-tanzu/tce/issues/1136

That said, reading into the comments from @jpmcb, it sounds like there may be a desire to get this merged sooner than later and follow up with CI integration, so I'm assuming that is not considered a blocker.

PR #1119, opened 2021-07-19
Fix releasedAt in Package CR to be metav1.time format

shyaamsn:
Signed-off-by: Shyaam Nagarajan <nagarajans@vmware.com>

Package CR releasedAt is a metav1.Time type and not string type.


seemiller (approved):
LGTM

PR #1137, opened 2021-07-20
update ako-operator images.yml

iXinqi:
update ako-operator images.yml after ako-operator templates are changed


XudongLiuHarold (approved):
LGTM

PR #1142, opened 2021-07-20
Update ako-operator/values.yaml

iXinqi:
update ako-operator/values.yaml after fixing avi_labels and node_network_list bugs


XudongLiuHarold (approved):
LGTM

PR #1143, opened 2021-07-20
Go modules have their own makefiles

### LINKED ISSUE #1120: Consider using nested makefiles in multiple go modules
## Feature Request

We should consider using multiple, nested makefiles to support testing and building multiple go modules. This would be in support of making our top level makefile more clean. However this may introduce more complexity into an already complex build process. 

With multiple go modules being introduced, we've gone down the road of using the top level makefile as an uber coordinater for all build processes and repository operations. This has lead to some chunks of code in the makefile like this:

```
prep-build-cli:
	@cd ./cli/cmd/plugin/ && for plugin in *; do\
		printf "===> Preparing $${plugin}\n";\
		working_dir=`pwd`;\
		cd $${plugin};\
		$(GO) mod download;\
		cd $${working_dir};\
	done
```
https://github.com/vmware-tanzu/tce/blob/main/Makefile#L232-L239

and some github actions steps like this:

```
      - name: Get dependencies
        run: |
          for i in $(find . -type d); do
            if [[ -f "${i}/go.mod" ]]; then
              pushd $i
                go get -v -t -d ./...
                go mod edit --replace github.com/vmware-tanzu/tanzu-framework=$GITHUB_WORKSPACE/tanzu-framework
              popd
            fi
          done
```
Ref https://github.com/vmware-tanzu/tce/blob/main/.github/workflows/build-staging.yaml#L47-L56

This is fine, but could be more streamlined using recursive calls to make deeper in the file-tree. This has the advantage of making the larger TCE a more composable system of smaller pieces of software:

```
build-plugin:
        cd plugin && $(MAKE)
```

jpmcb:
This is a proof of concept PR to show that we can do recursive makefiles in multiple go modules without having to constantly update the linting and building targets of those modules. This should fix breaking people when a new module is introduced or modified (including plugin authors and package authors writing tests).

It should also be noted that we'd expect module / package / plugin authors to have a standard subset of those these operations that can always be accessed by the top level makefile.

Fixes: #1120

```
make lint-poc
```

This will recurse all go modules and calls the `do-lint` make target for those "sub-projects"

Note that this in it's current form is a POC. This PR can either transform into a broader pr to introduce makefiles to the many modules or we can open a new one

N/a


sabbey37 (approved):

ankeesler (changes requested):
looks like some copy pasta in the pinniped makefile?

ankeesler (approved):
pinniped changes lgtm

davidvonthenen (approved):
LGTM

christianang (approved):

stmcginnis (approved):
Tested locally, looks good to me.

PR #1158, opened 2021-07-22
Allow building single platforms via Makefile and speed up build-all

### LINKED ISSUE #1140: Allow building a limited set of architectures
## Bug Report

Right now, we build _all_ architectures when building. This can take a long time, and is usually not what a developer wants for quick iteration.

## Expected Behavior

Allow building of limited architectures when running a build command.

## Steps to Reproduce the Bug

`make build-all`

nrb:
This PR introduces the `ENVS` (combination of OS and processor architecture) make variable within our own Makefile and seeds it down through to the tanzu-framework Makefile invocation.

Additionally, this PR now only installs binaries when building for the host OS. `make build-all` will still work as before since all platforms are being built. For building and installing locally now, please use the `install-cli-plugins` target.

Fixes: #1140

Building just TCE plugins:

```
make build-all ENVS=linux-amd64
make build-all ENVS=windows-amd64
make build-all ENVS=darwin-amd64
make clean
make build-all
make build-all ENVS=linux-i386 -> should fail due to tanzu-framework error that checks platform info
make install-cli-plugins
```

Building everything, including the tanzu CLI from our `Makefile`.

```
make clean
make build-all
```

Using `time` on my local macOS machine, here's the difference in build times for `make build-all` after a `make clean`

| Change SHA | Time |
| --- | --- |
| this PR | 281.14s |
|  96e76e6d7c21950266b3ecaad3d8f45047e0b80 (head of main) | 545.88s |

```release-note
Add support for an `ENVS` makefile variable, in order to build for a single platform at a time.
When specifying an `ENVS` value that does not include the current host OS, neither the `tanzu` CLI nor plugins will be installed.

`make build-all` and `make release` both still build all platforms, but do *not* install the plugins anymore.
`make install-cli-plugins` will install plugins, but not build them.
`make build-install-plugins` will build, then install TCE-specific plugins, but not the `tanzu` CLI or tanzu-framework plugins.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ REQUIRED_BINARIES := imgpkg kbld ytt
 ROOT_DIR := $(shell git rev-parse --show-toplevel)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
+GOHOSTOS ?= $(shell go env GOHOSTOS)
+GOHOSTARCH ?= $(shell go env GOHOSTARCH)
+# Add supported OS-ARCHITECTURE combinations here
+ENVS := linux-amd64 windows-amd64 darwin-amd64
 
 TOOLS_DIR := $(ROOT_DIR)/hack/tools
 TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):
Checked out locally and ran `make build-cli ENVS=linux-amd64`. Only built linux-amd64 target, and was considerably faster than a full build of all platforms.

davidvonthenen (approved):

stmcginnis (approved):
Tested locally, all appears to be working as expected.

PR #1184, opened 2021-07-27
Contour fix job

jorgemoralespou:
Contour runs a Job that has a ttl set to 0, so that it gets immediately removed and kapp-controllers fails to validate existence of the job.
This PR removes the ttl and add a kapp annotation so that Job is recreated every time as instructed by @cppforlife 

No issue is opened for this.

Tested with `ytt -f config` in the bundle dir and `ytt -f config -v certificates.useCertManager=true` to test both cases defined in the modified overlay file. Tested both versions of the package.


skriss (approved):
LGTM, thanks @jorgemoralespou

PR #1227, opened 2021-08-03
Update external-dns test for new package plugin

tylerschultz:
The new packaging plugin has a different syntax and the CRDs have
changed, so the e2e test must be updated to be compatible.

- auto-detect package name, which can be different when the package is
  installed from different repositories for different editions of tanzu.
- renamed a few things for clarity
- added external-dns logs to diagnostic output when test fails
- add check for load balancer status to help debug cases where load
  balancer service is not available (e.g. metallb has not been installed
  yet on docker provider cluster).

Fixes: #

Running the tests against the main branch.
Manual validation of external-dns README instructions.

N/A

NONE

cc @mcwumbly


mcwumbly (approved):

jpmcb (approved):
This is looking great - let's get this in today but first, can we get a rebase ontop of `main` so we can capture the new automatic go linting that happens? That would be amazing!!

ShwethaKumbla (commented):
@tylerschultz @jpmcb 
We have verified these changes using our e2e automation framework and it is working fine.!

jpmcb (approved):

PR #1276, opened 2021-08-11
Implement PR method when bumping to next release

davidvonthenen:
This should allow us to create a PR to track changes for automation instead of checking directly into main. This can be super helpful for the following reasons:
- visibility into what gets checked into main
- when we need to rollback a tag/release for a pipeline failure, we just need to revert the commit

We should introduce this after we cut the v0.7.0 release.

NA

NA

NA

No

### DIFF
--- a/.github/workflows/release-fake.yaml
+++ b/.github/workflows/release-fake.yaml
@@ -26,8 +26,6 @@ jobs:
           git config --global url."https://git:$GITHUB_TOKEN@github.com".insteadOf "https://github.com"
 
       - name: Check out code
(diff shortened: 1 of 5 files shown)


carlisia (approved):
Hi,

When I run the `ensure-dependencies.sh`, this file gets downloaded (and not cleaned up): `make cut-release`. This happens with the `main` branch too. so I'm approving as is  since everything else lgtm. But leaving it unmerged for now.

seemiller (commented):
LGTM

PR #1345, opened 2021-08-18
For CI releases, we want to use a dedicated account token

davidvonthenen:
Apparently, in the `rc.4` release, we hit a GitHub API rate limit when attempting to automatically generate the release notes. Details here: https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps

The GH token is pretty much being used for every single GitHub action. We probably need to break that down to use several tokens and it's also going to be very important to stop running unnecessary CI jobs when they don't pertain to the PR.

```release-note
NONE
```

NA

NA

NA

### DIFF
--- a/.github/workflows/release-fake.yaml
+++ b/.github/workflows/release-fake.yaml
@@ -20,14 +20,14 @@ jobs:
 
       - name: Config credentials
         env:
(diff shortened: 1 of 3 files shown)


joshrosso (approved):

jpmcb (approved):
Sweet! LGTM

PR #1371, opened 2021-08-19
Remove adc from akoo templates

iXinqi:
Currently, we deploy default AKODeploymentConfig using add-on manager, which makes it uneditable.

To allow users to enable L7 or modify other AKO configurations in the management cluster or default select-all AKODeploymentConfig, we need to change it to use ClusterResourceSet to deploy default AKODeploymentConfig

Related PR: https://github.com/vmware-tanzu/tanzu-framework/pull/471


XudongLiuHarold (approved):
LGTM

PR #1391, opened 2021-08-20
Remove unnecessary GitHub Token from checkout step in GitHub Actions workflows

### LINKED ISSUE #1390: Remove unncessary GitHub Token from checkout step in GitHub Actions workflows
We use `GITHUB_TOKEN` in the checkout steps in many GitHub Action Workflows

https://github.com/vmware-tanzu/tce/blob/main/.github/workflows/release-ga.yaml#L29-L32

https://github.com/vmware-tanzu/tce/blob/main/.github/workflows/release-fake.yaml#L28-L31

https://github.com/vmware-tanzu/tce/blob/main/.github/workflows/release-nonga.yaml#L30-L33

But the token isn't really necessary and isn't used, so it's better to remove them as it's kind of like unused code / dead code

karuppiah7890:

The token isn't really necessary and isn't used, so it's better to remove
them as it's kind of like unused code / dead code

Fixes #1390

```release-note
NONE
```

Fixes: #1390 

None

None

### DIFF
--- a/.github/workflows/release-fake.yaml
+++ b/.github/workflows/release-fake.yaml
@@ -26,8 +26,6 @@ jobs:
           git config --global url."https://git:$GH_ACCESS_TOKEN@github.com".insteadOf "https://github.com"
 
       - name: Check out code
-        env:
-          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
         uses: actions/checkout@v1
 
       - name: Restore Go Cache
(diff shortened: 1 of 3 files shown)


navidshaikh (approved):
lgtm

stmcginnis (approved):
Appears to be fine.

davidvonthenen (commented):
If I remember correctly, I thought the repo had to be checked out using a token in order for the repo to have write access. We can always try and revert if it doesn't work. It would have been nice to keep the repo workflows static before renaming then merging this changes.

PR #1400, opened 2021-08-20
Remove unnecessary checks for github actions

davidvonthenen:

We shouldn't run certain long-running actions on certain GitHub actions depending on what is submitted. This helps with minimizing the number of GH API calls we make on these tokens! What has changed:
- We should only build staging binaries when a `go`, `sh`, or `GitHub action` file is modified
- We should perform a full `tarball` build on a PR request only when a `go`, `sh`, or `GitHub action` file is modified
- We should always build the entire `tarball` when merging to main

```release-note
NONE
```

NA

NA

No

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "**.go"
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):
Good call - looks good to me.

PR #1462, opened 2021-08-25
fix default adc will be written back by add-on manager issue

XudongLiuHarold:

Currently, we deploy default AKODeploymentConfig using add-on manager, which makes it uneditable.

To allow users to enable L7 or modify other AKO configurations in the management cluster or default select-all AKODeploymentConfig, we need to add rebase rule for kapp-controller

```release-note
NONE
```

Fixes: #


chenlin07 (approved):
lgtm

Issue #1403, opened 2021-08-23
Speedup builds in E2E tests


karuppiah7890:
## Feature Request

Currently we run `make release` and it builds all CLI plugins for all platforms, windows, linux, darwin. This takes a lot of time. Given that we do the E2E test only on GitHub Actions runner on Ubuntu runners, we just need Linux artifacts. Also, we may not need to build all CLI plugins, probably just `standalone-cluster` CLI plugin for TCE standalone cluster E2E tests, not conformance plugin, not any core / Tanzu Framework CLI plugins, unless it's absolutely necessary for standalone cluster CLI plugin to work. For management + workload cluster E2E tests, we do need `management-cluster` CLI plugin and `cluster` CLI plugin

Also, for testing `make release` itself for various platforms, we can have a separate test, where do we `make release` and try installation in different platforms and run some commands like `version`. Or better, we can later do complete E2E test for each platform on specific runners, if necessary. That's something to discuss.

For now, `make release` runs for a long time and it's too much, and we can do better and improve the speed of the E2E test. Also, this will reduce the cost for us - in terms of any GitHub Actions runner cost, as time reduces, and also time itself is a cost, less means less cost, our E2E tests are faster then :)

## Describe alternatives you've considered

Let the E2E test pipeline build for all platforms like how it's currently happening

## Additional context

Related issues - #1255 , #1140 

Related PR - #1158

PR #1472, opened 2021-08-26
Used GitHub Action Self-Hosted Runners

davidvonthenen:

This implements using self-hosted runners which are at least 2-3x faster than the "free" or provided runners by github. Runners are completely on-demand. They are dynamically created and regardless of whether the job (ie PR fails), clean up is always executed in order to prevent ec2 instance nodes to be orphaned and thus driving up costs.

The second feature in the this PR is that we also do a create/delete a standalone-cluster and verify that the operation has succeeded. This is actually a working functional e2e test!

This replaces the existing "PR check" which only did a `make release` and also replaces "verifications on main" which effectively did a `make check` and a `make release` with main after PRs were merged.

```release-note
Implements self-hosted GitHub runners in AWS and implements standalone cluster e2e that creates and then deletes a cluster.
```

https://github.com/vmware-tanzu/community-edition/issues/1403

This PR is the test!

NA


joshrosso (changes requested):
@dvonthenen can you re-work that commit message?

stmcginnis (approved):
All looks good to me, though might be good to get another set of eyes since there are a lot of things going on here. Or merge and we can just follow up if there are issues.

joshrosso (approved):

PR #1525, opened 2021-09-01
Housekeeping - Make Actions Names/Filenames Consistent

davidvonthenen:

This just renames the github actions (and a couple of their filenames) so that they are more consistent and it's easier to tell at a glance what these actions do.

The `update-packages.yaml` is a very very old gh action that is no longer being used.

```release-note
NONE
```

NA

NA

NA


stmcginnis (approved):
Nice!

PR #1562, opened 2021-09-03
Enhance long running github actions

davidvonthenen:

This PR updates all GitHub action that takes some time to run to use the new runner method.

Some significant changes:
- The AMI was also updated to allow starting/stopping the instance I'm using to build the AMI without triggering an attempt to register to GitHub.
- A runner timeout is at 5mins. If we can't setup the AWS runner in 5 minutes, we error out. The job (and only that job) can always be restarted should there be a networking or etc issue.

Other small changes:
- spacing and line endings to support formatting
- text changes for jobs in actions to be user friendly
- Remove doing a "check-main" when tagging for a release
- update certain actions to trigger only when the action itself is updated vs any action is updated

```release-note
Enhance long running github actions by enabling jobs to complete faster
```

NA

NA

NA

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -8,30 +8,56 @@ on:
       - "**.go"
       - "cli/cmd/plugin/**/go.mod"
       - "**.sh"
-      - ".github/**"
+      - ".github/workflows/build-staging.yaml"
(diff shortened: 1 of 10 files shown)


stmcginnis (approved):
Jobs are all happy, and this looks good to me.

PR #1563, opened 2021-09-03
Refactors makefile to use more explicit build and install targets

### LINKED ISSUE #1558: `make build-all` doesn't install the TCE plugins
## Bug Report

After freshly cloning the `community-edition` repo, I want to build and install everything to my system. In the past `make build-all` would do that but now, after running that make target and attempting to run `standalone-cluster`:

```
$ tanzu standalone-cluster -h
Error: unknown command "standalone-cluster" for "tanzu"
Run 'tanzu --help' for usage.

✖  unknown command "standalone-cluster" for "tanzu"
```

I see that standalone is getting built but I noticed in the Makefile that `build-all` calls `build-cli-plugins` but there is no install step.

## Expected Behavior

I'd expect `make build-all` to still install the plugins that get built

## Steps to Reproduce the Bug
Run `make build-all` and notice that the TCE plugins do not get installed:
```
$ tanzu -h
Tanzu CLI

Usage:
  tanzu [command]

Available command groups:

  Admin
    builder                 Build Tanzu components
    codegen                 Tanzu code generation tool
    test                    Test the CLI

  Run
    cluster                 Kubernetes cluster operations
    kubernetes-release      Kubernetes release operations
    management-cluster      Kubernetes management cluster operations
    package                 Tanzu package management

  System
    completion              Output shell completion code
    config                  Configuration for the CLI
    init                    Initialize the CLI
    login                   Login to the platform
    plugin                  Manage CLI plugins
    update                  Update the CLI
    version                 Version information

Flags:
  -h, --help   help for tanzu

Use "tanzu [command] --help" for more information about a command.

Not logged in
```

## Environment Details

N/a cc @nrb thoughts on this?

jpmcb:
Updates the make targets to be more explicit and based on feedback

- `make build-tce-cli-plugins` builds local plugins into the artifacts directory
- `make install-tce-cli-plugins` builds and installs local plugins from the artifacts directory
- `make build-all-tanzu-cli-plugins` builds the `tanzu` CLI and it's plugins. Also builds local TCE plugins
- `make install-all-tanzu-cli-plugins` builds installs the `tanzu` CLI and it's plugins. Also builds and installs local TCE plugins from the artifacts directory

This PR also ...
- Updates contributing docs to reflect the more explicit targets
- Adds `TCE_RELEASE_DIR` to makefile (and subsequent scripts) which is where tanzu framework is built locally. Should make updating this directory in the future (if need be) much easier
- Breaks out installing the tanzu CLI into it's own script so it can have it's own make target

```release-note
- TCE build and install make targets are more explicit
```

Fixes: #1558

Ran all commands including `make release`. Everything looks good! Also updated GitHub actions where I found `make build`

Open to any and all feedback. Also, if there's chunks missing that also need to be updated, please advise!

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -77,7 +77,7 @@ jobs:
           make get-deps
       - name: Build
         shell: bash
-        run: make build
+        run: make build-tce-cli-plugins
       - name: Upload Artifacts to Staging Bucket
         id: upload-artifacts-staging
         uses: google-github-actions/upload-cloud-storage@main
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,15 +21,15 @@ The CLI and some of its plugins live in different repositories. To build the CLI
 hosted in the TCE repository, run the following.
 
 ```shell
-make build-all
+make install-all-tanzu-cli-plugins
 ```
 
 After build and install, you'll see an output similar to the following.
 
 ```text
-[COMPLETE] installed plugins at /home/josh/.local/share/tanzu-cli/. These plugins will be automatically detected by tanzu CLI.
+[COMPLETE] built and installed plugins at /home/josh/.local/share/tanzu-cli/. These plugins will be automatically detected by tanzu CLI.
 
-[COMPLETE] installed tanzu CLI at /home/josh/bin/tanzu. Move this binary to a location in your path!
+[COMPLETE] built and installed tanzu CLI at /home/josh/bin/tanzu. Move this binary to a location in your path!
 ```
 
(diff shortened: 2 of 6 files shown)


joshrosso (changes requested):
I think we need consistency in the idea of `make build-install` (exists) vs `make-build-install-all` (does not exist)

Personally, I believe `make install` should force a build. Depending on the artifacts directory being in the correct state has a use case, but I believe it will cause more pain than it's worth. If folks want to install plugins, without compiling, they can do that via the `tanzu` CLI.

**recommendations:**

* remove `make build-install`
* change `make install` to always call `make build`

PR #1601, opened 2021-09-07
Allow install script as root with override for self-hosted runner

davidvonthenen:

This fixes an issue with the GitHub self-hosted runner where it was being executed by the default AMI user, but now for some reason has changed to root.

```release-note
NONE
```

NA

Verified non-root user produces error. Verified that providing the override allows as root allows install.

NA


jpmcb (approved):

stmcginnis (approved):
Seems reasonable. Would be great if we could figure out why the runner is now running as root, but that might not be worth the effort to track down at this point.

PR #1604, opened 2021-09-07
Improve CI response time by detecting non-responsive runners

davidvonthenen:

Once in a blue moon when you deploy a self-hosted runner, the runner fails to initialize... this is easily detectable because the node for some reason doesn't report any kind of status for a good amount of time. This usually burns the entire 5min timeout for initialization which leads to failure (usually when the agent is registered, the status is "offline" until it's not). This allows to detect for early non-responsive or agents that failed to register which is detected in 1/3 of the total timeout time and then retries up to a maximum of 3 times by creating a new ec2 instance and starting the process over again automatically.

This happens roughly about 1-3 times a day based on the large number of jobs that get kicked off.

```release-note
NONE
```

NA

NA

NA

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -21,13 +21,13 @@ jobs:
       - name: Download Dependencies
         shell: bash
         run: |
-          curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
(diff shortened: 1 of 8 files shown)


jpmcb (dismissed):
A few commented out blocks - but this is sweet. Very cool enhancement

stmcginnis (approved):

PR #1636, opened 2021-09-09
Doc dpp linear flow

### LINKED ISSUE #1118: User request for a documentation structure change to support a more linear install workflow, with added content
Entering this on behalf of a design partner program partiucipant

The “Getting Started” pages are much better to follow rather than the “Installation” section which is broken into sections
.
would like to see the documentation follow more of a linear format rather than jumping to a link that then requirees rembering a return point in the workflow
.
- [x] 1. Would like to see all prerequisites called out up front so that resolving these mid-process does not disrupt workflow.

- IP Addresses
- Subnets
- DHCP
- routing, connectivity, DNS if any needed
- CPU, memory, storage min and recommended
- Downloads needed (swith links so all can be done in advance)

- [x]  2. The documentation is unclear regarding the system performing the install requires a GUI/Browser.  When installing from a system with no GUI, it still wants to have a browser for 
installation purposes.  (KOR: This can be added as a prereq?) #1482

- [x] 3. The documentation instructs the user to download an image OVA, but it was not clear regarding 
where to download the image. (KOR: @joshrosso is following up with PM) #1483

- [x] 4. The “Bootstrap machine” is referred to different names throughout the document. (KOR: This will be addressed in: #1288)

- [ ] 5. Having load balancer and DHCP requirements does not make the installation simple. (KOR: @joshrosso would need to comment here)

- [ ] 6. Having some design architecture design drawings would be helpful to help conceptualize the 
infrastructure (in docs) What is a management cluster?, workload cluster?, bootstrap machine and relationship to clusters? Comparison of standalone cluster to alternative? (KOR:This has been addressed in: [Architecture doc](https://quirky-franklin-8969be.netlify.app/docs/latest/architecture/))  #1486 

- [ ] 7. It would be helpful to provide customers with a spreadsheet to help get started with all the necessary 
prerequisites (IP Addresses, DNS, load balancer, etc.). This is like how vBlock used to onboard new 
customers. (KOR: this is the same as what is being asked for in 1 above?)

- [ ] 8. We should include the Docker installation process in our documentation. (useful for Kubernetes/container newbies) (KOR: will do) #1485 
- Including installation validation

### LINKED ISSUE #1482: Doc feedback from Steve Wong
## Feedback

Feedback from @cantbewong 

- [ ] change variable to exclude angle brackets and use convention similar to MY-MGMT-CLUSTER without angle brackets as this will copy over ok without need for edit if you are doing a quick test  - creating new issue for this and adding to ice box for now #1637 

- [x] Install 'to' is more accurate than Install 'on' 

- [x] Formatting suggestion: In preparing to install topic, change Install, Prepare, Create to bold formatting - issue 1 

- [x] Remove 'infrastructure provider' - it's an internal wordy term our users might not understand, use a sentence like, 'there are 4 supported target installation for deploying clusters' 'destination platform of your choice' 

- [x] split CLI prereqs by OS

kcoriordan:

- Restructures the main installation doc so that there is one flow per infrastructure provider - comes from DPP feedback #1118 
- Other minor style updates, #1482
- changed the infrastructure provider prereq formatting to check boxes as per feedback from @thesteve0  

```
Documentation structure is changed so that there is now one section for each infrastructure provider.
```

Fixes: #1482
Fixes: #1118

Most updates are to the toc, plus added 1 new intro file per provider
I messed up the formatting for the checkboxes in the prereq files - I'll fix that in another PR

### DIFF
--- a/docs/site/content/docs/assets/step1.md
+++ b/docs/site/content/docs/assets/step1.md
@@ -24,7 +24,9 @@ This topic describes how to use the Tanzu installer interface to deploy a manage
 
 ## Procedure
 
-Start the Installer. On the machine on which you downloaded and installed the Tanzu CLI, run the `tanzu management-cluster create` command with the `--ui` option:
+### Start the Installer in your Browser
+
+On the machine on which you downloaded and installed the Tanzu CLI, run the `tanzu management-cluster create` command with the `--ui` option:
 
    ```sh
    tanzu management-cluster create --ui
--- a/docs/site/content/docs/assets/support-matrix.md
+++ b/docs/site/content/docs/assets/support-matrix.md
@@ -1,0 +1,7 @@
+The following support matrix provides a summary of the target platforms that are supported for each operating system:
+||Linux|Windows|Mac|
+|:------------------------ |:------------------------ |:------------------------ |:------------------------ |
+|**Amazon EC2**|yes|yes|yes|
+|**Docker**|yes|no|yes|
+|**Microsoft Azure**|yes|yes|yes|
+|**vSphere**|yes|yes|yes||||
--- a/docs/site/content/docs/latest/aws-intro.md
+++ b/docs/site/content/docs/latest/aws-intro.md
@@ -1,0 +1,17 @@
+# Amazon EC2
+
+Complete the following steps for a full Tanzu Community Edition implementation to Amazon EC2:
+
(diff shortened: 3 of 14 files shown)


seemiller (approved):
@kcoriordan The changes to the TOC should be fine as a far as the automation code goes. We'll only have a problem if structural tags (title, subfolderitems, page, etc) change. Looks like this is just content, so all good.

Issue #1649, opened 2021-09-10
Multiple packages are failing to install due to container image sha being non-existent


cormachogan:
## Bug Report

Attempting to install Prometheus after deploying TKG workload cluster onto vSphere:

```
% tanzu package available list prometheus.community.tanzu.vmware.com
/ Retrieving package versions for prometheus.community.tanzu.vmware.com...
  NAME                                   VERSION  RELEASED-AT
  prometheus.community.tanzu.vmware.com  2.27.0


 % tanzu package  install prometheus -p prometheus.community.tanzu.vmware.com -v 2.27.0
| Installing package 'prometheus.community.tanzu.vmware.com'
| Getting namespace 'default'
/ Getting package metadata for 'prometheus.community.tanzu.vmware.com'
| Creating service account 'prometheus-default-sa'
| Creating cluster admin role 'prometheus-default-cluster-role'
| Creating cluster role binding 'prometheus-default-cluster-rolebinding'
\ Creating package resource
\ Package install status: Reconciling

Please consider using 'tanzu package installed update' to update the installed package with correct settings



Error: package reconciliation failed: Error: Syncing directory '0': Syncing directory '.' with imgpkgBundle contents: Imgpkg: exit status 1 (stderr: Error: Checking if image is bundle: Collecting images: Working with projects.registry.vmware.com/tce/prometheus@sha256:39324692165b84a1657ac6e422a84f8b50b278579b3ebd526abdb7552945afc2: GET https://projects.registry.vmware.com/v2/tce/prometheus/manifests/sha256:39324692165b84a1657ac6e422a84f8b50b278579b3ebd526abdb7552945afc2: NOT_FOUND: artifact tce/prometheus@sha256:39324692165b84a1657ac6e422a84f8b50b278579b3ebd526abdb7552945afc2 not found
)

Usage:
  tanzu package install INSTALLED_PACKAGE_NAME --package-name PACKAGE_NAME --version VERSION [flags]

Examples:

    # Install package contour with installed package name as 'contour-pkg' with specified version and without waiting for package reconciliation to complete
    tanzu package install contour-pkg --package-name contour.tanzu.vmware.com --namespace test-ns --version 1.15.1-tkg.1-vmware1 --wait=false

    # Install package contour with kubeconfig flag and waiting for package reconcilaition to complete
[...]

PR #1693, opened 2021-09-13
Use `kbld` image overrides instead of `ytt` overlays for VMware built images

jpmcb:
For VMware built images that need to replace upstream images, uses a `kbld` override instead of overlays. This prevents exposing users to the "image" overlay data value which should not be configurable in this case and should not removed.

```release-note
- Image replaces for vmware built images is not configurable
```

Related: #1649
Several image shas got blown away in the last update so we are getting this in with correct image replaces before bumping a new image repo

Ran 
```
kbld -f bundle --image-lock-output bundle/.imgpkg/images.yml | rg image
```
on each to confirm that the images are getting correctly updated in the bundle. Also note the image references in each of the `image.yml` lock files

N/a cc @seemiller


jpmcb (commented):
I've verified that all the image bundles referenced in the various `package.yml` files are pushed and the shas correct in Harbor

seemiller (approved):
LGTM

Issue #1660, opened 2021-09-11
Move to TCE to a TCE-specific plugin bucket


joshrosso:
## Feature Request

TCE should move to its own plugin bucket, such that we can:

1. Gate what tanzu-framework plugins appear in updates for TCE users
1. Ensure CLI init/update doesn't pull down the tanzu-framework plugins that were compiled for TCE
    - This is temporary until we solve https://github.com/vmware-tanzu/tanzu-framework/issues/569

This would have the following changes to our build automation.

1. We'd publish tanzu-framework built plugins in our bucket
1. We'd ensure that only the TCE plugin repo is registered with the tanzu-cli

## Describe alternatives you've considered

None

## Additional context

None

PR #1745, opened 2021-09-14
Replace TCE buckets with production

davidvonthenen:

Just like it says... stop using (but keep updating the old bucket) but reference the new bucket with a no-op/empty for plugin versions. This effectively pins the version from being upgraded.

There is a lot to this PR because of some workarounds that needed to be done.
Changes include:
- needed to make 2 new buckets (a TCE core and core-admin bucket)
- pass in bucket override to TF and repo plugins
- restructure the output of the TF plugins artifacts directory so we can just delete the binaries, put in an empty file, and then just upload
- restructure the install.sh/bat because of bucket issues
- modify the github actions to upload to all these new buckets
- finally moved some scripts around to start to clean up the "hack" folder

```release-note
Pin the TCE releases by replacing with a new production GCP bucket
```

https://github.com/vmware-tanzu/community-edition/issues/1660

NA

NA


joshrosso (changes requested):
Let's wait for `v0.9.0` to get this PR in.

PR #1792, opened 2021-09-17
Support testing external-dns pkg in non-global namespace

christianang:

The package repository may not be installed in the global namespace which doesn't work for the External-dns e2e tests. This change introduces a new environment variable to set the `PACKAGE_INSTALL_NAMESPACE` that lets you set the namespace the `tanzu package install` command runs with.

By default, when `PACKAGE_INSTALL_NAMESPACE` is not set we default to assuming the package is in the global namespace and we can create/test the package install in our own namespace.
When `PACKAGE_INSTALL_NAMESPACE` is set we assume the package is in that namespace and we should *also* install the package into the same namespace.

```release-note
Add PACKAGE_INSTALL_NAMESPACE environment variable to External-dns E2E test.
```

Fixes: n/a

Ran the test with both the package repository installed in the global namespace and in the `default` namespace.

### DIFF
--- a/addons/packages/external-dns/0.8.0/test/README.md
+++ b/addons/packages/external-dns/0.8.0/test/README.md
@@ -10,8 +10,8 @@ To run the `external-dns` end-to-end tests you need:
 
 * A TCE cluster and the cluster needs to be the current-context.
 * The cluster supports Service type `LoadBalancer`.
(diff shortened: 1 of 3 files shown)


mcwumbly (approved):
I just have a clarifying question about the fixture file added in this commit which I asked inline.

ShwethaKumbla (approved):

Issue #1759, opened 2021-09-15
Reenable mdlint checks on web content


stmcginnis:
## Feature Request

We have mdlint checking for our markdown files to make sure they are properly formatted and do not have any markdown syntax issues. We had disabled this on `doc/site` while things were being worked on. Now that things have settled a bit, those linting checks need to be turned back on and issues addressed.

## Describe alternatives you've considered

Anarchy

## Additional context

Markdown is being done for nearly all other markdown files in the repo, other than the ones that are used to publish our web site. We need to be consistent and make sure the content is well formatted.

PR #1816, opened 2021-09-17
Add links to mdlint rules documentation

stmcginnis:

When errors are encountered in the mdlint jobs, if you aren't very
familiar with markdown it can be confusing understanding what the error
is and how you are supposed to fix it. This adds a couple breadcrumbs to
the make target and mdlinting script to point to documentation that goes
over what each error is, with examples of formatting that can cause them
and suggestions on how to fix them.

```release-note
NONE
```

Related: #1759

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,8 @@ lint: tools get-deps
 	done
 
 mdlint:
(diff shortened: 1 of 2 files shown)


jpmcb (approved):
👍 🍞

PR #1832, opened 2021-09-19
Remove duplicate TOCs

### LINKED ISSUE #1761: [DOCS] Remove extra TOC from pages
## Description

Some pages duplicate the table of contents with a section at the top of the content, usually stating something like "This topic describes:". These are redundant to the TOC that is already built in to the theme and adds more distraction from the actual content of the page, so they should be removed.

## Location

Various - one example is `docs/latest/workload-clusters/`

joshrosso:

This removes the TOC's that were put at the header of some docs. This is
not needed because the TOC is generated in the right nav bar.

```release-note
NONE
```

* Fixes: https://github.com/vmware-tanzu/community-edition/issues/1761

Run hugo serve

I clicked through the website and these were the only ones I could find.

### DIFF
--- a/docs/site/content/docs/latest/ref-azure.md
+++ b/docs/site/content/docs/latest/ref-azure.md
@@ -1,10 +1,7 @@
 # Reference for Azure account
 
-If you encounter issues deploying a cluster to Azure, review the following troubleshooting and reference content:
-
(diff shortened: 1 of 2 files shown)


stmcginnis (commented):
I see one more in `docs/site/content/docs/latest/vsphere.md`. Mind picking that one up in this PR too?

kcoriordan (approved):

PR #1847, opened 2021-09-20
doc - adds supported providers to Contour readmes

### LINKED ISSUE #1801: [DOCS] Document which packages work on Docker provider
## Description

There's some plugins that will not work as they are on CAPD clusters, and configuration might be difficult (or not possible), others don't even make sense. We should add some notes to this.
An example are:

- [x ] - cert-manager can not integrate with Let's Encrypt on CAPD which is the documented approach
- [x ] - external-dns does not make sense on CAPD as there's no LoadBalancer
- [x ] - contour currently does not allow to create ClusterIP services which are the most meaningful for CAPD
- [ ] - harbor install requires to restart docker which makes you lose the cluster (https://github.com/vmware-tanzu/community-edition/issues/1800)
- [ ] - probably more

kcoriordan:

Adds table of supported providers to exteranal-dns readme as suggested here:
#1801

```release-note

```

Fixes: #1801

### DIFF
--- a/addons/packages/contour/1.15.1/README.md
+++ b/addons/packages/contour/1.15.1/README.md
@@ -2,6 +2,14 @@
 
 This package provides an ingress controller using [Contour](https://projectcontour.io/).
 
+## Supported Providers
+
+The following tables shows the providers this package can work with.
+
+| AWS  |  Azure  | vSphere  | Docker |
+|:---:|:---:|:---:|:---:|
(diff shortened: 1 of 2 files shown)


skriss (approved):
It's not entirely accurate to say this package doesn't work on Docker, but we don't have the configurations documented and it won't work as expected OOTB so I'm OK documenting it like this for now until we can address that.

PR #1855, opened 2021-09-20
Harden GitHub workflows for `pull_request` targets

### LINKED ISSUE #1023: Proposal: enable "Send secrets to workflows from fork pull requests" feature
### TLDR

Enable the "Send secrets to workflows from fork pull requests" feature for workflows originating from forks to enable the use of `GH_ACCESS_TOKEN`until TCE and other Tanzu github repos are public. Then, use the read-only `GITHUB_TOKEN` for workflows from forks.

### Proposal

We want to enable the running of CI / CD from forks to better serve contributors and package authors. This way, they can make PRs from their forks and see any MD or shell linting problems along with build errors. To do this, we enabled the "Run workflows from fork pull requests" feature of GitHub enterprise. But, no secrets are sent to workflows based from forks (for security reasons as [documented here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories)) and the default read-only `GITHUB_TOKEN` does not have enough permissions to complete workflows and pull from the various tanzu repos. Ex: #1021. But, we can enable the "Send secrets to workflows from fork pull requests" feature to send the higher permission `GH_ACCESS_TOKEN` secret. This _should_ enable the secret access token to be sent and allow those workflows to run appropriately.

### Future plan

So, in the future, to prevent any security risk, we will need to:

- [ ] Make tce and other tanzu github repos public
- [ ] Disable the "Send secrets to workflows from fork pull requests" feature
- [x] ~~Remove the `GH_ACCESS_TOKEN`~~
- [x] Replace workflows using `GH_ACCESS_TOKEN` and the `pull_request` target to use the default `GITHUB_TOKEN`

jpmcb:
- Don't populate the environment with secrets which could potentially be printed to the environment when not needed [0]
- Use the autogenerated `GITHUB_TOKEN` which is read only and populated by GitHub workflows on each run [1]
- Don't run workflows from our self hosted runners on Pull Requests as it's possible for attackers to introduce malicious code from opening a pull request [2]. Further, when `community-edition` is public, secrets being sent to workflows that originate from for pull requests are not enabled. So the workflows that broker a EC2 runner won't work as those secrets will be missing []. This changes those workflows to be post merge on the `main` branch

[0] [Limitations of secrets](https://docs.github.com/en/actions/reference/encrypted-secrets#limits-for-secrets)
[1] [About the `GITHUB_TOKEN` secret](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret)
[2] [Self hosted runner security with public repositories](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories)
[3] [GitHub Actions improvement for fork and pull req workflows](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/)

```release-note
Harden pull requests workflow runs
```

Fixes: #1023

This PR should be ok to merge and try on main as it stands but may need a followup for going public

N/a

### DIFF
--- a/.github/workflows/check-actionlint.yaml
+++ b/.github/workflows/check-actionlint.yaml
@@ -22,10 +22,8 @@ jobs:
         id: go
 
       - name: Config credentials
-        env:
-          GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
         run: |
-          git config --global url."https://git:$GH_ACCESS_TOKEN@github.com".insteadOf "https://github.com"
+          git config --global url."https://git:${{ secrets.GITHUB_TOKEN }}@github.com".insteadOf "https://github.com"
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v1
--- a/.github/workflows/check-imagelint.yaml
+++ b/.github/workflows/check-imagelint.yaml
@@ -23,10 +23,8 @@ jobs:
         id: go
 
       - name: Config credentials
-        env:
-          GITHUB_TOKEN: ${{ secrets.GH_CHECKS_ACCESS_TOKEN }}
         run: |
-          git config --global url."https://git:$GITHUB_TOKEN@github.com".insteadOf "https://github.com"
+          git config --global url."https://git:${{ secrets.GITHUB_TOKEN }}@github.com".insteadOf "https://github.com"
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v1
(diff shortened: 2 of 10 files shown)


joshrosso (changes requested):
Let's discuss today when we want to cut over

davidvonthenen (approved):
LGTM. Merge when the decision comes down!

PR #1852, opened 2021-09-20
Enables installing TCE plugins with experimental versions

jpmcb:
## What this PR does / why we need it
Fixes running `make install-all-tce-plugins` when the Tanzu CLI is _not_ already on the system. Before this PR, we'd see the following error:
```
✔  successfully built local repository 
make[1]: Leaving directory '/home/vonthd/go/src/github.com/vmware-tanzu/community-edition/hack/builder'
make[1]: Entering directory '/home/vonthd/go/src/github.com/vmware-tanzu/community-edition/hack/builder'
TANZU_CLI_NO_INIT=true GOPRIVATE="github.com/vmware-tanzu/*" go run -ldflags "-s -w -X "github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildDate=$(date -u +"%Y-%m-%d")" -X "github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildSHA=$(git rev-parse --short HEAD)" -X "github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli.BuildVersion=$(git describe --tags --abbrev=0)" -X 'main.BuildEdition=tce-standalone' -X 'github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/buildinfo.IsOfficialBuild=""' -X "github.com/vmware-tanzu/sonobuoy/pkg/buildinfo.Version=v0.53.2"" github.com/vmware-tanzu/tanzu-framework/cmd/cli/tanzu \
	plugin install all --local ../.././artifacts
Error: version cannot be empty for plugin "conformance"
```
Which was due to the missing call to Tanzu Framework's `make set-unstable-versions` call. This now enables maintainers / users to run `make install-all-tce-plugins` without error

## Details for the Release Notes (PLEASE PROVIDE)
```release-note
Running "make install-all-tce-plugins" works with experimental versions of TCE plugins
```

## Which issue(s) this PR fixes
Fixes: #1851

## Describe testing done for PR
`make install-all-tce-plugins` now passes!

## Special notes for your reviewer
N/a

PR #1858, opened 2021-09-20
Enable TCE buckets and update plugins to v0.2.1 of tanzu-framework

### LINKED ISSUE #1660: Move to TCE to a TCE-specific plugin bucket

### LINKED ISSUE #1814: Update `conformance` plugin to use tanzu-framework v0.2.1
## Feature Request

The `conformance` plugin uses a very old version of tanzu-framework which causes issues with properly configuring alternate buckets as a source. This should be fixed in order to avoid overwriting user's configuration with incorrect data.

As of this writing, PR #1745 works around this by installing plugins and specifying buckets in a very specific order.
This will work, but is definitely a work around.

Similarly, we've looked into updating how Chocolatey and Homebrew install plugins and want to make sure they match.

Fixing the version of tanzu-framework would mean that we don't have to do much special logic in the packaging scripts, however.

### LINKED ISSUE #1815: Update `diagnostics` plugin to use tanzu-framework v0.2.1 #1814
## Feature Request

The `diagnostics` plugin uses a very old version of tanzu-framework which causes issues with properly configuring alternate buckets as a source. This should be fixed in order to avoid overwriting user's configuration with incorrect data.

As of this writing, PR #1745 works around this by installing plugins and specifying buckets in a very specific order.
This will work, but is definitely a work around.

Similarly, we've looked into updating how Chocolatey and Homebrew install plugins and want to make sure they match.

Fixing the version of tanzu-framework would mean that we don't have to do much special logic in the packaging scripts, however.

## Additional context

We should try this out in the `conformance` plugin first (#1814) so there's a path for an external contributor to follow (cc @vladimirvivien)

nrb:

This is a combination of #1852, #1745, and #1848 since the changes are so inter-connected. Gathering them in one place makes it easier to reason about and test.

```release-note

```

Fixes: #1814 
Fixes: #1815 
Fixes: #1660

`make release`
`make install-all-tanzu-cli-plugins`

Do not merge until v0.8.0 is released.


joshrosso (changes requested):
extremely minor nit requested.

Otherwise lgtm, let's make this happen

joshrosso (changes requested):
@dvonthenen to run build/release simulation and verify

jpmcb (changes requested):

davidvonthenen (dismissed):

davidvonthenen (approved):
LGTM

joshrosso (approved):

jpmcb (approved):

PR #1901, opened 2021-09-22
Add dynamic link functionality for versions on the landing page

jonasrosland:

Add a new parameter setting in config.yaml for `release_latest`, which is then picked up by the new shortcode. Also changed the way our `include` shortcode worked, as `readFile` doesn't handle handle shortcodes within included files.

```release-note
NONE
```

Fixes: #

### DIFF
--- a/docs/site/config.yaml
+++ b/docs/site/config.yaml
@@ -50,6 +50,10 @@ params:
   docs_latest: latest
   docs_versions:
     - latest
+  release_latest: v0.8.0
+  release_versions:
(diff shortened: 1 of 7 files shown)


stmcginnis (dismissed):
This looks good to me. I think it would be good to have David take a look from a release automation perspective. But even if we don't automate this for now, I think it would be good to add this so at least manual updates would be easier.

davidvonthenen (approved):

PR #1936, opened 2021-09-23
Doc improvements for Work with Packages doc

### LINKED ISSUE #1775: [DOCS] Suggestions to the "work with package" doc
## Description

Recommendation 1:

Reading through the package docs , my eyes are immediately drawn to the package installation diagram even though we have only just started to read about repositories. I think this is a little overwhelming and should be placed further down in the doc once readers have had a chance to digest some of the other package information.

My first recommendations would be:

(a) remove the package installation diagram from the repositories section
(b) move the package installation diagram from the start of the Packages section to the start of the Install a Package section.

Recommendation 2:

Reword: "This is the namespace packages will be discoverable within." -> "This is the namespace where package repositories will be discovered". 

Reword: "It does not define the target namespace software is eventually run within" -> "This does not define the target namespace where packages will eventually run"

Recommendation 3:

Reword: "A package may have its own unique installation steps or requirements" -> A package may have its own unique installation steps or requirements and may have dependencies on other software, e.g. Contour has a dependency on Certificate Manager" << _should we call out this explicitly? I feel it would be useful to do that_

Recommendation 4: 

Regarding the following statement: "This does not determine which namespace(s) the software will run in.", would it be worth discussing the relationship between repository, installed packages and running software from a namespaces perspective, and have a dedicated diagram just for that. I feel this is the area that most readers will get tripped up.

Recommendation 5:

Regarding the following statement - "The Packages, or software available for install, are always available in the same namespace as the PacakgeRepository", won't this always be the default namespace unless the customer has specifically asked it to be placed somewhere else? If so, is it worth stating that?

## Location

https://quirky-franklin-8969be.netlify.app/docs/latest/package-management/#work-with-packages

kcoriordan:

As per recommendations from @cormachogan in #1775 
1. Removed duplicate image - line 15
2. Reworded
3. Reworded
4. Opened a nice to have issue and added to ice box as not in scope for launch  #1937
5. Reworded
Changed heading titles to fix lint issue

```NONE

```

Fixes: #1775

### DIFF
--- a/docs/site/content/docs/latest/package-management.md
+++ b/docs/site/content/docs/latest/package-management.md
@@ -145,9 +145,8 @@ To install a package, the `${PACKAGE_FQN}` and `${PACKAGE_VERSION}` is required.
 
 ### Installing a Package
 
-A package may have its own unique installation steps or requirements. Before
-installing a package, be sure to review its documentation. Documentation for
-each package can be found in the left navigation (`Packages > ${PACKAGE_NAME}`) of this site.
+A package may have its own unique installation steps or requirements, and may have dependencies on
+other software, for example, Contour has a dependency on Cert Manager. Before installing a package, be sure to review its documentation. Documentation for each package can be found in the left navigation (`Packages > ${PACKAGE_NAME}`) of this site.
 
 To install a package, run:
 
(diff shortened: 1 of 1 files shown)


joshrosso (changes requested):
Thanks for jumping on these improvements. I've left some comments throughout.

joshrosso (approved):
Thanks @kcoriordan !

PR #1958, opened 2021-09-24
Add tagger hack to prevent gc from breaking packages

### LINKED ISSUE #1867: Remediate existing imgpkg bundles with arbitrary tag
need arbitrary tag representative of SHA for bundles

joshrosso:

Packages are referenced from repositories via their digest. This means
that if a tag on a configuration bundle mutates, the previous artifact
may no longer have a tag associated with it. These are considered
"dangling" images and will be collected (deleted) when Harbor runs its
garbage collection cycle.

The issue with this is, if a package repository references that
no longer existent digest, it will fail to retrieve the bundle, breaking
the package.

There are two changes that must take place:

1. imgpkg should set a default (additional) tag on bundles it pushes[0]
2. we must scan our package bundles and add an arbitrary and unique tag to ensure
   that they are never garbage collected.

This commit solves 2. It scans all the projects that represent package
bundles in TCE. It then resolves the artifacts for each of those
bundles. For each artifact it looks at the tags and sees if it has the
expected, unique, tag. The tag we've got with is the first 10 characters
of the digest's sha256 value. As an example:

sha256:0c3d0f33c171e437268e57bdbe0d83feb1606362d8235f1b656556da8c944e18

would have a required tag value of c3d0f33c17.

If missing, this tag is added to the artifact. If it exists, no
operation is triggered.

No release notes on this PR, it's an internal detail.

```release-note
NONE
```

* Fixes: https://github.com/vmware-tanzu/community-edition/issues/1867

tagger

See the above


seemiller (approved):
My only change would be to shorten the tag to the first 8 characters of the SHA instead of 10. Git commands will shorten SHAs to the first 8 characters - guess I'm accustomed to that standard.

Otherwise, LGTM

PR #1980, opened 2021-09-27
Fluent bit defaults

LukeWinikates:

This PR improves the usability of the Fluent-bit package by default

It adds several Parsers from a standard list published by Fluent
It removes a Filter definition that was previously used in a TKG component, but isn't relevant for all Fluent-bit users

```release-note
Include additional Fluent-bit Parsers
```


akodali18 (approved):
Thanks for the changes, lgtm.

PR #2000, opened 2021-09-28
Doc-Draft- Adding OVA links to vSphere docs

kcoriordan:

```release-note

```

Fixes: #


joshrosso (changes requested):
@kcoriordan I think the previous wording, screenshots, etc were pretty clean.

I'd recommend we re-do this PR to only update the links and not these other changes.

If the image needs to change for the customer connect portal, we can easily update that.

PR #2083, opened 2021-10-01
Fix for exit non failure on Fail(Alpine Image) status

JitenPalaparthi:
This fixes issue as on fail instead of exit(1) linter is exit(0)
Add new item in matchPattern key in configuration file

Signed-off-by: Jiten Palaparthi <JPalaparthi@vmware.com>

This fixes issue as on fail instead of exit(1) linter is exit(0)
Add new item in matchPattern key in configuration file

```release-note

```

Fixes: # N/A

Tested on local machine


jpmcb (approved):
LGTM - but I think we should (at some point in the future) consider re-writting this to not use global variables that get mutated throughout the run.

PR #2085, opened 2021-10-01
Use a semantic version instead of stable tag for installing pkg repo …

### LINKED ISSUE #2084: Use semver tag instead of `stable` when referencing main package repo
## Description

Due to garbage collection and caching issues with our OCI registry, it is best to not re-use tags for our main package repository.

For now, we should reference a semantic version.

seemiller:

Due to garbage collection and caching issues with our OCI Registry, it is not advisable to (re)use a tag like `stable` or `latest` in installation instructions for the package repository. Instead, we should reference a semantic version of the repo.

```release-note
Use a semantic version instead of stable in package repo installation docs
```

Fixes: #2084


karuppiah7890 (approved):
LGTM!

joshrosso (approved):

PR #2146, opened 2021-10-05
Re-organize to highlight that it may take some time to reconcile

mikegehard:
The part about reconciliation was hidden below what seemed to be an unrelated piece of info about a system package repo.

This commit moves that system repo info up to where it talks about installing repos in namespaces and highlights the fact that it takes time to reconcile

```release-note
NONE
```

Doc change


stmcginnis (approved):
This makes sense to me - this should help make it a little more visible and a little more likely that folks will actually see it. Thanks!

PR #2148, opened 2021-10-05
Use docs template

mikegehard:
Changes to us the docs GH issue template for better docs bug/feature reporting.

```release-note
NONE
```

Viewed generated docs in browser.


stmcginnis:
@kcoriordan can you take a look. The change looks good to me, but I'm not sure why `issueQuery` was being used in the first place.

docs/site/themes/template/layouts/partials/docs-right-bar.html
```diff
@@ -5,7 +5,7 @@
             {{ if (or .IsNode .IsPage) }}
                 {{ $issueBody := printf "**On Page:** [%s](%s)" .Title .Permalink | htmlEscape }}
                 {{ $issueQuery := (querify "body" $issueBody) }}
-                <li><a href="{{ $.Site.Params.github_base_url }}/issues/new?{{ $issueQuery | safeURL }}" target="_blank">Report Issues</a></li>
+                <li><a href="{{ $.Site.Params.github_base_url }}/issues/new?assignees=&labels=kind%2Fdocs%2C+triage%2Fneeds-triage&template=documentation.md&title=" target="_blank">Report Issues</a></li>
```
jpmcb, line 8: I'm not sure we want to completely get rid of the docs shortcode here: this would add to the body of the GitHub page which docs page the user had clicked the "Report issue" button. So clicking the report link on page `https://tanzucommunityedition.io/docs/latest/getting-started-standalone/` would bring me to a new, blank issue with `**On Page:** [](/docs/latest/getting-started-standalone/)` populated in the issue body.

But yes, there is alot of room for improvement in this user experience. We should be using the docs template to report issues so that users automatically get the labels and tags filled in. But we should still _linking back_ to the page users were on when they clicked this link

docs/README.md
```diff
@@ -20,8 +20,8 @@ Hugo is available for many platforms. It can be installed using:
 * macOS: `brew install hugo`
 * Windows: `choco install hugo-extended -confirm`
 
-Once installed, you may run the following to access a rendered view of the
-documentation:
+Once installed, you may run the following from the `site` directory
```
jpmcb, line 23: We should use the fully qualified path since `site` is not a top level directory
```suggestion
- Once installed, you may run the following from the `site` directory
+ Once installed, you may run the following from the `docs/site` directory
```

jpmcb (changes requested):

kcoriordan:
@jonasrosland this is part of the supplied template, can you take a look.
It would be nice to have both labels/tags pre-filled and page detail, but the context of the page the problem is in is more important.

jonasrosland:
If you try this:
```
                {{ $issueBody := printf "[Docs] [%s](%s)" .Title .Permalink | htmlEscape }}
                {{ $issueQuery := (querify "" $issueBody) }}
                <li><a href="{{ $.Site.Params.github_base_url }}/issues/new?assignees=&labels=kind%2Fdocs%2C+triage%2Fneeds-triage&template=documentation.md&title{{ $issueQuery | safeURL }}" target="_blank">Report Issues</a></li>
```
This is the end result, which then uses the issue template and also calls out the location of the file in the title. Thoughts?
image

jonasrosland:
@mikegehard Would you like to update your PR with the changes I proposed above?

mikegehard:
Thanks for the changes @jonasrosland. I got pulled off onto something else so feel free to make the changes.

It looks like I have given editors in this repo the ability to edit my PR. Let me know if you have any problems.

jonasrosland:
I've updated your branch @mikegehard , so we should be good to go here. Just waiting for someone else to review it :)

stmcginnis:
Looks good to me. @jonasrosland could you take another look?

stmcginnis (approved):
Looks good - thanks!

PR #2149, opened 2021-10-05
Separate command from output

mikegehard:
This will allow user to copy the command but not
get the output in the copy as well.

Fix the docs so that you don't copy the output when copying the command.

```release-note
NONE
```

Viewed generated docs.


stmcginnis (approved):
Good call - thanks!

PR #2177, opened 2021-10-06
Cii badge Addition

pnbrown:
Adds Core Infrastructure Initiative Badge to the top of the README

```release-note
NONE
```

None

N/A

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
-[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/4906/badge)](https://bestpractices.coreinfrastructure.org/projects/4906)
-
 <!-- markdownlint-disable -->
 <img src="docs/images/logos/tce-logo-only.png" width="150" align="left">
 
 # Tanzu Community Edition
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/4906/badge)](https://bestpractices.coreinfrastructure.org/projects/4906)
```
jpmcb, line 1: Hmmm this is failing mdlint - maybe place this
```
<!-- markdownlint-disable -->
```
just above this line to ignore it?

jpmcb: And on that note - do we want badges at the top? Or would it be better just below the gif on line 18?

stmcginnis: Or it could work to move the badge as the first thing right after the page title header.

If we want to keep it here, place `<!-- markdownlint-disable MD041 -->` on the very first line so we only disable this check.

pnbrown: I tried having that line first and the image didn't appear. I'll try moving it.

pnbrown: It worked that time.

jpmcb (approved):
LGTM - nice work!

PR #2256, opened 2021-10-14
Add AKO 1.6.1 template files

HanFa:
This PR adds manifests for 1.6.1 version load-balancer-and-ingress-service.

```release-note

```

Fixes: #


addons/packages/load-balancer-and-ingress-service/1.6.1/bundle/config/upstream/loadbalancerandingressservice/secret.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: avi-secret
+  namespace: avi-system
+type: Opaque
```
XudongLiuHarold, line 7: I think we should change this to be `avi.cluster.x-k8s.io/secret`, which aligns with our current 1.4 version.

HanFa: fixed!

XudongLiuHarold (approved):
LGTM, thanks @HanFa

PR #2264, opened 2021-10-15
Incorrect kernel parameter specified.

larntz:
Modified the kernel parameter to change from cgroups v2 to v1. 

If committed this PR will change `system.unified_cgroup_hierarchy=0` to `systemd.unified_cgroup_hierarchy=0`

This correct setting can be verified [here](https://www.freedesktop.org/software/systemd/man/systemd.html#systemd.unified_cgroup_hierarchy) and  [here](https://wiki.archlinux.org/title/Cgroups#Enable_cgroup_v1).

Area affected docs:linux-prerequisites

### DIFF
--- a/docs/site/content/docs/assets/prereq-linux.md
+++ b/docs/site/content/docs/assets/prereq-linux.md
@@ -24,7 +24,7 @@
     Cgroup Version: 1
     ```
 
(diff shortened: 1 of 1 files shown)


davidvonthenen (approved):
Thanks for your contribution!

Issue #1663, opened 2021-09-11
Create automation for package repository updates based on releases


joshrosso:
## Feature Request

Today, we create our package definitions for homebrew and chocolatey statically.

We should create automation that can automatically update our package definitions based on releases of TCE.

Here is an example of how `octant` does this for the chocolatey package: https://github.com/vmware-tanzu/octant/blob/master/choco/update.sh

⚠️ Note: We should not do this work until our repository is open (not a private repo).

- [x] Homebrew - https://github.com/vmware-tanzu/homebrew-tanzu/issues/19 , #2268
- [ ] Chocolatey - #2228 

## Describe alternatives you've considered

None

## Additional context

None

PR #2268, opened 2021-10-19
Add automation to update TCE homebrew formula on release

### LINKED ISSUE #19: Automate updating TCE Homebrew formula when TCE releases
## Feature Request

Automate updating TCE Homebrew formula when TCE releases. We need to update the TCE version and the SHA sums for both Mac OS and Linux OS tar balls as part of the automation

https://github.com/vmware-tanzu/homebrew-tanzu/blob/452e0ad1f3ddebc25a4b545b5f0390339777aca6/tanzu-community-edition.rb#L7

https://github.com/vmware-tanzu/homebrew-tanzu/blob/452e0ad1f3ddebc25a4b545b5f0390339777aca6/tanzu-community-edition.rb#L12

https://github.com/vmware-tanzu/homebrew-tanzu/blob/452e0ad1f3ddebc25a4b545b5f0390339777aca6/tanzu-community-edition.rb#L15

None, except for the existing manual updation of Homebrew formula

## Additional context

Some ideas around implementation - 

We can put this automation in the TCE repo, where we can create a GitHub Action workflow that will be triggered when there's a stable release. GitHub Actions has support to trigger workflows during release events - https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release . We can trigger the workflow on `released` event for stable releases. We can also ensure that the git tag name does not have any `rc` or `dev` as an extra measure instead of just depending on GitHub pre-release vs release though we have that correctly configured too, for example - https://github.com/vmware-tanzu/community-edition/releases/tag/v0.9.0 (release) , https://github.com/vmware-tanzu/community-edition/releases/tag/v0.9.0-rc.1 (pre-release)

Once a stable release is created in the TCE repo, it means that we have tar balls with signed binaries which are ready to be consumed. The automation can confirm this too by ensuring that the tar ball name has no `unsigned` term in it. So, the automation can

- Get hold of the TCE stable release version from the git tag on the release
- Calculate the SHA sums for Mac OS and Linux OS tar balls or use the TCE release's `tce-checksums.txt` file
- Pull the Homebrew tag repo
- Make changes to the TCE Homebrew formula
- Git commit
- Git push to a fork of this Homebrew tap repo or directly to this Homebrew tap repo
- Create a PR
- Run automated tests as part of this PR to ensure that the Homebrew installation works (note about this below)
- Merge the PR automatically once the tests pass

This is just one idea. We can also think about other ways. The above way seemed like a straight forward thing to me. I created the issue in this repo to just track this issue as the final Homebrew formula change and PR happens here. But yeah, automation script for the update would be put in the TCE repo, or we could put it here too, and use the script from this repo after we clone this repo. But the workflow config would be in TCE repo

About the automated tests to run as part of the PR to ensure the change is valid, we need to create an issue to talk about that separately. We would need to test it for both Mac OS and Linux OS. Also, we need to check if upgrade also works and not just installation. To check if upgrade works as part of the PR, we can install TCE from the `main` branch Homebrew tap and then upgrade using the Homebrew tap along with the PR changes containing TCE Homebrew formula updates. Again, that's just one idea

karuppiah7890:

This is to add automation to update TCE homebrew formula on release. More details - https://github.com/vmware-tanzu/homebrew-tanzu/issues/19

```release-note
Add automation to update TCE homebrew formula on release
```

Fixes: https://github.com/vmware-tanzu/homebrew-tanzu/issues/19
Part of https://github.com/vmware-tanzu/community-edition/issues/1663

Sample dummy release - https://github.com/karuppiah7890/community-edition/releases/tag/v0.51.0
Sample workflow triggered after dummy release - https://github.com/karuppiah7890/community-edition/runs/3936479124?check_suite_focus=true
Sample dummy PR in dummy Homebrew Tanzu repo - https://github.com/karuppiah7890/homebrew-tanzu/pull/3
Sample dummy automated commit after PR merge - https://github.com/karuppiah7890/homebrew-tanzu/commit/2daf14df9f252d194f6c49212b27679d9fdb3d86

Some things in my mind
- The automated commit currently shows my account along with a dummy bot account name - https://github.com/karuppiah7890/homebrew-tanzu/commit/2daf14df9f252d194f6c49212b27679d9fdb3d86 . When we use https://github.com/tce-automation account, we could maybe ensure this doesn't happen. I'll check more on this. I think this is due to the git config used. The PR uses git config (user name and email) similar to the one used for package docs automation - https://github.com/vmware-tanzu/community-edition/blob/5ac772b04237a8180a4b9ac365000bd82a0761cc/.github/workflows/package-copy-readmes-to-docs.yaml#L20-L27 and there too the commit https://github.com/vmware-tanzu/community-edition/commit/57154910d389cc1b36e1d46f8a04097b7203238b shows both tce-automation account name and github-actions account name. I think we need to fix the git config and use proper user name and email in it according to the tce-automation account, I'll try it out and check if the git config change works
- Currently the script works as shown in the demo. The script does have little complicated code to update the Homebrew formula file (ruby file). It uses `sed` and `awk`, and I have tried to ensure that it works on Mac (my local environment) and Linux too, in the CI environment. I know it's a bit complicated, I'll try to see if I can simplify it further or change the whole approach itself. A few other ideas for different approaches - 
  - Parse ruby code and modify the formula to update the version and SHAs, for example using AST etc. This might be complex though
  - Use brew CLI tool to do the task, but not sure if it can handle the if-else for the Mac and Linux SHA sums https://github.com/vmware-tanzu/homebrew-tanzu/blob/452e0ad1f3ddebc25a4b545b5f0390339777aca6/tanzu-community-edition.rb#L10-L16 , gotta check more. As of now based on whatever I have checked I can't seem to find any such options in `brew bump-formula-pr --help` or any commands to help with it in `brew commands`
- I'm also looking to add E2E tests to homebrew tap repo to test if the the Homebrew formula update is correct. Check https://github.com/vmware-tanzu/homebrew-tanzu/issues/20 . These tests would run when the automation in this PR raise PR in homebrew tap repo. We also need to ensure that the PR is automatically merged only when the PR checks pass or it should Not be merged. Currently this PR has code which merges automation PRs automatically and immediately, the same is true for package docs automation which merges the PR before checks pass, for example - https://github.com/vmware-tanzu/community-edition/pull/2238 . There are failing checks https://github.com/vmware-tanzu/community-edition/pull/2238/checks , I think the failure is due to the fact that the PR got merged immediately and didn't wait for checks to finish with a result - a success or a failure and then accordingly merge (on success) and do nothing (on failure) and exit with non-zero code to show that the workflow / pipeline / automation failed when the checks fail

### DIFF
--- a/.github/workflows/update-homebrew-package.yaml
+++ b/.github/workflows/update-homebrew-package.yaml
@@ -1,0 +1,34 @@
+name: Update Homebrew Package
+
+on:
+  release:
+    types:
+      - released
+
+jobs:
+  update-homebrew-package:
+    name: Update Homebrew Package
+    # Only run this job if we're in the main repo, not a fork.
+    if: github.repository == 'vmware-tanzu/community-edition'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Config credentials
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        run: |
+          git config --global pull.rebase true
+          git config --global url."https://git:${GITHUB_TOKEN}@github.com".insteadOf "https://github.com"
+          git config --global user.name github-actions
+          git config --global user.email github-actions@github.com
+
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Update Homebrew Package
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        shell: bash
+        run: |
+          GITHUB_REF="${{ github.ref }}"
+          version=${GITHUB_REF/refs\/tags\//}
+          ./hack/homebrew/update-homebrew-package.sh ${version}
--- a/hack/homebrew/update-homebrew-package.sh
+++ b/hack/homebrew/update-homebrew-package.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+version="${1:?TCE version argument empty. Example usage: ./hack/homebrew/update-homebrew-package.sh v0.10.0}"
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is not set}"
+
+temp_dir=$(mktemp -d)
+
+pushd "${temp_dir}"
+
+TCE_REPO_RELEASES_URL="https://github.com/vmware-tanzu/community-edition/releases"
+TCE_DARWIN_TAR_BALL_FILE="tce-darwin-amd64-${version}.tar.gz"
+TCE_LINUX_TAR_BALL_FILE="tce-linux-amd64-${version}.tar.gz"
+TCE_CHECKSUMS_FILE="tce-checksums.txt"
+TCE_HOMEBREW_TAP_REPO="https://github.com/vmware-tanzu/homebrew-tanzu"
+
+echo "Checking if the necessary files exist for the TCE ${version} release"
+
+curl -f -I -L \
+    "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_DARWIN_TAR_BALL_FILE}" > /dev/null || {
+        echo "${TCE_DARWIN_TAR_BALL_FILE} is not accessible in TCE ${version} release"
+        exit 1
+    }
+
+curl -f -I -L \
+    "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_LINUX_TAR_BALL_FILE}" > /dev/null || {
+        echo "${TCE_LINUX_TAR_BALL_FILE} is not accessible in TCE ${version} release"
+        exit 1
+    }
+
+wget "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_CHECKSUMS_FILE}" || {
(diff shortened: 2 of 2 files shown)


.github/workflows/update-homebrew-package.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - released
+
+jobs:
+  update-homebrew-package:
+    name: Update Homebrew Package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Config credentials
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```
dvonthenen, line 15: If we need a token for permissions access, will drop you a line in slack

dvonthenen: ooooh good call @nrb can help us with this!

dvonthenen: ooops looks like it is @seemiller that has access

nrb: I think @seemiller has access to that. I just checked, and I do not. Josh would be a last resort, I know he has admin access.

davidvonthenen (changes requested):
We also need to add this functionality to this PR as well:
https://github.com/vmware-tanzu/community-edition/pull/2292

.github/workflows/update-homebrew-package.yaml
seemiller: @nrb @dvonthenen Access is granted under the settings tab. I don't have that either. We'll have to find someone else that does.

davidvonthenen (approved):

PR #2277, opened 2021-10-19
[WIP] Add update strategy overlay to all core packages

blc1996:

Given that forwards compatibility with a particular Kubernetes version is not enforceable, and that certain add-ons like CNI are developed upstream, add-on updates are rollout first so that backwards compatible versions can roll out in synchronization with the new nodes joining the cluster.

It's important to prevent theadd-ons' `Pods` from being spun down and replaced with `Pods` that is intended for next release. To that end, built-in update strategies are applied and configured on the applications.

##### `DaemonSets`

For `DaemonSets`, we configure the [`.spec.updateStrategy.type`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#daemonsetupdatestrategy-v1-apps) to `OnDelete`. This ensures that the existing `Pods` will not be spun down on older nodes (thereby keeping the application available), but that new nodes entering the cluster will receive the new configuration and spin up the appropriate `Pods`. As the old nodes exit the cluster, so too will the `Pods` containing the stale configuration.

##### `Deployments`

For `Deployments` we configure four things:

* [`.spec.strategy.type`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#deploymentstrategy-v1-apps)
  is configured to `RollingUpdate`
* [`.spec.strategy.rollingUpdate.maxUnavailable`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#rollingupdatedeployment-v1-apps) is set to `0`.
* [`.spec.strategy.rollingUpdate.maxSurge`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#rollingupdatedeployment-v1-apps)
  is set to `1`.
* [`.spec.template.spec.nodeSelector`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#podspec-v1-core) is set to target only `Nodes` where the label
  `tanzuKubernetesVersion` contains the appropriate
  distribution version (i.e.
  `tanzuKubernetesVersion=v1.21.5---vmware.1-tkg.1-zshippable`). This label is set at creation time for every node in the cluster and will not be mutated (short of an end-user `cluster-admin` overwriting the label).

The `nodeSelector` ensures that the new `Pods` only target the nodes which will have the container image available, while the `maxUnavailable` and `maxSurge` settings ensure that the old `Pods` will not be spun down until there is a viable replacement.

```release-note

```

Fixes: #

### DIFF
--- a/addons/packages/antrea/0.13.3/bundle/config/overlay/update_overlay.yaml
+++ b/addons/packages/antrea/0.13.3/bundle/config/overlay/update_overlay.yaml
@@ -1,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  updateStrategy:
+    #@overlay/match missing_ok=True
+    type: OnDelete
+
+#@overlay/match by=overlay.subset({"kind":"Deployment"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  strategy:
+    #@overlay/match missing_ok=True
+    type: RollingUpdate
+    #@overlay/match missing_ok=True
+    rollingUpdate:
+      #@overlay/match missing_ok=True
+      maxUnavailable: 0
+      maxSurge: 1
+  template:
+    spec:
+    #@ if data.values.tanzuKubernetesRelease and data.values.tanzuKubernetesRelease != "":
+      #@overlay/match missing_ok=True
+      nodeSelector:
+        #@overlay/match missing_ok=True
+        tanzuKubernetesRelease: #@ data.values.tanzuKubernetesRelease
+    #@ end
--- a/addons/packages/calico/3.19.1/bundle/config/overlay/update_overlay.yaml
+++ b/addons/packages/calico/3.19.1/bundle/config/overlay/update_overlay.yaml
@@ -1,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  updateStrategy:
+    #@overlay/match missing_ok=True
+    type: OnDelete
+
+#@overlay/match by=overlay.subset({"kind":"Deployment"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  strategy:
+    #@overlay/match missing_ok=True
+    type: RollingUpdate
+    #@overlay/match missing_ok=True
+    rollingUpdate:
+      #@overlay/match missing_ok=True
+      maxUnavailable: 0
+      maxSurge: 1
+  template:
+    spec:
+    #@ if data.values.tanzuKubernetesRelease and data.values.tanzuKubernetesRelease != "":
+      #@overlay/match missing_ok=True
+      nodeSelector:
+        #@overlay/match missing_ok=True
+        tanzuKubernetesRelease: #@ data.values.tanzuKubernetesRelease
+    #@ end
--- a/addons/packages/metrics-server/0.5.1/bundle/config/overlays/update_overlay.yaml
+++ b/addons/packages/metrics-server/0.5.1/bundle/config/overlays/update_overlay.yaml
@@ -1,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  updateStrategy:
+    #@overlay/match missing_ok=True
+    type: OnDelete
+
+#@overlay/match by=overlay.subset({"kind":"Deployment"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  strategy:
+    #@overlay/match missing_ok=True
+    type: RollingUpdate
+    #@overlay/match missing_ok=True
+    rollingUpdate:
+      #@overlay/match missing_ok=True
+      maxUnavailable: 0
+      maxSurge: 1
+  template:
+    spec:
+    #@ if data.values.tanzuKubernetesRelease and data.values.tanzuKubernetesRelease != "":
+      #@overlay/match missing_ok=True
+      nodeSelector:
+        #@overlay/match missing_ok=True
+        tanzuKubernetesRelease: #@ data.values.tanzuKubernetesRelease
+    #@ end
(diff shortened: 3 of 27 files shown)


addons/packages/pinniped/0.4.4/bundle/config/values.yaml
```diff
@@ -1,6 +1,7 @@
 #@data/values
 #@overlay/match-child-defaults missing_ok=True
 ---
+tanzuKubernetesRelease: null
```
ankeesler, line 4: where does this get set? is it in the addon secret ytt template over in tanzu-framework?

it is kinda a bummer that we need to add yet another system-specific data value - the list of data values is already growing too large for a user to keep track of what they need to set. but i think i understand the need.

the new data value is also one of those that should always be provided by the underlying cluster (similar to `dex.dns.vsphere.*` or `pinniped.supervisor_ca_bundle_data`). i think this is confusing for users because they don't know which data values that they should set and which data values that the addon secret ytt template will provide. again, i understand the need for the data value here, but it seems like it hurts the UX.

i wonder if it would be more helpful to make data values like these more abstract in the future. for example, instead of `tanzuKubernetesRelease`, we could have a `nodeSelector` map data value where users can provide `nodeSelector` map values for their `Deployment`'s. then, the new data value is abstract enough that 1) it can be used to resolve this use case of needing to make sure deployments are scheduled on the right nodes and 2) it could be used by the user for other scheduling constraints outside of reason 1. just food for thought at this point.

addons/packages/pinniped/0.4.4/bundle/config/overlay/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"DaemonSet"}), expects="0+"
```
ankeesler, line 4: pinniped doesn't use `DaemonSet`'s, so this is dead code. i understand the need to make the `update_overlay.yaml` files uniform, but should we remove this part of the overlay for packages that don't have `DaemonSet`'s?

addons/packages/pinniped/0.4.4/bundle/config/overlay/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
```
ankeesler, line 1: this overlay seems like it is yet another duplicated piece of logic throughout multiple packages. do we have plans to get rid of this duplication? i remember talking about this with @danniel1205 a while back.

addons/packages/pinniped/0.4.4/bundle/config/overlay/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
...
+      #@overlay/match missing_ok=True
+      maxUnavailable: 0
+      maxSurge: 1
+  template:
+    spec:
+    #@ if data.values.tanzuKubernetesRelease and data.values.tanzuKubernetesRelease != "":
+      #@overlay/match missing_ok=True
+      nodeSelector:
+        #@overlay/match missing_ok=True
+        tanzuKubernetesRelease: #@ data.values.tanzuKubernetesRelease
```
ankeesler, line 30: i don't think that pinniped has a dependency on the kube version right now, i.e., if we update the package version and then we update the cluster version then the pinniped package will still continue to function (unlike CNI). are we sure it makes sense to add this logic to the package even though it is unnecessary? it seems like it just bloats the package with logic that we don't need.

addons/packages/metrics-server/0.5.1/bundle/config/overlays/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
+#@ load("@ytt:overlay", "overlay")
```
benjaminapetersen, line 1: Looking at all of these, is `update_overlay.yaml` a descriptive enough name that we will remember why these files exist months from now?  I'd probably prefer it be more descriptive.

Can we encode the "why" in the file name?

addons/packages/metrics-server/0.5.1/bundle/config/overlays/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
...
+#@ load("@ytt:data", "data")
```
benjaminapetersen, line 2: And if this is sort of "universal to all packages"  type of file, it would be nice if it was either:
1.) copy in 1 place outside each package, rather than duplicating and have it applied as an additional templating step.
or 2.) stored in a directory within `/overlays` as like `/overlays/system` or `/overlays/tkg-system`.  

At some point I imagine there is bound to be a template ordering bug, so knowing the "system" templates are applied as a separate step could be helpful.

addons/packages/pinniped/0.4.4/bundle/config/overlay/update_overlay.yaml
benjaminapetersen: And if they MUST be duplicated yet consistent, I think it warrants a good comment explaining why. But I definitely prefer a solution that doesn't require the [same file duplicated across all packages](https://github.com/vmware-tanzu/community-edition/pull/2277/files#r732801550).

addons/packages/antrea/0.13.3/bundle/config/overlay/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
...
+    #@overlay/match missing_ok=True
+    type: OnDelete
+
+#@overlay/match by=overlay.subset({"kind":"Deployment"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  strategy:
+    #@overlay/match missing_ok=True
+    type: RollingUpdate
```
12345lcr, line 18: For calico-kube-controllers, the default strategy is `Recreate`. Because controllers can only have one single active instance, otherwise there will be potential problem like reconciling at the same time and creating duplicate resources.

For more details see here: https://github.com/projectcalico/calico/blob/master/getting-started/kubernetes/installation/calico-kube-controllers.yaml#L19-L25

addons/packages/calico/3.19.1/bundle/config/overlay/update_overlay.yaml
```diff
@@ -0,0 +1,31 @@
...
+    #@overlay/match missing_ok=True
+    type: OnDelete
+
+#@overlay/match by=overlay.subset({"kind":"Deployment"}), expects="0+"
+---
+spec:
+  #@overlay/match missing_ok=True
+  strategy:
+    #@overlay/match missing_ok=True
+    type: RollingUpdate
```
12345lcr, line 18: For calico-kube-controllers, the default strategy is Recreate. Because controllers can only have one single active instance, otherwise there will be potential problem like reconciling at the same time and creating duplicate resources.

For more details see here: https://github.com/projectcalico/calico/blob/master/getting-started/kubernetes/installation/calico-kube-controllers.yaml#L19-L25

joshrosso:
Closing this PR due to inactivity due to merge conflicts. If you'd still like to make these improvements, please re-create a PR against `main`.

Additionally [we do recommend opening an issue describing the desired changes before opening a PR](https://tanzucommunityedition.io/docs/latest/contribute/contributing/#contribution-workflow).

PR #2296, opened 2021-10-20
Fix lint issues and cleanup

stmcginnis:

Fixes some of the linting errors causing job failures.

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/kapp/kapp.go
+++ b/cli/cmd/plugin/standalone-cluster/kapp/kapp.go
@@ -1,7 +1,14 @@
+// Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package kapp contains code for interacting with kapp controller.
 package kapp
 
 import (
 	"fmt"
+	"log"
+	"strings"
+
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -17,8 +24,6 @@ import (
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
 	apiRegv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
-	"log"
-	"strings"
 )
 
 const (
@@ -29,6 +34,8 @@ const (
 	kappControllerDeploymentName = "kapp-controller"
 )
 
+// KappClient is a client for interfacing with kapp-controller.
+//nolint:golint
 type KappClient struct {
 	dynClient  dynamic.Interface
 	clientSet  kubernetes.Interface
@@ -36,7 +43,8 @@ type KappClient struct {
 	scheme     *runtime.Scheme
 }
 
-// KappInstallOpts contains information about how to install kapp-controller
+// KappInstallOpts contains information about how to install kapp-controller.
+//nolint:golint
 type KappInstallOpts struct {
 	// MergedManifests represents the final (serialized) data for installing kapp-controller and its ancillary
 	// resources. This assumes all objects are merged into one byte array. Think of this like a single YAML file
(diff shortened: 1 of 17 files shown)


cli/cmd/plugin/standalone-cluster/kapp/kapp.go
```diff
@@ -77,7 +88,7 @@ func New(kubeconfig string) KappManager {
 
 	// setting up restMapper is an expensive operation, so do it here and re-use it
 	// in method invocations
-	groupResources, err := restmapper.GetAPIGroupResources(clientSet.Discovery())
+	groupResources, _ := restmapper.GetAPIGroupResources(clientSet.Discovery())
```
jpmcb, line 91: Should we check the error here as well?

jpmcb: Looks like a number of places in this file we are ignoring errors - I don't have much context on the `kapp` package, so if this is going to be refactored anyways, no worries

cli/cmd/plugin/standalone-cluster/kapp/kapp.go
```diff
@@ -119,25 +131,26 @@ func (k KappClient) Install(opts KappInstallOpts) (*v1.Deployment, error) {
...
-
 	pods, err := k.clientSet.CoreV1().Pods(ns).List(metav1.ListOptions{})
 	if err != nil {
 		return "failed to talk to cluster"
 	}
 
 	var kappPod *corev1.Pod
 	// super bad, should not be iterating through pods on every invocation
-	for _, pod := range pods.Items {
+	for i := range pods.Items {
```
jpmcb, line 151: hmmm was the linter failing on `for _, pod`? Seems weird to have to pull the item out manually using the index instead of the slice item itself

cli/cmd/plugin/standalone-cluster/packages/packages.go
```diff
@@ -93,7 +96,7 @@ func NewClient(kubeconfig string) PackageManager {
 	}
 
 	// register packaging APIs
-	packaging.AddToScheme(scheme.Scheme)
+	_ = packaging.AddToScheme(scheme.Scheme)
```
jpmcb, line 99: Handle the error?

cli/cmd/plugin/standalone-cluster/tkr/image.go
```diff
@@ -147,9 +151,7 @@ func (t *TkrImage) AddYttYamlValuesBytes(b []byte) error {
 }
 
 func (t *TkrImage) AddYttKVsFromYAML(kvs []string) {
-	for _, kv := range kvs {
-		t.YttKVsFromYAML = append(t.YttKVsFromYAML, kv)
-	}
+	t.YttKVsFromYAML = append(t.YttKVsFromYAML, kvs...)
```
jpmcb, line 154: Ooo nice!

cli/cmd/plugin/standalone-cluster/kapp/kapp.go
```diff
@@ -62,8 +72,9 @@ type KappManager interface {
 	Status(ns, name string) string
 }
 
+// New instantiates a new KappManager.
 func New(kubeconfig string) KappManager {
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	config, _ := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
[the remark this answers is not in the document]
stmcginnis: Probably. It was being ignored before, so I just made it explicit that we weren't doing anything with it right now. I have a feeling we will end up refactoring this quite a bit, so I think we are OK just skipping checks for now. If you'd rather I add the error checking now I can - just let me know. Either way is fine with me.

cli/cmd/plugin/standalone-cluster/kapp/kapp.go
stmcginnis: This is why I like this lint check. It's actual a very subtle bug, and not something you would expect based on just looking at the code. But when you get the pointer to the variable, you actually end up with the pointer to the loop variable and not the actual item. So even doing something like `pod := pod` fixes that, but looks like WTH.

https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

stmcginnis: This is why I like this lint check. It's actual a very subtle bug, and not something you would expect based on just looking at the code. But when you get the pointer to the variable, you actually end up with the pointer to the loop variable and not the actual item. So even doing something like `pod := pod` fixes that, but looks like WTH.

https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

Issue #1383, opened 2021-08-20
`tanzu conformance version` shows empty SHA for Sonobuoy


karuppiah7890:
## Bug Report

`conformance` command version shows empty SHA for Sonobuoy

```
$ tanzu conformance version
Sonobuoy Version: v0.52.0
Sonobuoy SHA: 
TCE Version: v0.7.0
TCE SHA f80a6676
```

## Expected Behavior

To show Sonobuoy SHA like [0a59430c](https://github.com/vmware-tanzu/sonobuoy/commit/0a59430cf835)

## Steps to Reproduce the Bug

Run `tanzu conformance version` with TCE version v0.7.0

## Environment Details

* Build version (`tanzu version`): TCE version -  f80a6676 / v0.7.0
* Operating System (client): MacOS

## Additional information

Looks like we use a particular version of Sonobuoy - https://github.com/vmware-tanzu/tce/blob/main/cli/cmd/plugin/conformance/go.mod#L7 , [0a59430cf835](https://github.com/vmware-tanzu/sonobuoy/commit/0a59430cf835)

And I think we will have to set the `buildinfo.GitSHA` at https://github.com/vmware-tanzu/tce/blob/main/cli/cmd/plugin/conformance/pkg/version.go#L33 during built time with the SHA in the [`go.mod`](https://github.com/vmware-tanzu/tce/blob/main/cli/cmd/plugin/conformance/go.mod#L7). A bit of a tricky thing

`GitSHA` info from Sonobuoy from version [0a59430cf835](https://github.com/vmware-tanzu/sonobuoy/commit/0a59430cf835) - https://github.com/vmware-tanzu/sonobuoy/blob/0a59430cf835618821bc66f732b6c8557d6a407b/pkg/buildinfo/version.go#L25-L26

PR #2304, opened 2021-10-21
Add workflow config and script to test MacOS, Linux OS and Windows OS release builds

### LINKED ISSUE #2269: Run automated tests using released artifacts
As of now, we run tests for commits pushed / merged to `main` branch and during releases, we tag the commits on `main` and then publish the release. This means that the commits are tested as part of the `main` branch tests

We currently don't use TCE release tar balls to do tests though. We only test after doing `make release` and building TCE. The main difference between TCE release tar balls and the `make release` is that - TCE release tar balls contain signed binaries for Mac and Windows, unlike unsigned binaries present in `make release` artifacts. For this reason, there's a small difference between the two and we might have to do a bit of testing for the release tar balls. For example, we can test -

- If the TCE installation from the release tar ball works on Mac and Windows. It was noticed in one of the releases that the MacOS TCE release tar ball had one improperly signed binary, due to some signing process issues, which lead to TCE installation on Mac to not work. This was noticed only after the release after manual installation. This cannot be caught in existing tests because existing tests run on unsigned binaries, not signed binaries. Also, the existing tests run on Linux. To catch such issues, we have to run tests on Windows and Mac and test the installation to see if there are any binary signing issues.
- If TCE (installed from release tar ball) works on different cloud providers (AWS, Azure, GCP) and in local Docker. This is not be necessary though, as a signed binary would not have any feature (code) level differences compared to unsigned binaries which we already test, unless signing process went wrong resulting in a corrupted binary, in which case nothing related to that binary will work. To check if the binary works at a basic level, we can do the below check
- If TCE (installed from release tar ball) works and shows versions of the different TCE and TF (core) plugins. This ensures that the binaries are in a good state and are not corrupted etc

Sub issues
- [x] #2302
- [x] #2354
- [x] #2355

### LINKED ISSUE #2302: Test TCE MacOS release build
Check #2269

### LINKED ISSUE #2354: Test TCE Linux release build
Check #2269

### LINKED ISSUE #2355: Test TCE Windows release build
Check #2269

What we will verify as part of the test
- Windows binaries are signed correctly
- Windows binaries work

This flow assumes that the user is installing TCE using TCE Windows release zip file, without Chocolatey on Windows. Also, Chocolatey installation however uses TCE Windows release zip file - https://github.com/vmware-tanzu/community-edition/blob/c00d2a3e9cce3339a773ff881a5db3fd2d7aadf2/hack/choco/tools/chocolateyinstall.ps1#L11 so the underlying artifact is going to be the same and we will test the zip file installation directly

The flow would look like this
- Download the TCE Windows zip file, for example https://github.com/vmware-tanzu/community-edition/releases/download/v0.9.1/tce-windows-amd64-v0.9.1.zip 
- Unzip the zip file
- Check if the binaries in `bin` folder are signed correctly
- Install TCE using `install.bat` script
- Optionally update Path environment variable to include `C:\Program Files\tanzu`. It's optional because it looks like `tanzu` can work without this change too if no new prompt has been opened / restarted just after the installation, but when a new command prompt window is opened, then it doesn't work if Path environment variable isn't updated correctly

Binary Digital Signature verification tool(s)
- We can use [`signtool`](https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool) to verify the signature in the binaries using `signtool verify` command. `signtool` command comes with Windows SDK, for example https://developer.microsoft.com/en-US/windows/downloads/windows-sdk/

Example -

```
C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x64>signtool.exe verify /pa C:\Users\Administrator\Downloads\tce-windows-amd64-v0.9.1\tce-windows-amd64-v0.9.1\bin\tanzu.exe
File: C:\Users\Administrator\Downloads\tce-windows-amd64-v0.9.1\tce-windows-amd64-v0.9.1\bin\tanzu.exe
Index  Algorithm  Timestamp
========================================
0      sha1       Authenticode

Successfully verified: C:\Users\Administrator\Downloads\tce-windows-amd64-v0.9.1\tce-windows-amd64-v0.9.1\bin\tanzu.exe
```

karuppiah7890:

This PR adds workflow config and script to test MacOS, Linux OS and Windows OS release builds by
- Checking if the MacOS release build binaries are signed correctly using `spctl`
- Checking if the Windows release build binaries are signed correctly using `signtool`
- Installing the MacOS, Linux OS and Windows OS release builds
- Running the `version` command of all the plugins (TCE + TF) plugins and also `tanzu` CLI, to check if all the binaries are working

```release-note
Add workflow config and script to test MacOS, Linux OS and Windows OS release builds
```

Fixes: #2302 
Fixes: #2354 
Fixes: #2355 
Fixes: #2269 

- Demo of a failing pipeline where there's a signing issue in TCE v0.9.0 release darwin tar ball - https://github.com/karuppiah7890/community-edition/runs/4043000515?check_suite_focus=true . Note the error log -

```
+ spctl -vv --type install --asses /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp.Oo4UK9zg/tce-darwin-amd64-v0.9.0/bin/tanzu-plugin-kubernetes-release
/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp.Oo4UK9zg/tce-darwin-amd64-v0.9.0/bin/tanzu-plugin-kubernetes-release: rejected
source=Unnotarized Developer ID
origin=Developer ID Application: VMware, Inc. (EG7KH642X6)
```

The error also triggers a cancel on the linux job - https://github.com/karuppiah7890/community-edition/runs/4043000548?check_suite_focus=true due to a default fail fast strategy in a matrix. 

The pipeline was triggered using a `workflow_dispatch` event like this -

```bash
$ gh workflow run --repo https://github.com/karuppiah7890/community-edition/ --ref release-build-test-macos-trial e2e-release-macos-linux-builds.yaml -f release_version=v0.9.0 

✓ Created workflow_dispatch event for e2e-release-macos-linux-builds.yaml at release-build-test-macos-trial

To see runs for this workflow, try: gh run list --workflow=e2e-release-macos-linux-builds.yaml
```

- Demo of successful pipelines when there are no issues in TCE v0.9.1 release darwin and linux tar ball - https://github.com/karuppiah7890/community-edition/runs/4043000920?check_suite_focus=true and https://github.com/karuppiah7890/community-edition/runs/4043000945?check_suite_focus=true . The pipeline was triggered using a `workflow_dispatch` event like this -

```bash
$ gh workflow run --repo https://github.com/karuppiah7890/community-edition/ --ref release-build-test-macos-trial e2e-release-macos-linux-builds.yaml -f release_version=v0.9.1

✓ Created workflow_dispatch event for e2e-release-macos-linux-builds.yaml at release-build-test-macos-trial

To see runs for this workflow, try: gh run list --workflow=e2e-release-macos-linux-builds.yaml
```

- Demo of a pipeline running Windows OS release build test - https://github.com/karuppiah7890/community-edition/runs/4064876679?check_suite_focus=true (failure due to #2397 installation script issue in `v0.9.1` TCE release for Windows). It was triggered like this -

```bash
$ gh workflow run --repo https://github.com/karuppiah7890/community-edition/ --ref release-build-test-macos-trial e2e-release-windows-builds.yaml -f release_version=v0.9.1

✓ Created workflow_dispatch event for e2e-release-windows-builds.yaml at release-build-test-macos-trial

To see runs for this workflow, try: gh run list --workflow=e2e-release-windows-builds.yaml
```

This pipeline will be triggered from downstream pipeline once the downstream pipeline builds and signs the binaries and uploads the tar ball to the draft release. We can use `gh` CLI downstream to trigger the workflow

The script in this PR uses `gh` to download release assets which supports downloading both published and draft/unpublished release artifacts (tar balls). Of course the user authenticating using `gh` has to have enough access to view draft releases. I guess in our case that would be the release access token user (`GH_RELEASE_ACCESS_TOKEN`). Let me know if I need to use some other token

The workflow takes the version / tag as input from which the release artifact is downloaded

### DIFF
--- a/test/release-build-test/check-release-build.sh
+++ b/test/release-build-test/check-release-build.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+version="${1:?TCE version argument empty. Example usage: ./test/release-build-test/check-release-build.sh v0.10.0}"
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is not set}"
+
+TCE_REPO_URL="https://github.com/vmware-tanzu/community-edition"
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH="amd64"
+
+temp_dir=$(mktemp -d)
+
+TCE_TAR_BALL="${temp_dir}/tce-${OS}-${ARCH}-${version}.tar.gz"
+TCE_INSTALLATION_DIR="${temp_dir}/tce-${OS}-${ARCH}-${version}"
+
+gh release download "${version}" --repo ${TCE_REPO_URL} --pattern "tce-${OS}-${ARCH}-${version}.tar.gz" --dir "${temp_dir}"
+
+tar xvzf "${TCE_TAR_BALL}" --directory "${temp_dir}"
+
+if [ "${OS}" == 'darwin' ]; then
+  for binary in "${TCE_INSTALLATION_DIR}"/bin/*; do
+    spctl -vv --type install --asses "${binary}"
+  done
+fi
+
+"${TCE_INSTALLATION_DIR}"/install.sh
+
+tanzu version
+
+tanzu cluster version
+
+tanzu conformance version
+
+tanzu diagnostics version
+
+tanzu kubernetes-release version
+
+tanzu management-cluster version
+
+tanzu package version
+
+tanzu standalone-cluster version
+
+tanzu pinniped-auth version
+
+tanzu builder version
+
+tanzu login version
--- a/.github/workflows/release-gate-macos-linux-builds.yaml
+++ b/.github/workflows/release-gate-macos-linux-builds.yaml
@@ -1,0 +1,27 @@
+name: Release Gate - MacOS and Linux OS Release Builds
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Release version that has to be tested"
+        required: true
+
+jobs:
+  release-gate-macos-and-linux-release-builds:
+    name: Release Gate - MacOS and Linux OS Release Builds
+    # Only run this job if we're in the main repo, not a fork.
+    if: github.repository == 'vmware-tanzu/community-edition'
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Release Gate - MacOS and Linux OS Release Builds
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        run: |
+          ./test/release-build-test/check-release-build.sh ${{ github.event.inputs.release_version }}
--- a/.github/workflows/release-gate-windows-build.yaml
+++ b/.github/workflows/release-gate-windows-build.yaml
@@ -1,0 +1,25 @@
+name: Release Gate - Windows OS Release Build
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Release version that has to be tested"
+        required: true
+
+jobs:
+  release-gate-windows-release-build:
+    name: Release Gate - Windows OS Release Build
+    # Only run this job if we're in the main repo, not a fork.
+    if: github.repository == 'vmware-tanzu/community-edition'
+    runs-on: windows-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Release Gate - Windows OS Release Build
+        shell: powershell
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        run: |
+          .\test\release-build-test\check-release-build.ps1 -version ${{ github.event.inputs.release_version }} -signToolPath "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x64\signtool.exe"
--- a/test/release-build-test/check-release-build.ps1
+++ b/test/release-build-test/check-release-build.ps1
@@ -1,0 +1,109 @@
+# Copyright 2020-2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+param (
+    # TCE release version argument
+    [Parameter(Mandatory=$True)]
+    [string]$version,
+
+    # Path to the signtool
+    [Parameter(Mandatory=$True)]
+    [string]$signToolPath
+)
+
+$ErrorActionPreference = 'Stop';
+
+if ((Test-Path env:GITHUB_TOKEN) -eq $False) {
+  throw "GITHUB_TOKEN environment variable is not set"
+}
+
+$tempFolderPath = Join-Path $Env:Temp $(New-Guid)
+New-Item -Type Directory -Path $tempFolderPath
+
+$TCE_REPO_URL = "https://github.com/vmware-tanzu/community-edition"
+
+gh release download $version --repo $TCE_REPO_URL --pattern "tce-windows-amd64-$version.zip" --dir $tempFolderPath
+
+Expand-Archive -LiteralPath "$tempFolderPath\tce-windows-amd64-$version.zip" -Destination $tempFolderPath
+
+# Check if the binaries are all signed
+Get-ChildItem -Path "$tempFolderPath\tce-windows-amd64-$version\bin\tanzu-*" -File -Recurse | Foreach-Object {
+  & $signToolPath verify /pa $_.FullName
+  if ($LastExitCode -ne 0) {
+    throw "Error verifying: " + $_.FullName
+  }
+}
+
+Push-Location "$tempFolderPath\tce-windows-amd64-$version"
+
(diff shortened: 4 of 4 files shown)


davidvonthenen (changes requested):
So the issue with this is we need to test the binaries *before* they are released. This should be done after the binaries are signed but before the release is published.

The goal should be preventing publishing a release with a signing or functionality issue while signing. If you want to discuss this further, please hit me up in slack.

.github/workflows/e2e-release-macos-linux-builds.yaml
```diff
@@ -0,0 +1,27 @@
+name: E2E Test - MacOS and Linux OS Release Builds
```
dvonthenen, line 1: I'm wondering if this Prefix for the action should be changed. This is a test to check the release bits before we publish so it's like an e2e in a way but it's also kinda different and special. I think maybe we should call it something like `Release Gate -` so it's kinda lumped in with the Release actions, but I'm open to ideas and suggestions.

davidvonthenen (commented):
This is another awesome check!

.github/workflows/e2e-release-macos-linux-builds.yaml
navidshaikh: I like `Release Gate - `

test/release-build-test/check-release-build.sh
```diff
@@ -0,0 +1,55 @@
...
+TCE_INSTALLATION_DIR="${temp_dir}/tce-${OS}-${ARCH}-${version}"
+
+gh release download "${version}" --repo ${TCE_REPO_URL} --pattern "tce-${OS}-${ARCH}-${version}.tar.gz" --dir "${temp_dir}"
+
+tar xvzf "${TCE_TAR_BALL}" --directory "${temp_dir}"
+
+if [ "${OS}" == 'darwin' ]; then
+    for binary in "${TCE_INSTALLATION_DIR}"/bin/*; do
+    spctl -vv --type install --asses "${binary}"
+    done
```
navidshaikh, line 30:
```suggestion
    for binary in "${TCE_INSTALLATION_DIR}"/bin/*; do
      spctl -vv --type install --asses "${binary}"
    done
```

test/release-build-test/check-release-build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+tanzu package version
+
+tanzu standalone-cluster version
+
+tanzu pinniped-auth version
+
+tanzu builder version
+
+tanzu login version
```
navidshaikh, line 55: can also check the version match for TCE plugins

navidshaikh (commented):
Neat!

test/release-build-test/check-release-build.sh
karuppiah7890: > can also check the version match for TCE plugins

@navidshaikh Makes sense from face value, there is still some complexity (more code etc) though. For example, for `standalone-cluster` plugin, the output is like this

```bash
$ tanzu standalone-cluster version
v0.9.1
```

for conformance, the output is like this

```bash
$ tanzu conformance version
Sonobuoy Version: v0.53.2
TCE Version: v0.9.1
TCE SHA 00d8d13a
```

We might probably end up using `grep` to do a simple check of the version `v0.9.1` in the above 3 line output, compared to an exact output check which could be a bit more verbose to check all versions including SHA, sonobuoy version. Both of it is possible. I'm just thinking of refraining from doing that though, as there is also a need for updating the tests before the release with the expected version when we include test for Sonobuoy version, but yeah other things can be inferred - `v0.9.1` we already have from `release_version` input and we can also find the corresponding `git` SHA for `TCE SHA`.

Also, are we only talking about the 3 TCE plugins here or even Tanzu Framework plugins? I think during one discussion or even in my though process while writing this script, there was an idea to do a version check for all of them, because - why not? Why stop at just checking if the CLI plugins work with `version` command

I think a good way to take a call in this case would be to think about the goal of the script. The goal of this part of the script is currently to simply check if all the `tanzu` CLI plugins (Tanzu Framework + TCE plugins) work by invoking the `version` command which invokes the CLI plugin binary to get the version. When I say `work` - I mean that the binary is present and installed and picked up by `tanzu` CLI and is _not_ corrupt after the binary signing process. We could also add the goal of checking if the version of the CLI plugins (TCE or both TCE + Tanzu Framework CLI plugins) are all correct

karuppiah7890: Adding a version command output check at release gate makes sense in terms of avoiding issues like #1383 , though in the case of #1383 the `Sonobuoy SHA` was entirely removed from the output. Also, doing `version` command output check doesn't exactly make sense in `main` for non-release commits, unless we want to see if SHA was set right to the version info or `dev` version shows up, which is less priority than a released CLI version output which people use while reporting issues

karuppiah7890: I think it makes sense to add it, let me know to what extent - like, just check for example `v0.9.1` TCE release version in the `version` command of TCE plugins or more in case of `conformance` CLI plugin and what about Tanzu Framework CLI plugins

navidshaikh: Sure, we can do version check in a follow up PR.

.github/workflows/e2e-release-macos-linux-builds.yaml
karuppiah7890: Makes sense, I'll change the GitHub Actions Workflow yaml file name and `name` field and related fields to use `Release Gate` than `E2E Test`. I'm assuming we are okay with bash script file name

karuppiah7890: Updated the PR 👍 Let me know if it makes sense

navidshaikh (approved):

davidvonthenen (approved):

PR #2316, opened 2021-10-21
Update documentation regarding default values #607

### LINKED ISSUE #607: Default namespace for packages should be upstream expected namespaces
## Feature Request
Make default namespaces for packages that bundle content that is supposed to be installed in a specific namespace to be the expected upstream/default namespace.

A user of TCE packages with knowledge on the technology will by default expect it to be installed in the same destination as if they consume the upstream package. We do still need to support moving into an alternate namespace, as some used might want to concentrate all/some packages into fewer namespaces as they are considered infrastructure components and hence reducing visibility problems for end users.

Why would someone want fewer namespaces for infrastructure components? Mostly because k8s does not provide proper RBAC around the namespaces a user can see, so it's all or none. If there's lots of namespaces in the cluster, users that have only RBAC in a few of them will still see the whole list. This is really cumbersome in UIs and namespace selector experiences.

seemiller:

Updates documentation regarding default values.

```release-note

```

Fixes: #607

### DIFF
--- a/docs/site/content/docs/latest/designs/package-process.md
+++ b/docs/site/content/docs/latest/designs/package-process.md
@@ -242,7 +242,10 @@ site](https://carvel.dev/ytt/#example:example-overlay)._
 ### 4. Default Values
 
 For every user-configurable value defined above, a `values.yaml` file should
-contain defaults and documentation for what the parameter impacts.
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
This looks good to me.

PR #2332, opened 2021-10-23
Load local cluster config by order of precedent

stmcginnis:

This updates how we determine local cluster configuration by setting the
values based on:

1. explicit arguments
2. environment variables
3. configuration files
4. default values

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/configure.go
+++ b/cli/cmd/plugin/standalone-cluster/configure.go
@@ -6,11 +6,13 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"os"
+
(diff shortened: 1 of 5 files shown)


joshrosso (commented):
For the struct to map conversion, would it be helpful to [copy this code over](https://github.com/fatih/structs/blob/master/structs.go#L81-L150)?

I think it could be really helpful if we bury it in a private function. We'd also want to convert it to functions, rather than methods, which reading it, should be quite trivial.

joshrosso (commented):
Since environment variable detection happens based on assuming the struct fields are in camel case, this PR needs to change the config struct values like `CNI` to `Cni`.

PR #2342, opened 2021-10-25
make avi controller version configurable

XudongLiuHarold:

- Since `AKO Operator` and `AKO` both support multi AVI Controller versions, this PR makes AVI Controller Version configurable, so users can set the controller version to be the one they actually deployed.

- Allow user to change default `AKODeploymentConfig(adc)` 's controller setting.

```release-note
This PR allows users to set AVI controller version and change the default adc's AVI controller configurations, so users now can switch AVI controllers.
```

Fixes: #


HanFa (approved):
Left a comment @XudongLiuHarold

PR #2344, opened 2021-10-25
Introduce config directory per cluster

joshrosso:
This introduces a configuration directory per cluster at
`~/.config/tanzu/tkg/local/${CLUSTER_NAME}`. It stores the rendered
(flattened) configuration there.

This commit also re-wires the delete command to support multiple
provides and load the config.

Signed-off-by: joshrosso <rossoj@vmware.com>


jpmcb (approved):
looks good!

I created 2 clusters, noted the directories created with the configs and then deleted both without problem.

Saw you just pushed a min ago - is this ready to merge? (feel free to do so!)

PR #2345, opened 2021-10-25
Set default tty mode based on runtime

stmcginnis:
This sets the default value for the `--tty` CLI setting for local
cluster create and configure to match what the current runtime supports.
This makes it so the user shouldn't need to specify the `--tty` flag in
most cases unless they have a specific reason for changing it.


joshrosso (changes requested):
Along with other comments, we should wire this into `delete.go` and `list.go`.

PR #2349, opened 2021-10-25
Inline some TKR info

stmcginnis:
This tries to keep things DRY and hopefully gets a step closer to having
a simplified TKR structure.


jpmcb (approved):
Sweet!!

PR #2375, opened 2021-10-27
Implements the `noop` cluster manager and implements `List` in tanzu package

jpmcb:

All changes related to the standalone cluster overhaul

- Enables users to bring their own cluster with their own kubeconfig
- Removes listing as part of the cluster package and moves it's functionality to the tanzu package. Now, users can list all clusters under the config dir (wheather they were created by our provider or were `provider: none`)
- Adds error logging in kapp controller manager
- Updates calls to NewKindClusterManager to be generic

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/cluster/cluster.go
+++ b/cli/cmd/plugin/standalone-cluster/cluster/cluster.go
@@ -19,7 +19,7 @@ type KubernetesCluster struct {
 	// Name is the name of the cluster.
 	Name string
 	// KubeConfig contains the Kubeconfig data for the cluster.
-	Kubeconfig string
+	Kubeconfig []byte
 }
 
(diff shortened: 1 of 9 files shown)


joshrosso (changes requested):
Nothing wrong with this PR just want to be included in the final review before merge.

stmcginnis (changes requested):
Looks great! One minor thing I'd like to see changed, but otherwise ready to go.

stmcginnis (approved):
Thanks! I'll leave to @joshrosso to push the big button.

joshrosso (commented):
Our Tanzu API needs to support bringing the `kubeconfig` in as a byte array rather than a path to a file. This will be more common for those that call us programatically.

joshrosso (changes requested):
overall looks great, added some small recommendations for change.

PR #2401, opened 2021-10-29
Standalone overhaul: implements context aware progress animation

jpmcb:
All changes related to the standalone-overhaul

---

This PR implements two new methods for the logger:

- `AnimateProgressf` which will animate dots based on the provided format message (and it's args)
- `AnimateStatusf` is similar to `AnimateProgressf` but also takes a string chan that contains the most recent "status". This is intended to be printed via a format string provided as the "message"

In short, this allows us to "animate" both blocking and non blocking calls. For example, this PR also implements animating and on calls to the TkrImageReader download image method.

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/log/log.go
+++ b/cli/cmd/plugin/standalone-cluster/log/log.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


cli/cmd/plugin/standalone-cluster/log/log.go
```diff
@@ -89,11 +91,26 @@ type Logger interface {
...
 	// The progress counter will render as a quantity of "." characters defined by the value of count.
-	// Usage of Progressf typically involves a for loop feeding in a series of numbers such as
+	// Usage of Progressf typically involves calling either AnimateProgressf, AnimateStatusf,
+	// or writing a custom for loop feeding in a series of numbers such as
 	// 1,2,3,1,2,3,1,2,{exit due to condition}.
-	// Progressf will start each message with '\r', which will overwrite the last line. This gives the appearance of
-	// updating.
+	// Progressf will start each message with '\r', which will overwrite the last line.
+	// This gives the appearance of updating.
 	Progressf(count int, message string, args ...interface{})
```
joshrosso, line 99: Is there a way we can make progressf private or something? Worried about this interface growing too large

joshrosso (changes requested):
@jpmcb this look really great!

My only feedback is that it feels like our interface declares the need for 3 definitions, where 1 could work. Thoughts? Any way to simply merge Progressf AnimateProgressf and AnimateStatusf?

joshrosso:
i'm going to merge this, but we'll need to circle back sometime to figure out the `--tty=false` use case. Seems it spreads the status logs out without a line break.

```txt
Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster             Creating cluster
```

When TTY is on, it looks great!

jpmcb:
https://github.com/vmware-tanzu/community-edition/pull/2473 addresses the problem with `tty=false` logging with the progress animator

PR #2402, opened 2021-10-29
Send standalone bootstrap log to file

stmcginnis:
This updates the logging for the local cluster creation process to
capture all activity to the bootstrap.log file in the cluster's config
directory.


joshrosso:
works great 🙏 .

We need to circle back and let any arbitrary level be logged to this file.

PR #2408, opened 2021-10-29
Separate image pull from cluster creation

stmcginnis:
In regions where the cluster image hasn't been cached yet, it can take
some time to pull down the image. Right now this just happens during
cluster creation, so the CLI prints out "Creating cluster", then appears
to hang for a very long time.

We can't do too much to speed up the image pull, but we can at least
break it out into its own step so it's more obvious to the user what is
happening. This adds a separate step for getting the image first before
continuing on to the cluster creation.

Also cleans up some log handling while we're at it.

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/cluster/cluster.go
+++ b/cli/cmd/plugin/standalone-cluster/cluster/cluster.go
@@ -33,6 +33,9 @@ type ClusterManager interface {
 	Get(clusterName string) (*KubernetesCluster, error)
 	// Delete will destroy a cluster or return an error indicating a problem.
 	Delete(c *config.StandaloneClusterConfig) error
+	// Prepare will fetch an image or perform any pre-steps that can be done
+	// prior to actually creating the cluster.
+	Prepare(c *config.StandaloneClusterConfig) error
 }
 
 // NewClusterManager provides a way to dynamically get a cluster manager based on the standalone cluster config provider
--- a/cli/cmd/plugin/standalone-cluster/cluster/kind.go
+++ b/cli/cmd/plugin/standalone-cluster/cluster/kind.go
@@ -88,6 +88,16 @@ func (kcm KindClusterManager) Delete(c *config.StandaloneClusterConfig) error {
 	return provider.Delete(c.ClusterName, "")
 }
 
+// Prepare will fetch a container image to the cluster host.
+func (kcm KindClusterManager) Prepare(c *config.StandaloneClusterConfig) error {
+	cmd := exec.Command("docker", "pull", c.NodeImage)
+	_, err := exec.Output(cmd)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 // patchForAntrea modifies the node network settings to allow local routing.
 // this needs to happen for antrea running on kind or else you'll lose network connectivity
 // see: https://github.com/antrea-io/antrea/blob/main/hack/kind-fix-networking.sh
--- a/cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
+++ b/cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
@@ -546,6 +546,14 @@ func runClusterCreate(scConfig *config.StandaloneClusterConfig) (*cluster.Kubern
 	}
 
 	clusterManager := cluster.NewClusterManager(scConfig)
+
+	log.Style(outputIndent, logger.ColorLightGrey).Info("Pulling base image...\n")
+	err := clusterManager.Prepare(scConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Style(outputIndent, logger.ColorLightGrey).Info("Creating cluster...\n")
 	kc, err := clusterManager.Create(scConfig)
 	if err != nil {
 		return nil, err
--- a/cli/cmd/plugin/standalone-cluster/cluster/noop.go
+++ b/cli/cmd/plugin/standalone-cluster/cluster/noop.go
@@ -36,3 +36,8 @@ func (ncm NoopClusterManager) Get(clusterName string) (*KubernetesCluster, error
 func (ncm NoopClusterManager) Delete(c *config.StandaloneClusterConfig) error {
 	return nil
 }
+
+// Prepare doesn't perform any preparation steps before cluster creation.
+func (ncm NoopClusterManager) Prepare(c *config.StandaloneClusterConfig) error {
+	return nil
+}


cli/cmd/plugin/standalone-cluster/log/log.go
```diff
@@ -243,7 +237,7 @@ func (l *CMDLogger) Style(indent int, color string) Logger {
 // unquoteCodePoint takes the unicode value of a symbol and makes it usable for printing.
 func unquoteCodePoint(s string) string {
 	r, _ := strconv.ParseInt(strings.TrimPrefix(s, "\\U"), 16, 32)
-	return string(r)
+	return fmt.Sprintf("%-2s", string(r))
```
joshrosso, line 240: This approach won't work because emojis have variable width size. Some taking up 2 columns vs the standard 1.

See the spacing inconsistency:

image

stmcginnis: Yeah, it's the same issue we have now where we put in spaces and have to hope it formats OK. Maybe we should vet our emoji selection to make sure they are consistent?

cli/cmd/plugin/standalone-cluster/cluster/cluster.go
```diff
@@ -33,6 +33,9 @@ type ClusterManager interface {
 	Get(clusterName string) (*KubernetesCluster, error)
 	// Delete will destroy a cluster or return an error indicating a problem.
 	Delete(c *config.LocalClusterConfig) error
+	// PullImage will fetch an image. This can be used to prefetch before calling
+	// create in order to get better feedback on the current activity.
+	PullImage(imageName string) error
```
joshrosso, line 38: Should we consider a more generic API here? Perhaps:

```go
Prepare(c *config.LocalClusterConfig) error
```

My fear with specific APIs like `PullImage` is that as more specific edge cases grow up, so does our interface. With will then mean that work will be required for any future provider?

cc @jpmcb

cli/cmd/plugin/standalone-cluster/log/log.go
jpmcb: +1 for vetting emoji selection. I'm also curious how kind does this since they seem to use emojis of consistent width throughout their output

cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
```diff
@@ -171,8 +171,14 @@ func (t *TanzuLocal) Deploy(lcConfig *config.LocalClusterConfig) error {
 	log.Event("\\U+1F4E6", "Selected kapp-controller image bundle\n")
 	log.Style(outputIndent, logger.ColorLightGrey).Infof("%s\n", t.kappControllerBundle.GetRegistryURL())
 
-	// 4. Create the cluster
-	log.Eventf("\\U+1F6F0", " Creating cluster %s\n", lcConfig.ClusterName)
+	// 4. Create the cluster)
+	log.Eventf("\\U+1F4E5", "Pulling base image %s\n", lcConfig.NodeImage)
```
joshrosso, line 175: Can we make this log message more consistent to the rest?

image

My intuition is something like:

```txt
📥  Pulling base image 
      projects.registry.vmware.com/tce/kind/node:v1.21.5
      Pulling..... <--- these tick
```

Alternatively, I also think this could go very well under the `Creating cluster` log message:

```txt
🛰 Creating cluster doggo
     Pulling image.... <-- these tick
     To troubleshoot, use:
     kubectl ${COMMAND} --kubeconfig /home/josh/.config/tanzu/tkg/local/doggo/kube.conf
```

joshrosso (changes requested):
@stmcginnis huge UX improvement, ty!

Requested changes throughout.

Regarding emoji log processing, probably best to mutate the log package is a diff PR.

cli/cmd/plugin/standalone-cluster/cluster/cluster.go
jpmcb: I agree - a `Prepare` method in the `ClusterManager` interface would fit nicely, that way, other programmer "users" implementing this would be able to define how the management of their k8s clusters can be prepared. 

If we pass the `LocalClusterConfig`, then we can introspect the node image and download it early for the kind provider

cli/cmd/plugin/standalone-cluster/log/log.go
joshrosso: Sounds good, can we do that and make changes the logging package in a diff PR?

Also, is there a clever way to determine the width?

joshrosso: > how kind does this since they seem to use emojis of consistent width throughout their output

I think kind prints emojis at the end of the line? Which means you probably don't notice this.

jpmcb: > I think kind prints emojis at the end of the line? Which means you probably don't notice this.

Ah! Good way to get around this 😂

I can take a look at the emoji / unicode stuff and see if there's a good way to approach this

stmcginnis: Yep, was just about to remove that part from this PR since it's really a different concern than the main point of this one. We can figure that out later.

I tried looking at rune width, but they all are the same. And play.golang.org prints them out with a consistent width, so I think this has more to do with the terminal than having a variable data width.

If we care stongly enough, we could maybe use a tab writer to force more consistency, but not sure if we could make the width small enough for that not to look weird.

Something to investigate later, at any rate.

cli/cmd/plugin/standalone-cluster/cluster/cluster.go
stmcginnis: That makes sense to me.

The only challenge then is the console output. Since the CLI code is just asking the `ClusterManager` to prepare for cluster creation, we can't really be specific about "Pulling image" since that may not be relevant for the given manager (like the noop one). Any suggestions on what we should show as output for this step?

cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
stmcginnis: I like the idea of just having it under the "Creating cluster" step. 👍

cli/cmd/plugin/standalone-cluster/cluster/kind.go
```diff
@@ -88,6 +88,16 @@ func (kcm KindClusterManager) Delete(c *config.LocalClusterConfig) error {
 	return provider.Delete(c.ClusterName, "")
 }
 
+// PullImage will fetch a container image to the cluster host.
+func (kcm KindClusterManager) PullImage(imageName string) error {
+	cmd := exec.Command("docker", "pull", imageName)
```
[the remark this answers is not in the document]
stmcginnis: Yeah, that was one of my concerns with this. I'm glad you brought it up on here.

I _think_ we are OK. Kind will interact with either docker or podman, so without this (and the antrea patching), we should work just fine with either container engine.

I figured this use would be OK since we do have the antrea code also calling out, but ideally I think we wouldn't want any exec calls here. But one saving thing in all this is I think many deployments `alias docker=podman`, and for most operations they are compatible. So if they are not already doing that, it would be a fairly easy workaround to get this step working.

I'm open to just closing this PR for now too. We could wait and see if this is a big enough issue that we need to take extra steps like this. Ideally we just make the calls out to kind and leave it at that.

cli/cmd/plugin/standalone-cluster/cluster/cluster.go
joshrosso: This is a really solid and valid concern. We need to talk today about better ways to propagate implementations of our Logger interface into lower-level packages such that they can propagate details around progress, which can vary per-provider.

Another option is to provide channels where we pass messages around, while neat architecturally, I really don't dig when projects do this, it invokes too much complexity IMO when a simple, shared logger is wildly adequate.

cli/cmd/plugin/standalone-cluster/cluster/kind.go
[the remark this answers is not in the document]
joshrosso: Not necessarily. Keep in mind this implementation is specific to the `KindClusterManager`.

The reason I am OK with this, is that kind itself calls out to docker this way, thus the dependency for our `--provider=kind` does not change.

cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
stmcginnis: For now at least, not making the `...` animated to show progress due to the added complexity it would take to make these synchronous calls async. We can certainly do it though. If you'd like that done now I will get it added, but otherwise this will now at least show the line so there is some indication to the user that there is something being done.

jpmcb (approved):
Nice!! LGTM - will see if Josh has any final thoughts

cli/cmd/plugin/standalone-cluster/log/log.go
jpmcb: The 2 column emojis were handled in: https://github.com/vmware-tanzu/community-edition/pull/2420

cli/cmd/plugin/standalone-cluster/tanzu/tanzu.go
jpmcb: > the added complexity it would take to make these synchronous calls async.

I was thinking the same thing - it probably warrants a separate PR that creates go routines to log those animations based on a `context`

cli/cmd/plugin/standalone-cluster/log/log.go
stmcginnis: Yep, this was removed in the latest update. Looks much better in 2420.

joshrosso:
works great for me!

PR #2420, opened 2021-11-01
Standalone overhaul: Implements a set of standard 2 column width emojis for logging

jpmcb:
All changes related to the standalone cluster overhaul

- Implements a set of standard 2 column width emojis that can be used with the `Event` and `Eventf` logger methods.
- Adds breaks to each "event" as it's logged
- Removes logging for the list operation (which is more in line with what `kubectl get pods` might do)

Take a look!

Screen Shot 2021-11-01 at 4 25 06 PM


joshrosso:
Tried it locally, looks super good.

PR #2426, opened 2021-11-02
Use client-go rest.Inteface for all packaging interactions

joshrosso:
By using rest.Interface, we won't be in a place where client-go versions
are broken between us and our dependencies, in this case,
kapp-controller.

Signed-off-by: joshrosso <rossoj@vmware.com>

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/packages/packages.go
+++ b/cli/cmd/plugin/standalone-cluster/packages/packages.go
@@ -13,11 +13,6 @@ import (
 	kappapis "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
 	packaging "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/packaging/v1alpha1"
 	datapackaging "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apiserver/apis/datapackaging/v1alpha1"
-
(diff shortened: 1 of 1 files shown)


cli/cmd/plugin/standalone-cluster/packages/packages.go
```diff
@@ -312,7 +319,12 @@ func (am *PackageClient) CreateRootServiceAccount(ns, name string) (*v1.ServiceA
 }
 
 func (am *PackageClient) ListPackagesInNamespace(ns string) ([]datapackaging.Package, error) {
-	pkgList, err := am.pcs.DataV1alpha1().Packages(ns).List(context.TODO(), metav1.ListOptions{})
+	pkgList := &datapackaging.PackageList{}
+	err := am.arc.Get().
```
jpmcb, line 323: Pulled this PR down and I'm getting the following compile error:
```
github.com/vmware-tanzu/community-edition/cli/cmd/plugin/standalone-cluster/packages
packages/packages.go:323:11: am.arc undefined (type *PackageClient has no field or method arc)
```

did this actually need to be `aggRestClient`?:
```diff
- err := am.arc.Get().
+ err := am.aggRestClient.Get().
```

joshrosso: oops! very weird 🤔

PR #2446, opened 2021-11-04
Plumb standalone cluster network arguments

stmcginnis:
This plumbs through the --pod-cidr and --service-cidr CLI arguments to
kind cluster creation. Input validation still needs to be done, but
assuming the correct CIDR format is used, clusters will be set up with
the corresponding pod and service subnets.

This also gets things in place to be able to perform other future kind
configuration customizations by populating the kind cluster struct
rather than parsing a static string.

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/cluster/kind.go
+++ b/cli/cmd/plugin/standalone-cluster/cluster/kind.go
@@ -4,33 +4,38 @@
 package cluster
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"regexp"
 	"strings"
 
-	kindCluster "sigs.k8s.io/kind/pkg/cluster"
+	"gopkg.in/yaml.v3"
+	kindconfig "sigs.k8s.io/kind/pkg/apis/config/v1alpha4"
+	kindcluster "sigs.k8s.io/kind/pkg/cluster"
 	"sigs.k8s.io/kind/pkg/exec"
 
 	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/standalone-cluster/config"
 )
 
-const defaultKindConfig = `kind: Cluster
-apiVersion: kind.x-k8s.io/v1alpha4
-nodes:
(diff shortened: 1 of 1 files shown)


stmcginnis:
Pod CIDR setting can be verified with `kubectl get nodes -o jsonpath='{.items[*].spec.podCIDR}'` and service CIDR can be seen with `kubectl get svc -A`, and I'm sure 20 other possible ways.

cli/cmd/plugin/standalone-cluster/cluster/kind.go
```diff
@@ -77,14 +84,50 @@ func (kcm KindClusterManager) Create(c *config.StandaloneClusterConfig) (*Kubern
...
+		kindConfig.Nodes[i].Image = c.NodeImage
+	}
+
+	// Marshal it into the raw bytes we need for creation
+	var rawConfig bytes.Buffer
+	yamlEncoder := yaml.NewEncoder(&rawConfig)
+
+	err := yamlEncoder.Encode(kindConfig)
+	if err != nil {
+		return []byte{}, fmt.Errorf("failed to generate Kind configuration. Error: %s", err.Error())
```
joshrosso, line 113: should we return `nil` here?

cli/cmd/plugin/standalone-cluster/cluster/kind.go
```diff
@@ -77,14 +84,50 @@ func (kcm KindClusterManager) Create(c *config.StandaloneClusterConfig) (*Kubern
...
+	}
+	err = yamlEncoder.Close()
+	if err != nil {
+		return []byte{}, fmt.Errorf("failed to generate Kind configuration. Error: %s", err.Error())
```
joshrosso, line 117: should we return `nil` here?

joshrosso (changes requested):
works great for me. tiny nit: Thinking we should return nil when config fails to parse. Outside of that, good to go!

PR #2486, opened 2021-11-09
Fix e2e tests for external-dns v0.10.0

christianang:
Previously they were failing because the external-dns-values file was
missing (it was being gitignored so it was not obvious).

Signed-off-by: Christian Ang <angc@vmware.com>
Co-authored-by: Edwin Xie <exie@vmware.com>

The external-dns e2e tests were missing the values file. Looks like we missed committing it because it was gitignored. We renamed the file to prevent missing the file again.

```release-note
NONE
```

Fixes: n/a

Ran the tests and ensured it was using the correct filename.


adobley (approved):
LGTM

PR #2489, opened 2021-11-10
Fix cleanup of kubeconfig entry for vSphere, AWS and Azure E2E tests

### LINKED ISSUE #2481: Fix cleanup of the kubeconfig users, clusters and contexts
This has to be ideally done for all tests (AWS, Azure, Docker etc) to ensure that the test data in the kubeconfig is cleaned up or a temp kubeconfig is used all the time, either one of those. We are noticing the side effects of this not being done in the vSphere E2E tests as the pipeline uses the same GitHub Actions self hosted Runner (that we host) all the time, unlike GitHub hosted Runners, and the side effect can also be seen in the local / dev environment when running the automation scripts in local. More details can be found in #2454 and #2460

karuppiah7890:

Fixes cleanup of kubeconfig entry for vSphere, AWS and Azure E2E tests

```release-note
NONE
```

Fixes: #2481 

--

In terms of implementation, I had two ideas -
- One was to individually use `kubconfig_cleanup` function in all places where it was needed. This caused a bit of repetition at times. This is the current implemenation
- Another was to use `kubeconfig_cleanup` inside existing cleanup functions like `govc_cleanup`, `aws-nuke-tear-down`, `azure_cluster_cleanup`. This causes very low repetition and favors DRY. I didn't do this though, as the name `govc_cleanup` or `aws-nuke-tear-down` suggested some implementation details that it uses `govc` or `aws-nuke` and if I add `kubeconfig_cleanup` function usage to that, it didn't make sense. `azure_cluster_cleanup` does make a generic name to say it's a cleanup of an Azure cluster, but I was still thinking it should contain only `az` (Azure CLI) related code. If we want to choose this way, I could go ahead and change the code accordingly and also maybe rename `govc_cleanup` to `vsphere_cluster_cleanup` or similar and `aws-nuke-tear-down` to `aws_cluster_cleanup` or similar and put `kubeconfig_cleanup` code in it

### DIFF
--- a/test/aws/deploy-tce-managed.sh
+++ b/test/aws/deploy-tce-managed.sh
@@ -31,13 +31,15 @@ function delete_management_cluster {
     echo "$@"
     export AWS_REGION="us-east-2"
     export CLUSTER_NAME="${MGMT_CLUSTER_NAME}"
-    tanzu management-cluster delete "${CLUSTER_NAME}" -y || { aws-nuke-tear-down "MANAGEMENT CLUSTER DELETION FAILED! Deleting the cluster using AWS-NUKE..." "${CLUSTER_NAME}"; }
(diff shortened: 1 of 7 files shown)


karuppiah7890:
Fixing the shellcheck lint errors now

stmcginnis (approved):
Looks good to me.

navidshaikh (approved):

PR #2492, opened 2021-11-10
removing two lines from the harbor values.yaml

kacole2:
Carvel fails to install Harbor because Error: package reconciliation failed: ytt: Error: Extracting data value from file based on the top two lines of #@. After removing #@ the install worked.

```
Error: package reconciliation failed: ytt: Error: Extracting data value from file:
  Checking data values file '/etc/kappctrl-mem-tmp/kapp-controller-template-values263829594/values.yaml':
    Expected to not find annotations inside data values file (hint: remove comments starting with '#@')
```

```release-note
Update of values.yaml template for harbor to install correctly according to the documentation.
```

m/a

Installed it on a cluster running in AWS.

n/a

### DIFF
--- a/addons/packages/harbor/2.2.3/bundle/config/values.yaml
+++ b/addons/packages/harbor/2.2.3/bundle/config/values.yaml
@@ -1,6 +1,3 @@
-#@data/values
-#@overlay/match-child-defaults missing_ok=True
-
 ---
 #! The namespace to install Harbor
 namespace: harbor


seemiller:
I ran `ytt -f addons/packages/harbor/2.2.3/bundle/config/` and got this error:

```shell
ytt: Error: Evaluating starlark template: 
- struct has no .namespace field or method
    in validate_harbor_namespace
      6 |   values.namespace or assert.fail("harbor namespace should be provided")
    in validate_harbor
      101 |      validate_func()
    in <toplevel>
      109 | validate_harbor()
```

The `namespace` is provided in the `values.yaml` file so it shouldn't be erroring on that.

Looking closer, I believe that the `values.yaml` file needs to have `#@data/values` at the top to indicate that it contains data values for use with the overlays.

addons/packages/harbor/2.2.3/bundle/config/values.yaml
```diff
@@ -1,6 +1,3 @@
-#@data/values
```
seemiller, line 1: I ran `ytt -f addons/packages/harbor/2.2.3/bundle/config/` and got this error:

```shell
ytt: Error: Evaluating starlark template: 
- struct has no .namespace field or method
    in validate_harbor_namespace
      6 |   values.namespace or assert.fail("harbor namespace should be provided")
    in validate_harbor
      101 |      validate_func()
    in <toplevel>
      109 | validate_harbor()
```

I think that you want to keep the `#@data/values` at the top here to indicate that this  is a data values file.

seemiller (changes requested):

addons/packages/harbor/2.2.3/bundle/config/values.yaml
kacole2: Have you tried installing it based on the documentation and not ytt? I'm going by what the documentation says on installing Harbor and that's the error I received.

jkjell: I agree with @kacole2 that this does fix the issue but, I don't think it's the right fix.  The other packages include those two lines in their values.yaml as well.

heww: I don't think this fix is correct. We will start to update the doc to resolve the problem when installing the harbor package today.

MinerYang: We've already updated the documentation, pls refer to the [#PR2648](https://github.com/vmware-tanzu/community-edition/pull/2648).

Issue #2453, opened 2021-11-05
Outdated / out of sync `go.mod` and `go.sum` files


karuppiah7890:
Creating this issue based on the conversations in #2167 (https://github.com/vmware-tanzu/community-edition/pull/2167#issuecomment-935675031) and the data in #2444 and few other similar PRs where we noticed that `go.mod` and `go.sum` files were outdated / out of sync.

Maybe to solve the problem of outdated `go.mod` and `go.sum` and ensuring we do not have such outdated Golang module files, we can update the `lint` target in our `Makefile` like this -

```
lint: verify-modules golangci-lint

.PHONY: modules
modules: ## Runs go mod to ensure proper vendoring.
	go mod tidy

.PHONY: verify-modules
verify-modules: modules
	@if !(git diff --quiet HEAD -- go.sum go.mod); then \
		echo "go module files are out of date"; exit 1; \
	fi

golangci-lint: ## Run Go code linting
ifeq ($(origin GOLANGCI_LINT),undefined)
	@echo "Error! GOLANGCI_LINT env var not set"
else
	$(GOLANGCI_LINT) run -v --timeout=5m
endif
```

This is similar to the Golang module files check that CAPZ repo does - 

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/ace988dec38dcdf27176a1d6f081cc64274fffb9/Makefile#L610-L614

https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/ace988dec38dcdf27176a1d6f081cc64274fffb9/Makefile#L286-L288

stmcginnis:
+1

The `tanzu-framework` repo had something similar to the verify-modules target (though I don't see it now). I think it would be good to add that check so we don't ignore these being out of date.

PR #2470, opened 2021-11-08
Run go mod tidy before verifying modules to lint modules

karuppiah7890:
## What this PR does / why we need it

Fixes a core issue that was missed as part of #2467 PR

- Run go mod tidy before verifying modules to lint modules
- Update the outdated `go.mod` and `go.sum` files

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
NONE
```

## Which issue(s) this PR fixes

Related - #2453 

Fixes #2469 - related issue and gets fixed as this PR updates all outdated `go.mod` and `go.sum` files to pass the lint checks. cc @chenliu1993 

## Describe testing done for PR

```bash
make verify-modules
```

OR

```bash
make lint
```

Run twice and check there are no `go.mod` and `go.sum` diffs in the git repo and no errors in the execution of the above command(s). Also `cd` into any go modules directory and do `go mod tidy` and check if there are any diffs in `go.mod` and `go.sum` files

## Special notes for your reviewer

--

PR #2498, opened 2021-11-10
Update get-deps to use "go mod tidy"

stmcginnis:

Most get-deps targets were using "go mod download", which isn't as
accurate as "go mod tidy" for evaluating dependencies. From one of the
go maintainers:

```
go mod download is downloading all of the modules in the dependency
graph, which it can determine from reading only the go.mod files. It
doesn't know which of those modules are actually needed to satisfy a
build, so it doesn't pull in the checksums for those modules (because
they may not be relevant).

On the other hand, go mod tidy has to walk the package graph in order
to ensure that all imports are satisfied. So it knows, for a fact,
without doing any extra work, that the modules it walks through are
needed for a go build in some configuration.
```

This also includes the updates to go.sum/go.mod based on the more
thorough dependency checks.

```release-note
NONE
```

Related: #2453

Ran local `make release` to get a full build.


stmcginnis:
Lost https://github.com/vmware-tanzu/community-edition/pull/2470 in all the linked issues and PRs and thought I must have been mis-remembering something. We should stick with that. :)

PR #2502, opened 2021-11-10
Fix ec2 instance leak when tagging fails

davidvonthenen:

This fixes a small ec2 instance leak when using self-hosted runners. This doesn't happen often (like once every day), but it does happen which leads to manual clean up for orphaned ec2 instances. If the tagging fails, delete the instance. Then (like normal) let the outer loop start the create-tag loop all over again.

```release-note
Fix ec2 instance leak when tagging fails
```

NA

NA

NA

### DIFF
--- a/hack/runner/webhook/aws.go
+++ b/hack/runner/webhook/aws.go
@@ -6,6 +6,7 @@ package main
 import (
 	"os"
 	"strings"
+	"time"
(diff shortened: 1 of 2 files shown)


jpmcb (approved):
Nice! A little delay never hurt!

Issue #2506, opened 2021-11-11
update DEV.md in harbor package 2.3.3


zyyw:
## Description

Update generate-manifests.sh in DEV.md under harbor package 2.3.3

## Location

the link page of the current page is this:
https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/harbor/2.3.3/DEV.md

PR #2507, opened 2021-11-11
update harbor 2.3.3 DEV.md

zyyw:
Signed-off-by: Shengwen Yu <yshengwen@vmware.com>

We update generate-manifests.sh within DEV.md within TCE harbor package 2.3.3 so that it can automatically generate the
desired upstream without extra manual operation.

```release-note
DONE
```

Fixes: #2506 

Generated harbor 2.3.3 upstream with the updated generate-manifests.sh, and they are the same as expected.
Only doc is updated not extra test is needed.

### DIFF
--- a/addons/packages/harbor/2.3.3/DEV.md
+++ b/addons/packages/harbor/2.3.3/DEV.md
@@ -21,12 +21,12 @@ The manifests from 02 to 11 are generated from [Harbor Helm Chart](https://githu
     rm -rf manifests
     mkdir -p manifests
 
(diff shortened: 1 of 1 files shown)


stmcginnis (dismissed):
This looks good, but just to note that all of the `sed` calls in this will fail if run on macOS.

If you want to follow up later on that, with the version of `sed` that comes on Macs, the command would need to switch from `sed -i` to `sed -i '' -e`.

heww (approved):

PR #2510, opened 2021-11-11
Collect diagnostics data from E2E pipelines for vSphere, Azure and AWS

### LINKED ISSUE #1443: Automatically capture debugging / diagnotic data from failed E2E test pipelines
## Feature Request

We currently have a lot of failed E2E test pipelines, refer #1355 , #1245 , #1324 . We are currently not capturing any data from these pipelines to help us debug the issues

Currently how we debug pipelines is - try to reproduce the error in local or fork repo pipeline and debug the issue using local terminal if running in local, or by using debug scripts like [this](https://github.com/vmware-tanzu/community-edition/blob/dave/delete-me/.github/workflows/e2e-package-external-dns.yaml#L60-L131) or [this](https://github.com/karuppiah7890/tce/blob/debug-tce-install/test/debug-tce-install.sh)

So, basically, the current pipelines don't capture debugging data for debugging purposes, and developers do extra stuff to debug the pipeline running E2E or debug the local E2E test when running in local, surely debugging in local being easier, but debugging in local doesn't always as pipeline environment is not the same as local environment, there are lot of differences in terms of - hardware resources (CPU, RAM, Disk) and also software (OS version, Kernel version, shared libraries, other installed softwares)

The benefits of capturing data from the pipelines are -
- We know what exactly happened in the pipeline, if we are looking at the pipeline after the pipeline fails and is finished executing and also gets rid of all the kind bootstrap clusters along with runners and also cleans up cloud resources to avoid paying unnecessarily for unused temporary cloud resources. Since everything will be gone, we won't have a way to capture the data we need
- We get more and more data around our pipeline failures. This helps us find patterns, debug and fix issues
- Helps us get data for one off failures too. For example, this failure was noticed only once in our pipelines - https://github.com/vmware-tanzu/community-edition/runs/3405792810?check_suite_focus=true#step:5:1554

For all the reasons described above, it's important to capture all possible data from failed pipelines and failed processes before the pipeline and all the resources are cleaned up and gone

What kind of data do we capture? Everything that seems relevant, and almost every data we can get our hands on. This way, we get all the data and looks for answers to questions we don't even know or we aren't even asking. Or else we would only capture data for questions we already know while debugging and are asking

The data that we surely need, around which we will surely ask questions while debugging - 
- Logs of all containers inside all pods in kind bootstrap cluster and also the created cluster / cluster that is going to be deleted. We especially need the logs of CAPI controller pods but other pod logs are important too, for example static pods like kube-proxy and others that make up the control plane components. We can get this from `kubectl` using `kubectl logs` or exec into the control plane and worker nodes and get the container logs from low level from the container runtime, containerd, using `crictl`
- For every pod, capture `kubectl describe` data too, to get events data
- For every CRD, capture `kubectl describe` data
- For every K8s resource, capture `yaml` resource file maybe? But yeah, most details would be present in `kubectl describe` itself. Something to think about it
- Logs of the processes running in the kind bootstrap cluster nodes. In kind, nodes = Docker containers. We need to exec into the docker containers and get logs of processes like kubelet and maybe also any other process that are of interest. Something to check out. Logs can be obtained from particular process's log files or from `journalctl` or from `syslog` log file

The above is for cluster creation and cluster deletion too

We can think more about what other data we can pull as well, so that we have all the data we need to debug failed pipelines

No alternatives that I know of. We have to either use an existing solution like Tanzu Diagnostics plugin or build a solution for this

## Additional context

There is a Tanzu Diagnostics plugin that is going to come up soon. I think it takes care of capturing all the data we need? Not sure though. PR for the plugin #1369

karuppiah7890:

Collects diagnostics data from E2E pipelines for vSphere, Azure and AWS

```release-note
Collect diagnostics data from E2E pipelines for vSphere, Azure and AWS
```

Fixes: #1443 

E2E Azure Standalone Cluster Test failure collecting diagnostics and uploading it to GitHub Actions - https://github.com/karuppiah7890/community-edition/runs/4190341648?check_suite_focus=true . The tar ball can be seen at `diagnostics-data` in https://github.com/karuppiah7890/community-edition/actions/runs/1453280816

E2E Azure Management and Workload Cluster Test  failure collecting diagnostics and uploading it to GitHub Actions - https://github.com/karuppiah7890/community-edition/runs/4190341651?check_suite_focus=true. The tar ball can be seen at `diagnostics-data` in https://github.com/karuppiah7890/community-edition/actions/runs/1453280810

One tricky thing is, there is a problem in one scenario

- Create management cluster
- It's created. Now run tests on the management cluster
- One of the tests fail, for example a test verifying TCE package installation. Automatically collect diagnostics data about the management cluster and store it in `management-cluster.<cluster-name>.diagnostics.tar.gz`. This data will help in understanding the test failure in TCE package installation
- Delete the cluster using `tanzu`
- Cluster deletion fails. Automatically collect diagnostics data about the management cluster and store it in `management-cluster.<cluster-name>.diagnostics.tar.gz`. This data will help in understanding the failure in deleting the management cluster - it could be some issue in the kind bootstrap cluster - some CAPI controller didn't work, or maybe some credentials issue - for example the token used to interact with the cloud APIs expired, for example Azure credentials

Note that - in such a scenario, we will end up with collecting diagnostics twice but the second tar ball will overwrite the first tar ball, as they both use the same name for the tar ball. Also, such a situation could happen to standalone clusters too

### DIFF
--- a/test/util/utils.sh
+++ b/test/util/utils.sh
@@ -27,3 +27,64 @@ function kubeconfig_cleanup {
     kubectl config delete-user "${cluster_name}-admin" || true
     kubectl config delete-cluster "${cluster_name}" || true
 }
+
+function collect_standalone_cluster_diagnostics {
+    cluster_infra=$1
+    cluster_name=$2
+
+    if [[ -z "${cluster_infra}" ]]; then
+        error "Cluster infra (vsphere, azure, aws, docker) not passed to collect_standalone_cluster_diagnostics function. Ignoring error. Usage example: collect_standalone_cluster_diagnostics vsphere test-cluster-1234"
+    fi
+
+    if [[ -z "${cluster_name}" ]]; then
+        error "Cluster name not passed to collect_standalone_cluster_diagnostics function. Ignoring error. Usage example: collect_standalone_cluster_diagnostics vsphere test-cluster-1234"
+    fi
+
+    echo "Collecting ${cluster_name} standalone cluster diagnostics data"
+
+    tanzu diagnostics collect --workload-cluster-infra "${cluster_infra}" \
+        --workload-cluster-name "${cluster_name}" || {
+        error "There was an error collecting tanzu diagnostics data. Ignoring the error"
+    }
+}
+
+function collect_management_cluster_diagnostics {
+    cluster_name=$1
+
+    if [[ -z "${cluster_name}" ]]; then
+        error "Cluster name not passed to collect_management_cluster_diagnostics function. Ignoring error. Usage example: collect_management_cluster_diagnostics management-cluster-1234"
+    fi
+
+    echo "Collecting ${cluster_name} management cluster diagnostics data"
+
+    tanzu diagnostics collect --management-cluster-name "${cluster_name}" || {
+        error "There was an error collecting tanzu diagnostics data. Ignoring the error"
+    }
+}
+
+function collect_management_and_workload_cluster_diagnostics {
+    cluster_infra=$1
+    management_cluster_name=$2
+    workload_cluster_name=$3
+
+    if [[ -z "${cluster_infra}" ]]; then
+        error "Workload cluster infra (vsphere, azure, aws, docker) not passed to collect_management_and_workload_cluster_diagnostics function. Ignoring error. Usage example: collect_management_and_workload_cluster_diagnostics vsphere management-cluster-1234 workload-cluster-1234"
+    fi
+
+    if [[ -z "${management_cluster_name}" ]]; then
+        error "Management cluster name not passed to collect_management_and_workload_cluster_diagnostics function. Ignoring error. Usage example: collect_management_and_workload_cluster_diagnostics vsphere management-cluster-1234 workload-cluster-1234"
+    fi
+
+    if [[ -z "${workload_cluster_name}" ]]; then
+        error "Management cluster name not passed to collect_management_and_workload_cluster_diagnostics function. Ignoring error. Usage example: collect_management_and_workload_cluster_diagnostics vsphere management-cluster-1234 workload-cluster-1234"
+    fi
+
+    echo "Collecting ${management_cluster_name} management cluster and ${workload_cluster_name} workload cluster diagnostics data"
+
+    tanzu diagnostics collect --bootstrap-cluster-skip \
+        --management-cluster-name "${management_cluster_name}" \
+        --workload-cluster-infra "${cluster_infra}" \
+        --workload-cluster-name "${workload_cluster_name}" || {
+        error "There was an error collecting tanzu diagnostics data. Ignoring the error"
+    }    
+}
--- a/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
+++ b/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
@@ -49,3 +49,13 @@ jobs:
 
           go env -w GOFLAGS=-mod=mod
           make aws-management-and-workload-cluster-e2e-test
+
+      - name: Collect tanzu diagnostics data
+        uses: actions/upload-artifact@v2
+        if: ${{ always() }}
+        with:
+          name: diagnostics-data
+          path: |
+            bootstrap.*.diagnostics.tar.gz
+            management-cluster.*.diagnostics.tar.gz
+            workload-cluster.*.diagnostics.tar.gz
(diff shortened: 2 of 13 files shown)


test/util/utils.sh
```diff
@@ -27,3 +27,70 @@ function kubeconfig_cleanup {
     kubectl config delete-user "${cluster_name}-admin" || true
     kubectl config delete-cluster "${cluster_name}" || true
 }
+
+function collect_standalone_cluster_diagnostics {
```
karuppiah7890, line 31: When any issue occurs during collection of diagnostics data - do we still go ahead and cleanup the cluster? This would mean that we might not be able to debug the failure that happened. Or should we leave the cleanup of the cluster as is so that we can debug and get diagnostics data and then clean up the cluster? As of now, it does not cleanup the cluster. For example - https://github.com/karuppiah7890/community-edition/runs/4189838503?check_suite_focus=true#step:5:400 it abruptly stops when there is an issue (in this case bug in script) in collecting diagnostics.

test/util/utils.sh
```diff
@@ -27,3 +27,70 @@ function kubeconfig_cleanup {
...
+    fi
+
+    if [[ -z "${cluster_name}" ]]; then
+        echo "Cluster name not passed to collect_standalone_cluster_diagnostics function. Usage example: collect_standalone_cluster_diagnostics vsphere test-cluster-1234"
+        return 1
+    fi
+
+    echo "Collecting ${cluster_name} standalone cluster diagnostics data"
+
+    tanzu diagnostics collect --workload-cluster-infra "${cluster_infra}" \
```
karuppiah7890, line 47: There are just a few things we run in the diagnostics collect function, mainly the `tanzu diagnostics collect` command runs and if that fails - what do we do is the question

test/util/utils.sh
dvonthenen: If this script is being used for workflows (which I believe it is), we should probably still clean up the environment in order to prevent orphaned clusters requiring manual intervention to clean and also possibly these stacking up an eating resource which (depending on the environment) might cause other failures due to lack of resources.

test/util/utils.sh
dvonthenen: If we fail to get logs, it's no better than it is today. We should just display the error for someone to attempt to address why the logs aren't being collected.

davidvonthenen (commented):
This is great! I checked out the workflow runs, but don't see where you can download the tarball. I'm probably just blind and am overlooking where.

Also, we need to make sure there isn't any sensitive information be collected in the tarball (tokens, passwords, etc). If there are, we might need to redact them or place the log files behind some access barriers.

test/util/utils.sh
karuppiah7890: Cool ! Then I can swallow any errors in command using `|| true`

test/util/utils.sh
karuppiah7890: Cool, then I'll add a `|| true` to absorb any errors in the log collection command

karuppiah7890:
> This is great! I checked out the workflow runs, but don't see where you can download the tarball. I'm probably just blind and am overlooking where.

In my fork's test run, it's here - https://github.com/karuppiah7890/community-edition/actions/runs/1453280816 at the bottom of the page under `Artifacts`

> Also, we need to make sure there isn't any sensitive information be collected in the tarball (tokens, passwords, etc). If there are, we might need to redact them or place the log files behind some access barriers.

Ah yes. I did think about this. I checked the logs, I didn't see any tokens being mentioned there. I did see IP addresses being printed out in some cases - like in vSphere E2E. We currently put the IP addresses in Secrets even though they are private IPs. The IP addresses are still a config, but not present in plain text in the worfklow config yaml. Apart from that I didn't notice any such issues. I think the CAPI projects have good standards in not logging secrets like tokens

dvonthenen:
> https://github.com/karuppiah7890/community-edition/actions/runs/1453280816

I totally miss it at the bottom. I think my browser was scrunched and didn't end up scrolling down.

LGTM!

davidvonthenen (dismissed):

dvonthenen:
Looks like we need a rebase

karuppiah7890:
Rebased with `main` 👍

davidvonthenen (approved):

dvonthenen:
Awesome!

PR #2246, opened 2021-10-13
Adding shell scripts tests in e2e-framework

aman556:
## What this PR does / why we need it

This PR have two motive in e2e-framework.
- First it will add shell scripts to test :
   - Fluent-bit
   - cert-manager
   - contour
   - local-path-storage
## Details for the Release Notes (PLEASE PROVIDE)

```release-note
e2e-framework can test the packages like contour, fluent-bit, cert-manager and local-path-storage. No change in the
commands in e2e-framework for testing these packages. 
e2e-framework will be also add as a plugin in github workflow.
```

## Which issue(s) this PR fixes

Fixes: [#2243](https://github.com/vmware-tanzu/community-edition/issues/2243) 

## Describe testing done for PR

Run `ginkgo -v -- --packages="" --version="" --guest-cluster-name=""` to check the working of shell scripts.
## Special notes for your reviewer

The bash scripts to test packages follow these steps:
- Running `tanzu package installed list` to check the package which we are testing is already installed or not.
- Running `tanzu package available list contour.community.tanzu.vmware.com` for fetching the latest version of that package.
- Running `tanzu package install pkgName --package-name pkgName.community.tanzu.vmware.com -f config.yaml --version "${version}"` to install the package.
- Taking care of all the prerequisites.
- Ensuring that the package is working fine.
- Deleting all the prerequisites and the package.

PR #2517, opened 2021-11-11
Improvements in smoke tests

aman556:

This PR is a follow-up PR to [#2246](https://github.com/vmware-tanzu/community-edition/pull/2246).The changes in this PR are:
- Smoke tests under `test/smoke` are improved as per the review comments in the last PR.
- Gatekeeper test which is in `test/gatekeeper` is hooked in the `test/smoke` 

```release-note
These are the improved smoke tests for the sanity test of the packages.
```

Fixes: # NA

`make smoke-test` is triggered to run all the smoke tests. 

Please refer to [#2246](https://github.com/vmware-tanzu/community-edition/pull/2246) for complete picture.

### DIFF
--- a/test/smoke/Makefile
+++ b/test/smoke/Makefile
@@ -4,7 +4,7 @@
 ROOT_DIR := $(shell git rev-parse --show-toplevel)
 MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-smoke-test: install-ginkgo  cert-manager contour fluent-bit local-path-storage external-dns multus-cni 
+smoke-test: install-ginkgo  cert-manager fluent-bit local-path-storage contour gatekeeper external-dns multus-cni 
 
 install-ginkgo:
 	go get github.com/onsi/ginkgo/ginkgo
@@ -27,3 +27,6 @@ external-dns:
 
 multus-cni:
 	cd "${ROOT_DIR}"/test/e2e && ginkgo -v -- --packages="multus-cni" --version="3.7.1" --guest-cluster-name="tce.public"
+
+gatekeeper:
+	./"${ROOT_DIR}"/test/gatekeeper/e2e-test.sh && tanzu package installed delete gatekeeper -y
--- a/test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
+++ b/test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
@@ -1,47 +1,45 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
 # Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
 
 
 
 # Checking package is installed or not
-tanzu package installed list | grep "cert-manager.community.tanzu.vmware.com"
 
-exitcode=$?
-
-if [ "${exitcode}" == 1 ] 
-then
-version=$(tanzu package available list cert-manager.community.tanzu.vmware.com | tail -n 1 | awk '{print $2}')
-tanzu package install cert-manager --package-name cert-manager.community.tanzu.vmware.com --version "${version}"
-sleep 30s
-fi
+tanzu package installed list | grep "cert-manager.community.tanzu.vmware.com" || {
+  version=$(tanzu package available list cert-manager.community.tanzu.vmware.com | tail -n 1 | awk '{print $2}')
+  tanzu package install cert-manager --package-name cert-manager.community.tanzu.vmware.com --version "${version}"
+}
 
 
 # Providing prerequisite 
+
+NAMESPACE_SUFFIX=${RANDOM}
+NAMESPACE="cert-manager-${NAMESPACE_SUFFIX}"
+kubectl create ns ${NAMESPACE}
+
 cat <<EOF | kubectl apply --filename -
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
   name: self-signed
+  namespace: ${NAMESPACE}
 spec:
   selfSigned: {}
 EOF
 
-cat <<EOF | kubectl apply --filename -
-apiVersion: cert-manager.io/v1
-kind: Issuer
-metadata:
-  name: private-ca
-spec:
-  ca:
-    secretName: private-ca
-EOF
 
 cat <<EOF | kubectl apply --filename -
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: private-ca
+  namespace: ${NAMESPACE}
 spec:
   isCA: true
   duration: 2160h
@@ -58,10 +56,22 @@ EOF
 
 cat <<EOF | kubectl apply --filename -
 apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: private-ca
+  namespace: ${NAMESPACE}
+spec:
+  ca:
+    secretName: private-ca
+EOF
+
+
+cat <<EOF | kubectl apply --filename -
+apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: example-com
-  namespace: default
+  namespace: ${NAMESPACE}
 spec:
   secretName: example-com-tls
   issuerRef:
@@ -73,17 +83,19 @@ spec:
     - www.example.com
 EOF
 
-sleep 45s
 
-status="$(kubectl get certificate/example-com | tail -n 1 | awk '{print $2}')"
-echo "${status}"
+echo "Waiting for certificate/example-com to Ready..."
+kubectl wait --for=condition=Ready certificate/example-com -n ${NAMESPACE} --timeout=300s
+status="$(kubectl get certificate/example-com -n ${NAMESPACE} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}')"
+
 tanzu package installed delete cert-manager -y
+kubectl delete ns ${NAMESPACE}
 
 if [ "${status}" != "True" ]
 then 
     echo "certificate status is not true"
-    echo "cert-manager failed"
+    printf '\E[31m'; echo "cert-manager failed"; printf '\E[0m'
     exit 1
 else
-    echo "cert-manager Passed"
+    printf '\E[32m'; echo "cert-manager Passed"; printf '\E[0m'
 fi
--- a/test/smoke/packages/contour/1.18.2/contour-test.sh
+++ b/test/smoke/packages/contour/1.18.2/contour-test.sh
@@ -3,31 +3,34 @@
 # Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+set -o errexit
+set -o nounset
+set -o pipefail
 
+MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 # Checking package is installed or not
-tanzu package installed list | grep "contour.community.tanzu.vmware.com"
-exitcode_contour=$?
-MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-echo "${MY_DIR}"
-if [ "${exitcode_contour}" == 1 ] 
-then
+
+tanzu package installed list | grep "contour.community.tanzu.vmware.com" || {
     version=$(tanzu package available list contour.community.tanzu.vmware.com | tail -n 1 | awk '{print $2}')
     tanzu package install contour --package-name contour.community.tanzu.vmware.com -f "${MY_DIR}"/contour-values.yaml --version "${version}"
-fi
-sleep 15s
+}
+
 
 # Providing prerequisite 
-kubectl create namespace contour-example-workload
-kubectl create deployment nginx-example --image nginx --namespace contour-example-workload
-kubectl create service clusterip nginx-example --tcp 80:80 --namespace contour-example-workload
+
+NAMESPACE_SUFFIX=${RANDOM}
+NAMESPACE="contour-${NAMESPACE_SUFFIX}"
+kubectl create ns ${NAMESPACE}
+kubectl create deployment nginx-example --image nginx --namespace ${NAMESPACE}
+kubectl create service clusterip nginx-example --tcp 8080:80 --namespace ${NAMESPACE}
 
 kubectl apply -f - <<EOF
 apiVersion: projectcontour.io/v1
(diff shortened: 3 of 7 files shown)


test/smoke/Makefile
```diff
@@ -27,3 +27,6 @@ external-dns:
 
 multus-cni:
 	cd "${ROOT_DIR}"/test/e2e && ginkgo -v -- --packages="multus-cni" --version="3.7.1" --guest-cluster-name="tce.public"
+
+gatekeeper:
+	cd "${ROOT_DIR}"/test/gatekeeper && ./e2e-test.sh && tanzu package installed delete gatekeeper -y
```
karuppiah7890, line 32: I think we can avoid using `cd` here. Something like this

```suggestion
	./"${ROOT_DIR}"/test/gatekeeper/e2e-test.sh && tanzu package installed delete gatekeeper -y
```

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
```diff
@@ -73,17 +82,18 @@ spec:
     - www.example.com
 EOF
 
-sleep 45s
+timeout=300
+while [[ $(kubectl get certificate/example-com -n ${NAMESPACE} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" && ${timeout} -ne 0 ]]; do echo "waiting for certificate" && sleep 1 && timeout=${timeout}-1 ; done
```
karuppiah7890, line 86: We can add some new lines here to avoid a very long single line. And if possible capture the `kubectl get` command in a variable or function and use it to keep it shorter and easier to understand what's going on

Also, we can have a bit more of `sleep` time to avoid checking every 1 second and printing to wait, as that will be a lot of printing but yeah, it will be a quick check

On a side note, I was thinking if this whole while loop along with body can be replaced with a `kubectl wait`, I think it could be and it's more easier to read and reason out `kubectl wait` and lesser code to maintain. So if that's possible, I would prefer it any day

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
```diff
@@ -73,17 +82,18 @@ spec:
...
 
-status="$(kubectl get certificate/example-com | tail -n 1 | awk '{print $2}')"
-echo "${status}"
-tanzu package installed delete cert-manager -y
+status="$(kubectl get certificate/example-com -n ${NAMESPACE} | tail -n 1 | awk '{print $2}')"
```
karuppiah7890, line 88: We could use `kubectl get certificate/example-com -n ${NAMESPACE} -o jsonpath={...}` here too, similar to https://github.com/vmware-tanzu/community-edition/pull/2517/files#diff-693b938505f694eb48282d5b45e5c3bd1e6e17363b0a7b9e7198b5fdf9115e18R86

test/smoke/packages/contour/1.18.2/contour-test.sh
```diff
@@ -42,18 +43,19 @@ spec:
 EOF
 
 sleep 15s
-kubectl --namespace projectcontour port-forward svc/envoy 5436:80 &
+kubectl --namespace projectcontour port-forward svc/envoy 5436:80 & pid=$1
```
karuppiah7890, line 46: I think the `pid` has to be assigned like this

```suggestion
kubectl --namespace projectcontour port-forward svc/envoy 5436:80 & pid=$!
```

Also, we can use a new line I think, with a specific name for `pid`

```suggestion
kubectl --namespace projectcontour port-forward svc/envoy 5436:80 &
portforward_pid=$!
```

or similar

test/smoke/packages/local-path-storage/0.0.20/local-path-storage-test.sh
```diff
@@ -33,28 +41,26 @@ then
...
+timeout=300
+while [[ $(kubectl get pod volume-test -n ${NAMESPACE} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" && ${timeout} -ne 0 ]]; do echo "waiting for pod volume-test to get ready" && sleep 1 && timeout=${timeout}-1 ; done
 
-Output="$(kubectl exec volume-test cat /data/test)"
+Output="$(kubectl exec volume-test -n ${NAMESPACE} cat /data/test)"
 
 if [ "${Output}" != "local-path-storage-test" ]
 then
-    echo "local-path-storage test failed"
+    printf '\E[31m'; echo "local-path-storagefailed"
```
karuppiah7890, line 58: Nitty gritty

```suggestion
    printf '\E[31m'; echo "local-path-storage failed"
```

karuppiah7890 (changes requested):
LGTM ! I just have some thoughts on a few things to simplify the code and some nitty gritty things. I'll also execute and try this out next async to the review

test/smoke/packages/local-path-storage/0.0.20/local-path-storage-test.sh
```diff
@@ -13,15 +13,23 @@ then
...
+sleep 10s
+
 MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-kubectl create -f "${MY_DIR}"/pvc.yaml
-kubectl create -f "${MY_DIR}"/pod.yaml
+kubectl apply -n ${NAMESPACE} -f "${MY_DIR}"/pvc.yaml
+kubectl apply -n ${NAMESPACE} -f "${MY_DIR}"/pod.yaml
+
+timeout=300
+while [[ $(kubectl get pvc local-path-pvc -n ${NAMESPACE} -o 'jsonpath={.status.phase}') != "Bound" && ${timeout} -ne 0 ]]; do echo "waiting for get pvc local-path-pvc to get bound" && sleep 1 && timeout=${timeout}-1 ; done
```
karuppiah7890, line 26: Same thoughts as https://github.com/vmware-tanzu/community-edition/pull/2517#discussion_r747968589

test/smoke/packages/local-path-storage/0.0.20/local-path-storage-test.sh
```diff
@@ -13,15 +13,23 @@ then
...
 
-echo "Creating pod and PVC to test local-path storage..."
-sleep 30
+timeout=300
+while [[ $(kubectl get pod volume-test -n ${NAMESPACE} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" && ${timeout} -ne 0 ]]; do echo "waiting for pod volume-test to get ready" && sleep 1 && timeout=${timeout}-1 ; done
```
karuppiah7890, line 29: Same thoughts as https://github.com/vmware-tanzu/community-edition/pull/2517#discussion_r747968589

karuppiah7890 (changes requested):

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
karuppiah7890: I haven’t tried `kubectl wait` with custom resources like `Certificate` but reading the `kubectl wait -h` output, it just says it works with resources in resource. You can dig deep. I’ll also check it out sometime

test/smoke/packages/contour/1.18.2/contour-test.sh
```diff
@@ -15,19 +15,20 @@ then
...
 
 # Providing prerequisite 
-kubectl create namespace contour-example-workload
-kubectl create deployment nginx-example --image nginx --namespace contour-example-workload
-kubectl create service clusterip nginx-example --tcp 80:80 --namespace contour-example-workload
+NAMESPACE_SUFFIX=${RANDOM}
+NAMESPACE="contour-${NAMESPACE_SUFFIX}"
+kubectl create ns ${NAMESPACE}
+kubectl create deployment nginx-example --image nginx --namespace ${NAMESPACE}
+kubectl create service clusterip nginx-example --tcp 80:80 --namespace ${NAMESPACE}
```
karuppiah7890, line 24: I have noticed some machines require root access to bind to reserved ports like `80` or anything less than `1024`. Something to note

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
```diff
@@ -73,17 +82,18 @@ spec:
...
 
+tanzu package installed delete cert-manager -y
+kubectl delete ns ${NAMESPACE}
 if [ "${status}" != "True" ]
 then 
     echo "certificate status is not true"
-    echo "cert-manager failed"
+    printf '\E[31m'; echo "cert-manager failed"
```
karuppiah7890, line 95: Reset color using `printf '\E[0m'` for failed message too

test/smoke/packages/contour/1.18.2/contour-test.sh
```diff
@@ -42,18 +43,19 @@ spec:
...
 exitcode=$?
-kubectl delete ns contour-example-workload
+kubectl delete ns ${NAMESPACE}
 tanzu package installed delete contour -y
+trap '{ kill -9 $pid; }' EXIT
 
 if [ "${exitcode}" == 1 ] 
 then
-    echo "Contour failed"
+    printf '\E[31m'; echo "Contour failed"
```
karuppiah7890, line 57: Reset color using `printf '\E[0m'` for failed message too

test/smoke/packages/fluent-bit/1.7.5/fluent-bit-test.sh
```diff
@@ -24,8 +21,8 @@ tanzu package installed delete fluent-bit -y
 
 if [ "${exitcode}" == 1 ] 
 then
-    echo "Fluent-bit failed"
+    printf '\E[31m'; echo "Fluent-bit failed"
```
karuppiah7890, line 24: Reset color using `printf '\E[0m'` for failed message too

test/smoke/packages/local-path-storage/0.0.20/local-path-storage-test.sh
karuppiah7890, line 58: Reset color using `printf '\E[0m'` for failed message too

karuppiah7890 (changes requested):

karuppiah7890:
In short, these are some things we could do as part of this PR or another PR

- using kubectl wait instead of for loop with timeout calculation
- adding `set -o ...`
    - taking care of errors and exit codes using `||`
- using `kubectl get -o jsonpath` or `kubectl get -o template` for getting status, phase
- reset to white color after red error color
- check port 80 root access - use sudo or different port if it's an issue
- remove pid (ideally this PR as this can cause unnecessary problems)
- remove trap (ideally this PR as this can cause unnecessary problems)

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
```diff
@@ -73,17 +83,19 @@ spec:
     - www.example.com
 EOF
 
-sleep 45s
 
-status="$(kubectl get certificate/example-com | tail -n 1 | awk '{print $2}')"
-echo "${status}"
+echo "Waiting for certificate/example-com to Ready..."
+kubectl wait --for=condition=Ready certificate/example-com -n ${NAMESPACE} --timeout=300s
+status="$(kubectl get certificate/example-com -n ${NAMESPACE} -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}')"
```
karuppiah7890, line 89: If `kubectl wait` fails, it means that the certificate is not ready, so it will stop the script execution as we have `set -o errexit`. If it passes, then it means that the certificate is ready and `status` will always be `True`.

So how our script will work is - if wait fails, then script ends with failure in pipeline but with no log like `cert-manager failed`. If wait passes, then script ends with success in pipeline with a log like `cert-manager Passed`

We could do something like this if we want to get around this

```bash
kubectl wait --for=condition=Ready certificate/example-com -n ${NAMESPACE} --timeout=300s || {
  echo "Failed waiting for certificate status to be ready"
  tanzu package installed delete cert-manager -y
  kubectl delete ns ${NAMESPACE}
  printf '\E[31m'; echo "cert-manager failed"; printf '\E[0m'
}

tanzu package installed delete cert-manager -y
kubectl delete ns ${NAMESPACE}
printf '\E[32m'; echo "cert-manager Passed"; printf '\E[0m'
```

We can also pull delete process into a function -

```bash
function cleanup {
  tanzu package installed delete cert-manager -y
  # but ${NAMESPACE} has to be set and available here
  kubectl delete ns ${NAMESPACE}
}
```

test/smoke/packages/local-path-storage/0.0.20/local-path-storage-test.sh
```diff
@@ -3,58 +3,64 @@
...
 then
-    echo "PVC did not bound to PV. check the storage class name"
+    printf '\E[31m'; echo "PVC did not bound to PV. check the storage class name"; printf '\E[0m'
 fi
 
-if [ "${Pod_status}" != "Running" ]
+if [ "${Pod_status}" != "True" ]
 then
-    echo "Pod for PV is not in Running state"
+    printf '\E[31m'; echo "Pod for PV is not in Running state"; printf '\E[0m'
```
karuppiah7890, line 43: I think here you mean `Ready`, correct? I was also thinking, maybe `Running` makes sense? `Ready` is mainly derived from the readiness probe of the Pod. But if there's no readiness probe, I gotta check how else it understands that the pod is ready. Does it show something else other than `Ready` status when PV is not bound to the Pod? Like, `Pending`? I think it does show `Pending` when PV volumes are not attached, gotta confirm. `Pending` is shown when scheduler hasn't scheduled the Pod to a node, that I have seen

karuppiah7890 (approved):
Overall looks good to me! Great work 😄 We can merge this, let me know what you think about the `kubectl wait` error case alone

test/smoke/packages/cert-manager/1.5.3/cert-manager-test.sh
karuppiah7890: The only small downside of the current script is, when wait fails, the cleanup misses to happen but the pipeline will fail if it runs the script. And cleanup is not so important in pipelines that are ephemeral - where the cluster is docker and the pipeline environment is a container or a throw away VM like in our case for most providers (AWS, Docker, Azure). Also, the cluster deployment tests will take care of deleting and cleaning up the whole cluster however. It's just that in local, if someone is using the tests independently and deletes the cluster later etc, it might be left lying around. Other than that, not much issues

We actually gotta fix many tricky cleanup issues that we missed all over the E2E scripts

aman556:
thanks @karuppiah7890 agree with the `kubectl wait` error issue as i have some other scripts to add so thinking to do this change in next PR.

karuppiah7890:
Makes sense! 👍

karuppiah7890:
Btw, the port thing, I didn't realize that it was just a service exposing it. I was thinking it was a port-forward, sorry! Anyways, 8080 is fine too 😅

PR #2520, opened 2021-11-11
[WIP] Updating Framework to v0.10.0

dvonthenen:
## Details for the Release Notes (PLEASE PROVIDE)

```release-note

```

## Which issue(s) this PR fixes

Fixes: #

PR #2523, opened 2021-11-11
Use `buildinfo.Version` instead of `buildVersion` when building with tanzu framework v0.10.0

jpmcb:
This uses the newer `buildinfo.Version` api in tanzu framework instead of the `cli.buildVersion` api (which has been deprecated https://github.com/vmware-tanzu/tanzu-framework/commit/3162254543de7d6850fff5d1f2a1bbfff05c2182#diff-287b2c13779f4ba2d8e41cc7c201cb67605ab8ef8cce54a977df7a5c83b00dddR20-R21)

N/a - This is more to demonstrate using the new API so we can bump our tanzu framework version

Related to https://github.com/vmware-tanzu/community-edition/pull/2520

Able to `make install-all-tanzu-cli-plugins`

See note below on the standalone-cluster version


cli/cmd/plugin/standalone-cluster/main.go
```diff
@@ -15,6 +15,11 @@ var descriptor = cliv1alpha1.PluginDescriptor{
 	Name:        "standalone-cluster",
 	Description: "Create clusters without a dedicated management cluster",
 	Group:       cliv1alpha1.RunCmdGroup,
+
+	// Since standalone cluster is being deprecated, no changes are being made to it's dependencies,
+	// and cli.BuildVersion was deprecated in the build version of tanzu framework v0.10.0,
+	// we need to manually set the plugin version in order to compile
+	Version: "v0.9.0",
```
jpmcb, line 22: Note that we're calling out an explicit version here. This means that running `make release` locally fails since it can't find the `0.9.2-dev...` release.

This was necessary since we won't be bumping the tanzu-framework version for standalone-cluster but we still need to be able to build with `hack/builder` which is injecting the `buildinfo.Version`. So otherwise, we end up with the "version cannot be empty" problem that we run into from time to time.

To get around this, I propose we keep this as a literal string, change it as needed when releasing, and then eventually it will be fully deprecated and we won't have to mess with changing the version of tanzu framework we use in standalone cluster

dvonthenen: Yea, this is totally fine. I would suggest keeping the last release version `v0.9.1` for the value and then we just need then to hardcore the version for just this plugin in the package script which needs to be done in 3 places (here is the first place https://github.com/vmware-tanzu/community-edition/blob/main/hack/release/package-release.sh#L68).

jpmcb: Which other locations were needed for an update here? I updated the link you mentioned and ran `make release` without problem and the tarball looks good

dvonthenen: just that file in those 3 places. yup, it looks good!

dvonthenen: I forgot... we have a loop now for all 3. just in the one place!

davidvonthenen (dismissed):
woot woot!

davidvonthenen (approved):

PR #2566, opened 2021-11-17
Fix ako statefulset upgrade issue

XudongLiuHarold:

Currently, when users upgrade `load-balancer-and-ingress` package, they may meet following error:
```bash
 USEFUL-ERROR-MESSAGE: kapp: Error: Applying update statefulset/ako (apps/v1) namespace: avi-system:
 Updating resource statefulset/ako (apps/v1) namespace: avi-system:
 StatefulSet.apps "ako" is invalid: spec:
 Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden (reason: Invalid)
```
if they changed anything in those fields.

To avoid this problem, we add annotations to ignore default kapp-controller rules to ensure upgrade successfully.

```release-note
Users can change statefuleset's template, replica and etc field now.
```

Fixes: #


HanFa (approved):
LGTM

Issue #1590, opened 2021-09-07
Release Engineeering - Docs and / code checks on pre-requisites


karuppiah7890:
Noting down some stuff for future reference

There are some pre-requisites needed to build and run TCE - mostly in CI pipelines. Some of them are documented and also codified to do check and even do automatic installations. Some of them don't have automatic installation in some scripts, some of them don't have both checks and automatic installation

Checks but no automatic installation
- `curl` in [`ensure-dependencies.sh`](https://github.com/vmware-tanzu/community-edition/blob/main/hack/ensure-dependencies.sh)
- `docker` in [`ensure-dependencies.sh`](https://github.com/vmware-tanzu/community-edition/blob/main/hack/ensure-dependencies.sh). But there is check and automation available here for debian alone - https://github.com/vmware-tanzu/community-edition/blob/main/test/install-dependencies.sh#L12-L43
- `go` in [`ensure-dependencies.sh`](https://github.com/vmware-tanzu/community-edition/blob/main/hack/ensure-dependencies.sh).

No checks and no automatic installation
- `make`
- `git`
- `zip` - there's an automatic installation for GitLab alone here - https://github.com/vmware-tanzu/community-edition/blob/main/hack/fix-for-ci-build.sh#L14

Not sure what other tools are there which need checks and/ automatic installation. The above are the ones I noticed. But yeah, `make`, `git` and `zip` are popular and usually available everywhere, but sometimes, CI pipelines are too lean and light weight, so they may not have it

PR #2570, opened 2021-11-18
add script and make target to trigger release gate pipelines

karuppiah7890:

This is a follow up of #2304 . This is to add a script and `make` target to trigger release gate pipelines. This would be used downstream when the downstream pipelines signs artifacts and uploads them to the TCE draft release, it would trigger the release gate pipelines using `make release-gate`

This also has some related minor changes -
- extract out ensure-gh-cli from ensure-dependencies script and use it in release-gate make target. Why? As `make release-gate` only needs `gh` and nothing else, also, when using `make release-gate` in a Docker container like `golang:1.16.2` or even something like `ubuntu:20.04`, the container image does not have `docker` installed by default causing the `ensure-deps` to fail as `docker` installation is not automated, and there are too many ways to handle the Docker install. So, to avoid such failures, I extracted out `gh` installation and used that alone in `make release-gate`
- install `make`, `curl` and `sudo` to ensure they are part of CI. Related - #1590. Why? As `make` is not installed by default in container images like `ubuntu:20.04`, fortunately `golang:1.16.2` has `make` and `git` installed by default, which is what is used in downstream to run `make release-gate` even though `go` and `git`  is not required to use `make release-gate`. But, `sudo` is not installed in `golang:1.16.2` and `sudo` is used in the `gh` installation and the installation of other dependencies in `ensure-dependencies.sh` script. For this reason I ended up adding essential tools like `curl`, `make`, `sudo` as part of the `fix-for-ci-build`. I didn't add `git` though, though I could have added it
- update gh CLI version from `v1.14.0` to `v2.2.0`. Why? Just to keep using updated CLI with bug fixes and any new features. And there don't seem to be any breaking changes for `v2.x.y` - https://github.com/cli/cli/releases/tag/v2.0.0

```release-note
NONE
```

--

Ran `make release-gate` locally -

```bash
tce $ BUILD_VERSION=v0.10.0-rc.1 make release-gate
hack/ensure-dependencies.sh
++ uname
+ BUILD_OS=Darwin
++ uname -m
+ BUILD_ARCH=x86_64
++ go env GOBIN
+ GOBIN=
++ go env GOPATH
+ GOPATH=/Users/karuppiahn/go
+ GOBINDIR=/Users/karuppiahn/go/bin
+ case "${BUILD_OS}" in
+ i=0
++ command -v go
+ [[ -z /Users/karuppiahn/go/bin/go ]]
++ command -v docker
+ [[ -z /usr/local/bin/docker ]]
++ command -v curl
+ [[ -z /usr/bin/curl ]]
+ [[ 0 -gt 0 ]]
+ CMD=imgpkg
++ command -v imgpkg
+ [[ -z /usr/local/bin/imgpkg ]]
+ CMD=kbld
++ command -v kbld
+ [[ -z /usr/local/bin/kbld ]]
+ CMD=ytt
++ command -v ytt
+ [[ -z /usr/local/bin/ytt ]]
+ CMD=shellcheck
++ command -v shellcheck
+ [[ -z /usr/local/bin/shellcheck ]]
+ CMD=gh
++ command -v gh
+ [[ -z /usr/local/bin/gh ]]
+ CMD=release-notes
++ command -v release-notes
+ [[ -z /usr/local/bin/release-notes ]]
+ CMD=aws
++ command -v aws
+ [[ -z /usr/local/bin/aws ]]
+ CMD=jq
++ command -v jq
+ [[ -z /usr/local/bin/jq ]]
+ CMD=kubectl
++ command -v kubectl
+ [[ -z /usr/local/bin/kubectl ]]
+ CMD=kind
++ command -v kind
+ [[ -z /usr/local/bin/kind ]]
+ echo 'No missing dependencies!'
No missing dependencies!
./hack/release/trigger-release-gate-pipelines.sh
--- Checking dependencies ---
--- Dependencies verified --- 

Checking if v0.10.0-rc.1 release exists
Triggering MacOS and Linux build tests
✓ Created workflow_dispatch event for release-gate-macos-linux-builds.yaml at main

To see runs for this workflow, try: gh run list --workflow=release-gate-macos-linux-builds.yaml
Triggering Windows build tests
✓ Created workflow_dispatch event for release-gate-windows-build.yaml at main

To see runs for this workflow, try: gh run list --workflow=release-gate-windows-build.yaml
tce $ 
```

Which created the following two pipelines -

https://github.com/vmware-tanzu/community-edition/actions/runs/1475386920
https://github.com/vmware-tanzu/community-edition/actions/runs/1475386944

Obtained lot of helper functions and code from https://github.com/vmware-tanzu/community-edition/blob/main/hack/lint-repo-links.sh


hack/release/trigger-release-gate-pipelines.sh
```diff
@@ -0,0 +1,86 @@
...
+# MacOS and Linux Build Test
+gh workflow run --repo https://github.com/vmware-tanzu/community-edition/ \
+  --ref main \
+  release-gate-macos-linux-builds.yaml \
+  --raw-field release_version="${BUILD_VERSION}"
+
+info "Triggering Windows build tests"
+
+# Windows Build Test
+gh workflow run --repo https://github.com/vmware-tanzu/community-edition/ \
```
stmcginnis, line 83: Just to confirm, is there some auth protection on this to make sure only certain people can trigger a release workflow?

hack/ensure-dependencies.sh
```diff
@@ -141,39 +142,7 @@ case "${BUILD_OS}" in
...
-        rm gh_1.14.0_macOS_amd64.tar.gz
-        ;;
-      arm64)
-        brew install gh
-        ;;
-    esac
-    ;;
-esac
-fi
+"${TCE_REPO_PATH}/hack/ensure-gh-cli.sh"
```
dvonthenen, line 145: If we want to break this up into individual scripts, we should probably do this all at once.

hack/ensure-gh-cli.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
dvonthenen, line 1: If we decide to break this up into individual scripts, I have been trying to clean up the hack directory... if we do this, we should probably probably move these individual scripts into a `ensure-deps` folder.

hack/release/trigger-release-gate-pipelines.sh
dvonthenen: Agreed. This is how we do this when we cut a release:
https://github.com/vmware-tanzu/community-edition/blob/main/hack/release/cut-release.sh#L36-L40

This should be done at the top of the script.

davidvonthenen (commented):
Left some feedback.

hack/release/trigger-release-gate-pipelines.sh
karuppiah7890: > Just to confirm, is there some auth protection on this to make sure only certain people can trigger a release workflow?

@stmcginnis I believe it's the default settings on the GitHub repo and the GitHub API. Not everyone can trigger GitHub Actions workflows. The authn/authz completely depends on the account being used with `gh`. In this case, we use `GITHUB_TOKEN` environment variable or ask the user to use `gh auth login` to authenticate - https://github.com/vmware-tanzu/community-edition/pull/2570/files#diff-dedc80ea59f623bfdbbaddf70b6c30e6fc18a7cc3fcadf7dd840c8346cee65caR50-R53

karuppiah7890: > Agreed. This is how we do this when we cut a release: https://github.com/vmware-tanzu/community-edition/blob/main/hack/release/cut-release.sh#L36-L40
> 
> This should be done at the top of the script.

Ah yeah, I did see this, along with `# IMPORTANT: This should only ever be called CI/github-action` in `Makefile`.

I think the script can be used by devs too, it's okay and for the same reason I was a bit skeptical to put inside the `# IMPORTANT: This should only ever be called CI/github-action` section of `Makefile` as that was a conflicting thing. I could move it out of that section in `Makefile`

To answer the core of `Why should this be allowed to be run by devs and not just CI?` - Well, what this script does is simply trigger some workflows in this repo with a particular input - the release version. And not all can trigger this workflow like I already mentioned in https://github.com/vmware-tanzu/community-edition/pull/2570#discussion_r752973765 . The workflow simply runs and does a check if the release build works by installing and running some commands. A dev running this from local is not gonna affect anything, unlike https://github.com/vmware-tanzu/community-edition/blob/main/hack/release/cut-release.sh script where there's an effect - git commit, git push, PR creation, PR merge etc. So I think it's okay to not add anything like https://github.com/vmware-tanzu/community-edition/blob/main/hack/release/cut-release.sh#L36-L40 `TCE_CI_BUILD`. A run from local by a dev is not gonna negatively affect anything

hack/ensure-dependencies.sh
karuppiah7890: Okay, I'll do that 👍

hack/ensure-gh-cli.sh
karuppiah7890: Ah okay. If you are already doing this, then I'll leave it out of this PR. Does that make sense? I think the same could be said about https://github.com/vmware-tanzu/community-edition/pull/2570#discussion_r752434686

dvonthenen: I'm not currently doing this work. I was saying if we want to break them apart, we should do it for everything. In the mean time until that work happens, can you move this into a subfolder called `ensure-deps`.

hack/release/trigger-release-gate-pipelines.sh
dvonthenen: Yea, it's just a warning to users. The whole reason for putting this in the makefile is because it then becomes testable by devs.

davidvonthenen (approved):

dvonthenen:
Oops forgot to ask for a squash and fix commit message

PR #2548, opened 2021-11-16
[E2E] Fix cleanup of kind cluster when standalone-cluster or management-cluster delete fails

karuppiah7890:
## What this PR does / why we need it

Fixes cleanup of kind cluster when standalone-cluster or management-cluster delete fails. Also removes unnecessary cleanup of kind cluster

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
NONE
```

## Which issue(s) this PR fixes

Fixes: #2533 

## Describe testing done for PR

--

## Special notes for your reviewer

--

PR #2601, opened 2021-11-25
[E2E] delete kind bootstrap cluster created for deletion only when management cluster deletion fails

### LINKED ISSUE #2533: Fix delete / cleanup of kind cluster when standalone-cluster or management-cluster delete fails
Fix delete / cleanup of kind cluster when standalone-cluster or management-cluster delete fails. Currently this is not being done, for example -

https://github.com/vmware-tanzu/community-edition/blob/0bf658cbbb3ce8b5469f2ff76a04bbdb8ce2aa4a/test/azure/deploy-standalone-cluster.sh#L63-L70

https://github.com/vmware-tanzu/community-edition/blob/0bf658cbbb3ce8b5469f2ff76a04bbdb8ce2aa4a/test/azure/deploy-management-and-workload-cluster.sh#L278-L281

Also, as part of this issue, we can also fix unnecessary kind cluster delete processes. For example I made a mistake adding this extra `delete_kind_cluster` here -

https://github.com/vmware-tanzu/community-edition/blob/0bf658cbbb3ce8b5469f2ff76a04bbdb8ce2aa4a/test/vsphere/run-tce-vsphere-standalone-cluster.sh#L63-L68

Ideally, if `tanzu standalone-cluster create` is working correctly, it would have deleted the kind bootstrap cluster once the standalone cluster creation is done. So, when it comes to `check-tce-cluster-creation.sh`, the kind bootstrap cluster would ideally not be present, but I added the `delete_kind_cluster` due to some confusion with #2457 issue and it's PR #2459 . So the `delete_kind_cluster` after `check-tce-cluster-creation.sh` but it doesn't cause any problems in a clean CI environment, no failures etc, as we use `xargs` -

https://github.com/vmware-tanzu/community-edition/blob/0bf658cbbb3ce8b5469f2ff76a04bbdb8ce2aa4a/test/util/utils.sh#L10-L13

But yeah, in local (dev machine) it would unnecessarily delete some kind bootstrap clusters, which is also a generic problem mentioned here - #1270 with a possible solution but it hasn't been implemented yet, so I guess that's a separate issue but I just wanted to call it out here. But in any case, the `delete_kind_cluster` just right after `check-tce-cluster-creation.sh` failure is not necessary

karuppiah7890:

Fixes #2533 - https://github.com/vmware-tanzu/community-edition/issues/2533#issuecomment-978795910 . Fixes some issues in #2548 

```release-note
NONE
```

Fixes: #2533 

I'll update here all links after I do the tests

Full green test where there are no failures - https://github.com/karuppiah7890/community-edition/runs/4366382786?check_suite_focus=true

Failure when doing cluster check (using `kubectl cluster-info`) - https://github.com/karuppiah7890/community-edition/runs/4378510233?check_suite_focus=true - cleanup happens correctly, without any errors

Should we also absorb any errors here? -

https://github.com/vmware-tanzu/community-edition/blob/65dc609f95b3e78dda4c06e9662e0abee7a8bb02/test/util/utils.sh#L12

To avoid issues like https://github.com/vmware-tanzu/community-edition/issues/2533#issuecomment-978796801

The downside is during cluster creation failure, if kind bootstrap cluster is not deleted, then our cleanup script will not work as bootstrap cluster will keep trying to create the cluster. But deletion not happening is a rare case I guess, only basic reason I can think of this can happen is - the docker container doesn't exist in the first place

And during cluster deletion, if cluster deletion using bootstrap cluster fails, and bootstrap cluster is not deleted and error is absorbed, it's not a problem I guess, at least theoretically. However our cleanup script will also try to delete resources, bootstrap cluster will also to delete resources. But I'm not sure

The other downside is, docker containers remain when deletion fails, again, rare case, as I don't see popular reasons for it

### DIFF
--- a/test/azure/deploy-management-and-workload-cluster.sh
+++ b/test/azure/deploy-management-and-workload-cluster.sh
@@ -60,7 +60,6 @@ function cleanup_management_cluster {
     echo "Using azure CLI to cleanup ${MANAGEMENT_CLUSTER_NAME} management cluster resources"
     export CLUSTER_NAME="${MANAGEMENT_CLUSTER_NAME}"
     set_azure_env_vars
-    delete_kind_cluster
     kubeconfig_cleanup ${CLUSTER_NAME}
     azure_cluster_cleanup || error "MANAGEMENT CLUSTER CLEANUP USING azure CLI FAILED! Please manually delete any ${MANAGEMENT_CLUSTER_NAME} management cluster resources using Azure Web UI"
     unset_azure_env_vars
@@ -289,6 +288,7 @@ kubeconfig_cleanup ${WORKLOAD_CLUSTER_NAME}
 
 delete_management_cluster || {
     collect_management_cluster_diagnostics ${MANAGEMENT_CLUSTER_NAME}
+    delete_kind_cluster
     cleanup_management_cluster
     exit 1
 }
--- a/test/vsphere/run-tce-vsphere-management-and-workload-cluster.sh
+++ b/test/vsphere/run-tce-vsphere-management-and-workload-cluster.sh
@@ -53,7 +53,6 @@ export MANAGEMENT_CLUSTER_NAME="test-management-cluster-${random_id}"
 export WORKLOAD_CLUSTER_NAME="test-workload-cluster-${random_id}"
 
 function cleanup_management_cluster {
-    delete_kind_cluster
     kubeconfig_cleanup ${MANAGEMENT_CLUSTER_NAME}
     echo "Using govc to cleanup ${MANAGEMENT_CLUSTER_NAME} management cluster resources"
     govc_cleanup ${MANAGEMENT_CLUSTER_NAME} || error "MANAGEMENT CLUSTER CLEANUP USING GOVC FAILED! Please manually delete any ${MANAGEMENT_CLUSTER_NAME} management cluster resources using vCenter Web UI"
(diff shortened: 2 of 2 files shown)


davidvonthenen (approved):

dvonthenen:
Just need to take the PR out of draft when ready to merge

joshrosso:
@karuppiah7890 noticed this PR is still in `draft` mode.

What's your preference on moving forward with this PR?

karuppiah7890:
hey @joshrosso ! I think I wanted to do some more tests with some scenarios and that's why had this in draft mode. I'll try to get back to this soon. The fix is straight forward, but I just wanted to try out different scenarios 😅

karuppiah7890:
I think some scenarios are hard to simulate. Some of the ideas I had to simulate failures were -

- For simulating failure in bootstrap cluster deleting a management cluster - make the API credentials invalid while the bootstrap cluster tries to delete the management cluster. This will fail the deletion process and then the cleanup happens - where ideally the bootstrap cluster along with the management cluster should be deleted

- For simulating failure in pod not ready within max timeout time (15 minutes), continuously kill the pod that tries to come up

Given that it's a low key change and we know the logic of the cleanup, I'm leaving off the testing of the above scenarios. But yeah, it's an interesting idea to think about how to test (manually or automated) the E2E test code ;) Test for test, haha

PR #2637, opened 2021-12-01
add antrea configuration value for serviceCIDRv6

tylerschultz:
Signed-off-by: Tyler Schultz <tschultz@vmware.com>
Co-authored-by: Aidan Obley <aobley@vmware.com>

This change is being made to support dualstack configurations.
When running single stack, v4 or v6, only the serviceCIDR needs to be
set.

When running dualstack, both a v4 and v6 CIDR needs to be set in the
ConfigMap antrea-agent.conf serviceCIDR and serviceCIDRv6 config variables.
This change makes that possible.

An example of how the config vars are set can be found in the framework
repo:
https://github.com/vmware-tanzu/tanzu-framework/blob/5338922f196a8cbb33114df45bf9f725c5f9579c/pkg/v1/providers/ytt/02_addons/cni/antrea/antrea_addon_data.lib.yaml#L41-L47

```release-note
NONE
```

Part of https://github.com/vmware-tanzu/tanzu-framework/issues/616

- Unit tests added
- Manual configuration of running cluster


addons/packages/antrea/1.2.3/test/Makefile
```diff
@@ -0,0 +1,28 @@
...
+	$(GOLANGCI_LINT) run -v --timeout=5m
+endif
+
+get-deps: ## Get all dependencies
+	go mod download
+
+test: ## Run unit testing suite
+	ginkgo -v .
+
+e2e-test: ## Run e2e testing suite
```
knabben, line 24: can we get ride of the targets we don't need?

addons/packages/antrea/1.2.3/test/antrea_test.go
```diff
@@ -0,0 +1,116 @@
...
+  EndpointSlice: false
+  Traceflow: true
+  NodePortLocal: false
+  AntreaPolicy: true
+  FlowExporter: false
+  NetworkPolicyStats: false
+  Egress: false
+trafficEncapMode: encap
+noSNAT: false
+serviceCIDR: 10.96.0.0/12
```
knabben, line 59: what about adding `serviceCIDRv6: null` in this list, so we match the `values.yaml` content

addons/packages/antrea/1.2.3/bundle/config/values.yaml
```diff
@@ -6,6 +6,7 @@ infraProvider: vsphere
 antrea:
   config:
     serviceCIDR: 10.96.0.0/12
+    serviceCIDRv6: null
```
antoninbas, line 9: These are not required when AntreaProxy is enabled, which seems to be the same (see `featureGates` below)

https://github.com/antrea-io/antrea/blob/9a52de90e90a8beeb512f56c51b670d93ec82325/build/yamls/antrea.yml#L4050-L4059

addons/packages/antrea/1.2.3/test/Makefile
adobley: A standard subset of make tasks was added to all of the modules in https://github.com/vmware-tanzu/community-edition/pull/1143 and it sounds like the expectation is that they are present.

addons/packages/antrea/1.2.3/bundle/config/values.yaml
adobley: In the overlay we currently omit the `serviceCIDRv6` if it is null or unset. [On line 103 of the overlay.](https://github.com/tylerschultz/community-edition/blob/5736884161ab0fd127b9adac5524380b8c0fd5e0/addons/packages/antrea/1.2.3/bundle/config/overlay/antrea_overlay.yaml#L103)

We could add a check to see if the `featureGate` for `AntreaProxy` is true, to avoid setting the `serviceCIDRv6` in that case. Would that be the best thing to do? Is there a risk in setting the `serviceCIDRv6` if the `AntreaProxy` flag is true?

addons/packages/antrea/1.2.3/test/antrea_test.go
adobley: The current behavior of other fields is to not display them if they are null. Such as the `defaultMTU`.

```
#@ if/end values.antrea.config.defaultMTU:
defaultMTU: #@ values.antrea.config.defaultMTU
```
[source](https://github.com/tylerschultz/community-edition/blob/5736884161ab0fd127b9adac5524380b8c0fd5e0/addons/packages/antrea/1.2.3/bundle/config/overlay/antrea_overlay.yaml#L86)

addons/packages/antrea/1.2.3/bundle/config/values.yaml
antoninbas: there is no risk as the field is ignored (there isn't even a warning in the logs IIRC) when AntreaProxy is enabled. so no issue with always emitting these CIDRs, if there is a possibility for the user to disable AntreaProxy.

adobley: The users can set any of the values in the `values.yaml`, so they could set the `AntreaProxy` feature to `false`. We'll leave it as it is in this case.

addons/packages/antrea/1.2.3/test/Makefile
knabben: it's an odd default but fair enough as a standard.

knabben (approved):

PR #2668, opened 2021-12-06
Inform user that Tanzu cli is also installed

### LINKED ISSUE #2665: The readme.md should tell the user that install also adds the tanzu cli
## Description

When you do any of the different install methods, including manual, not only is TCE is installed but so is the Tanzu CLI. 
This is important to note to the user because it is only if they click the getting started guide, then scroll through all the instructions and happen to notice the light grey instructions which state the CLI is also installed. 

Some users will probably install, go to look at some examples, and not be sure where the tanzu cli came from. 

All that is needed is adding the line, right above the packages heading that says:

```## At the end of the installation you now have both Tanzu Community Edition and the Tanzu CLI installed on your computer```

## Location

(https://github.com/vmware-tanzu/community-edition/blob/main/README.md)

### PR DESCRIPTION
**Pull request by thesteve0**
Fixes #2665

See the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ choco install tanzu-community-edition
     * `install.sh` on Mac/Linux
 
 ### Installed Binaries
+
 Upon installation completion, both the Tanzu Community Edition and the Tanzu CLI binaries are installed on your computer
 
 ## Packages
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ choco install tanzu-community-edition
     * `install.bat` on Windows as Administrator.
     * `install.sh` on Mac/Linux
 
+### Installed Binaries
+Upon installation completion, both the Tanzu Community Edition and the Tanzu CLI binaries are installed on your computer
+
 ## Packages
 
 Packages provide higher-level functionality to build atop your Kubernetes-based

### CODE REVIEW

**Comment by dvonthenen** [comment]
I left a comment in the issue created. Not sure where the appropriate place for clarification should be.

File: README.md [hunk 1]
```diff
@@ -70,6 +70,10 @@ choco install tanzu-community-edition
     * `install.bat` on Windows as Administrator.
     * `install.sh` on Mac/Linux
 
+### Installed Binaries
+
+Upon installation completion, both the Tanzu Community Edition and the Tanzu CLI binaries are installed on your computer
```
stmcginnis, line 75: Please end the sentence with a period. If you can, also squash the commits so we have one clean commit message (though I can fix that up at the end if that's too much of a hassle for this).

dvonthenen: Yea, I think we should think about this sentence for a bit. It sounds like 2 different packages are getting installed when it's really only one thing.

stmcginnis: I do agree. I'm not sure about this issue - like you point out, TCE is the tanzu CLI so I am not sure why it would be confusing that the `tanzu` binary is present after installation.

thesteve0: Well I was confused. Nowhere is it explained that the TCE binary is also the tanzu cli

dvonthenen: you do bring up a good point though... if there are sections of documentation where it sounds like they are two different things, we should probably fix those.

Yea, there really isn't a TCE binary... TCE is a distribution or packaging of the Tanzu CLI.

joshrosso:
Closing this PR due to inactivity. If you'd still like to make these improvements, please re-create a PR against `main`.

PR #2694, opened 2021-12-09
Add prefix to data hash, so it cannot be parsed as integer.

HanFa:

We observe scenarios that some data value hash, e,g, `5386e49` will be treated as an integer by the ytt template. And kubectl cannot apply the rendered yaml file. Submit a fix that prefix "h-" before the hash to enforce it as a string.

```release-note
Add prefix to data hash, so it cannot be parsed as integer.
```

Fixes: #

Unit test of  the vsphere-cpi addon.


lubronzhan:
/lgtm
Hi @vijaykatam could you take a look as well

lubronzhan (approved):

vijaykatam (approved):

PR #2514, opened 2021-11-11
Update `get-deps` Makefile targets

karuppiah7890:
## What this PR does / why we need it

_Note: This is a follow up of #2470 PR_

Most get-deps targets were using "go mod download", which isn't as
accurate as "go mod tidy" for evaluating dependencies. From one of the
go maintainers:

```
go mod download is downloading all of the modules in the dependency
graph, which it can determine from reading only the go.mod files. It
doesn't know which of those modules are actually needed to satisfy a
build, so it doesn't pull in the checksums for those modules (because
they may not be relevant).

On the other hand, go mod tidy has to walk the package graph in order
to ensure that all imports are satisfied. So it knows, for a fact,
without doing any extra work, that the modules it walks through are
needed for a go build in some configuration.
```

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
NONE
```

## Which issue(s) this PR fixes

--

## Describe testing done for PR

--

## Special notes for your reviewer

This is a follow up of #2470 based on the conversation [here](https://github.com/vmware-tanzu/community-edition/pull/2470#discussion_r746874456). This is based on the commit from #2498 by @stmcginnis

PR #2712, opened 2021-12-10
Create a single daily build that is downloadable for users and tests

davidvonthenen:

This PR implements **public** daily builds that can be consumed by users in the community and also can be used for other GitHub actions to trigger very long-running tests that should only be run daily.

IMPORTANT NOTE: These are daily test builds and therefore are **UNSIGNED** on purpose.

With the availability of this workflow, this updates the current vSphere e2e so that instead of each separately building their own install package, after the nightly build is created and uploaded, these e2e tests now just download/use this single nightly build to exercise their tests.

Then create a very basic rss feed about the build, commit the rss.xml file to the repo, so that file can be scraped in the k8s slack channel to post a public message to users.

Side note: when copying cut-release.sh for the basis for the other script, noticed there was a logic error when doing the rebase within the script since we support creating releases from branches other than main as well.

```release-note
Create a single daily build that is downloadable for users and tests
```

NA

Verified that:
- the tarballs are uploaded to the GCP bucket using the make command
- the rss.xml is created successfully

NA


stmcginnis (commented):
Looks good to me once linting issue is addressed.

stmcginnis (dismissed):

Makefile
```diff
@@ -228,6 +228,10 @@ version:
 	@echo "FRAMEWORK_BUILD_VERSION:" ${FRAMEWORK_BUILD_VERSION}
 	@echo "XDG_DATA_HOME:" $(XDG_DATA_HOME)
 
+.PHONY: upload-daily-build
+upload-daily-build:
+	cd ./hack/dailybuild && $(MAKE) run && cd ../..
```
karuppiah7890, line 233: We could use `pushd` and `popd` too here or `cd -` instead of `cd ../..` which is usually tricky when directory structures change and we miss out to change it

hack/dailybuild/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+lint: ## Run Go code linting. Requires the $GOLANGCI_LINT env var to be set.
+ifeq ($(origin GOLANGCI_LINT),undefined)
+	@echo "Error! GOLANGCI_LINT env var not set"
+else
+	$(GOLANGCI_LINT) run -v --timeout=5m
+endif
+
+get-deps: ## Get all go dependencies
+	go mod download
```
karuppiah7890, line 17: Based on #2514 I think we could use `go mod tidy` here

hack/dailybuild/go.mod
```diff
@@ -0,0 +1,5 @@
+module dailybuild
+
+go 1.16
+
+require cloud.google.com/go/storage v1.15.0
```
karuppiah7890, line 5: Just an FYI, there is a newer version of the package https://pkg.go.dev/cloud.google.com/go/storage `v1.18.2` as of this writing

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	var token string
+	if v := os.Getenv("GCP_BUCKET_SA"); v != "" {
+		token = v
+	}
+
+	flag.Parse()
+
+	if tag == "" {
+		fmt.Printf("A tag must be provided\n")
+		return
```
karuppiah7890, line 109: Should we exit with non-zero exit code?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if token == "" {
+		fmt.Printf("A token must be provided\n")
+		return
```
karuppiah7890, line 113: Should we exit with non-zero exit code?

hack/dailybuild/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+e2e-test:
+	@echo "N/A: No e2e tests for hack/dailybuild"
+
+build: ## Build the executable
+	go build -o dailybuild dailybuild.go
+
+run: ## Run the dailybuild program to upload dailybuild. Requires the $BUILD_VERSION env var to be set.
+ifeq ($(origin BUILD_VERSION),undefined)
+	@echo "Error! BUILD_VERSION env var not set"
```
karuppiah7890, line 30: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	err := os.WriteFile(gcpFile, []byte(token), 0644)
+	if err != nil {
+		fmt.Printf("WriteFile failed: %v\n", err)
+		return
```
karuppiah7890, line 119: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	err = upload("linux", "amd64", tag, "tar.gz")
+	if err != nil {
+		fmt.Printf("uploadLinux failed: %v\n", err)
+
+		err = os.Remove(gcpFile)
+		if err != nil {
+			fmt.Printf("Remove failed: %v\n", err)
+		}
+
+		return
```
karuppiah7890, line 131: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	err = upload("windows", "amd64", tag, "zip")
+	if err != nil {
+		fmt.Printf("uploadWindows failed: %v\n", err)
+
+		err = os.Remove(gcpFile)
+		if err != nil {
+			fmt.Printf("Remove failed: %v\n", err)
+		}
+
+		return
```
karuppiah7890, line 143: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	err = upload("darwin", "amd64", tag, "tar.gz")
+	if err != nil {
+		fmt.Printf("uploadDarwinAmd64 failed: %v\n", err)
+
+		err = os.Remove(gcpFile)
+		if err != nil {
+			fmt.Printf("Remove failed: %v\n", err)
+		}
+
+		return
```
karuppiah7890, line 155: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	err = upload("darwin", "arm64", tag, "tar.gz")
+	if err != nil {
+		fmt.Printf("uploadDarwinArm64 failed: %v\n", err)
+
+		err = os.Remove(gcpFile)
+		if err != nil {
+			fmt.Printf("Remove failed: %v\n", err)
+		}
+
+		return
```
karuppiah7890, line 167: Should we exit with non-zero exit code here?

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	err = uploader.UploadFile(assetFilename)
+	if err != nil {
+		fmt.Printf("UploadFile failed: %v\n", err)
+		return err
+	}
+
+	urlFile := fmt.Sprintf("%s-%s.txt", platform, arch)
+	urlUpload := fmt.Sprintf("https://storage.googleapis.com/tce-cli-plugins-staging/%s%s", uploadPath, assetFilename)
```
karuppiah7890, line 86: I guess we could use `bucketName` variable instead of `tce-cli-plugins-staging`

karuppiah7890 (dismissed):
LGTM!

dvonthenen:
updated based on feedback!

hack/dailybuild/dailybuild.go
```diff
@@ -0,0 +1,232 @@
...
+
+func uploadAndUpdateRss(tag, gcpToken string) error {
+	// upload
+	err := os.WriteFile(gcpFile, []byte(gcpToken), 0644)
+	if err != nil {
+		fmt.Printf("WriteFile failed: %v\n", err)
+		return err
+	}
+
+	urlLinuxAmd64, err := upload("linux", "amd64", tag, "tar.gz")
```
stmcginnis, line 99: Something to think about as a future optimization... I wonder if we can plumb through `ENVS` from the Makefile so this automatically picks up all platforms we're building for. Then in here, we can just loop through each `ENV` entry to process it.

dvonthenen: ahhhhh that's a good idea. will need to think about hooking that up later.

stmcginnis (approved):
This looks good overall to me. I say let's merge it and see how things go.

hack/dailybuild/dailybuild.go
dvonthenen: one problem we currently have is that we are building darwin arm64 but do we want to include it in the list of the daily build. at some point when we do support it we definitely will... but listing it right now might lead people to think it's a support platform/arch

karuppiah7890 (approved):
LGTM!

Issue #2717, opened 2021-12-13
Release `community-edition` (project) decoupled from `Tanzu Kubernetes Grid` (product)


joshrosso:
At the time of writing, Tanzu Community Edition (TCE) relies on the Bill of Materials (BOMs) produced for VMware's paid product, Tanzu Kubernetes Grid (TKG). This means that creation of clusters, both management and workload, always use the same packages, images, and underlying components as the rest of TKG. This binding to a TKG BOM was a decision made to ensure we could release TCE. Now we are at stage where we must unbind from TKG's BOM and begin producing TCE-specific BOMs.

Doing so will enable:

- Building and packaging in the open, rather than behind VMware's network.
- Packages that can be re-produced by users.
- Releasing Kubernetes versions ahead of TKG.
- Releasing core packages (e.g. antrea) ahead of TKG.
- Releasing core-packages specific to TCE.
- Patching TCE releases out-of-band from TKG.

## Issue Tracking

The following list tracks the issues required to resolve in order to achieve this capability. Please see the next section to understand the larger proposal.

### Tanzu Framework

* https://github.com/vmware-tanzu/tanzu-framework/issues/877

## Proposal

* [Pull Request](https://github.com/vmware-tanzu/community-edition/issues/2717)
* [Rendered View](https://github.com/vmware-tanzu/community-edition/blob/proposal-tce-distro/docs/designs/2717-produce-tce-specific-releases.md)

joshrosso:
Updates the issue description to include the https://github.com/vmware-tanzu/tanzu-framework/issues/877 work done by framework which will declare what we need to create in order to produce a Tanzu Kubernetes Release (TKr).

PR #2720, opened 2021-12-14
Add TCE proposal for creating community-edition releases

joshrosso:
Proposal: https://github.com/vmware-tanzu/community-edition/issues/2717

At the time of writing, Tanzu Community Edition (TCE) relies on the Bill
of Materials (BOMs) produced for VMware's paid product, Tanzu Kubernetes
Grid (TKG). This means that creation of clusters, both management and
workload, **always** use the same packages and versions as the rest of
TKG. This binding to a TKG BOM was a decision made to ensure we could
release TCE. Now we are at stage where we must unbind from TKG's BOM and
begin producing TCE-specific BOMs.


docs/designs/2717-produce-tce-specific-releases.md
```diff
@@ -0,0 +1,223 @@
...
+* Building and packaging in the open, rather than behind VMware's network.
+* Packages that can be re-produced by users.
+* Releasing Kubernetes versions ahead of TKG.
+* Releasing core packages (e.g. antrea) ahead of TKG.
+* Releasing core-packages specific to TCE.
+* Patching TCE releases out-of-band from TKG.
+
+## Proposal
+
+### Creating a Core Package Repository
```
nrb, line 28: I'm not seeing this called out, but the current BOM implementation uses a single registry URL, and we need to be able to support multiple repositories.

I think this aligns well with having a core repository that we control, and tools for users to set up and use repositories of their own.

docs/designs/2717-produce-tce-specific-releases.md
```diff
@@ -0,0 +1,510 @@
...
+```yaml
+version: 1.25+tce.1
+corePackages:
+    - name: antrea
+      packages:
+        - name: antrea.2.1.2
+          imageBundleUri: projects.registry.vmware.com/tce/packages/antrea:2.1.2
+          options: {}
+        - name: antrea.2.5.4
+          imageBundleUri: projects.registry.vmware.com/tce/packages/antrea:2.5.4
```
nrb, line 296: I'm not understanding why each package has multiple versions specified. What process/controller picks the "correct" one for a given cluster?

docs/designs/2717-produce-tce-specific-releases.md
```diff
@@ -0,0 +1,510 @@
...
+  // the version of Kubernetes this TKR represents
+  Version string
+  // the kubernetes components used in bootstrap
+  // these should match what is inside the host images
+  // otherwise kubeadm will pull down new images on each
+  // host during bootstrap
+  Components []ContainerImage
+}
+
+type ContainerImage struct {
```
nrb, line 274: Are these listed as special case container images because the `kapp` controller isn't actually running at this point of cluster creation?

jpmcb:
Still relevant, removing stale label and marking as frozen.

PR #2763, opened 2022-01-04
Rename new standalone to unmanaged

### LINKED ISSUE #2704: Rename new standalone to unmanged-cluster
## Feature Request

For consistency's sake as well as the ability to have both the new and the old features installed simultaneously, please nename the new plugin in code and doc.

stmcginnis:

This renames the new implementation for standalone clusters to be
unmanaged-clusters.

```release-note
NONE
```

Fixes: #2704


addons/packages/harbor/2.2.3/test/go.sum
```diff
@@ -1,10 +1,14 @@
+github.com/creack/pty v1.1.9 h1:uDmaGzcdjhF4i/plgjmEsriH11Y0o7RKapEf/LDaM3w=
```
joshrosso, line 1: @stmcginnis can we revert go.sum modifications to this package. While I'm sure they're correct, we should prevent this tree from being mutated by this PR.

addons/packages/multus-cni/3.7.1/test/go.sum
```diff
@@ -1,8 +1,10 @@
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
joshrosso, line 1: @stmcginnis can we revert go.sum modifications to this package. While I'm sure they're correct, we should prevent this tree from being mutated by this PR.

hack/tagger/go.mod
```diff
@@ -4,18 +4,4 @@ go 1.16
 
 require github.com/google/go-containerregistry v0.6.0
 
-require (
```
joshrosso, line 7: @stmcginnis can we revert go.sum modifications to this package. While I'm sure they're correct, we should prevent this tree from being mutated by this PR.

joshrosso (approved):
Works great after a build for me.

PR #2764, opened 2022-01-04
Use regex for copyright header year match

stmcginnis:

Our policy is to have the copyright header reflect the year the file was
created, or a range from when it was created to the most recent
modification. With the existing header linting config in place, it would
expect that range to be updated even if there were no code changes,
which is not what we want to enforce.

This changes the linting config to use a regex that will accept either a
range or an individual year. This also matches what has been done in the
tanzu-framework repo.

```release-note
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -9,7 +9,7 @@ linters-settings:
       const:
        - LICENSE: Apache-2.0
       regexp:
-       YEAR: 20\d*-*\d*
+       YEAR: (?:20\d{2} |20\d{2}-2\d{3})
     template: |-
       Copyright {{ YEAR }} VMware Tanzu Community Edition contributors. All Rights Reserved.
       SPDX-License-Identifier: {{ LICENSE }}


nrb (dismissed):

.golangci.yaml
```diff
@@ -8,8 +8,10 @@ linters-settings:
     values:
       const:
        - LICENSE: Apache-2.0
+      regexp:
+       YEAR: 20\d*-*\d*
```
nrb, line 12: This may be going too deep, but should the regex be `20\d*-20\d*`?

A 20xx - 19xx range is invalid since it's backwards, and I doubt we'll need to worry about 100-1000 years in the future :)

I'm fine with leaving it so I gave an approval, but wanted to point this out, just in case.

jpmcb: >  doubt we'll need to worry about 100-1000 years in the future

Tanzu is beyond time Nolan 🤯 

but true, using `20\d*-20\d*` would at least give us today (`2022`) through `2099`

joshrosso: @stmcginnis are you able to update this to the more sensible limit such that we can catch edge-case typos in dates.

joshrosso (changes requested):
changes requested around limiting date range.

.golangci.yaml
stmcginnis: Done!

joshrosso (dismissed):

jpmcb (approved):
LGTM!

PR #2769, opened 2022-01-05
Bootstrap visiblity design document

### LINKED ISSUE #2730: Improve visibility into management and workload cluster creation
# Abstract

Our current bootstrapping process emits a lot of output, but it's very hard to follow what is happening. Even if you understand the overall process of deployment, parsing the output can be confusing, and it's not very clear where in the deployment process we are.

When deployment takes awhile, it's hard to tell if things are locked up or if there's something happening under the covers that we are waiting to complete.

When the deployment does fail, it's not always clear what caused the failure. We only give generic troubleshooting steps that may not even be relevant to the failure they are encountering.

**Current Issues:**

Current cluster bootstrap process is very noisy, with a lot of output that is not meaningful to the user and can be potentially confusing:

image

There are a few problems with this output:

1. There is no formatting (or a mix of formatting)
2. High level steps are not clear from low level steps being taken
3. It's hard to tell what is relevant to pay attention to, versus what can be ignored
4. Not clear what step is being done and how far along in the process it is
5. When things go wrong, not clear what the cause of the failure is
6. Not clear what the user needs to do to resolve the problem
7. Not clear if there is cleanup that needs to be done before trying again

## Issue Tracking

The following list tracks the issues required to resolve in order to achieve this capability. Please see the next section to understand the larger proposal.

### Tanzu Framework

- https://github.com/vmware-tanzu/tanzu-framework/issues/1524
- https://github.com/vmware-tanzu/tanzu-framework/issues/1662
- https://github.com/vmware-tanzu/tanzu-framework/issues/1663

### Community Edition

- TBD

### Other

- TBD

## Proposal

- [Pull Request](https://github.com/vmware-tanzu/community-edition/pull/2769)
- [Rendered View](https://github.com/vmware-tanzu/community-edition/blob/3a2ca013945b7aa0eb48a5fe0c74ced737b0b64c/docs/designs/2730-visibility.md)

stmcginnis:

This adds a design document that goes into more detail for the proposed
changes to make our bootstrap process more intuitive and easier to
understand for our end users.

```release-note
NONE
```

Fixes: #2730


docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,292 @@
...
+
+To implement these changes, the following steps will need to be taken
+
+* Review and group current output into high level steps
+* Review all output and ensure it is at an appropriate log level
+  * Low level details should be at a higher verbosity level so they are not
+    seen unless requested
+  * What is emitted at normal logging levels should be reviewed for its
+    usefulness
+  * All output should be reviewed and removed if it only serves as development
```
nrb, line 92: I'd prefer these get pushed down to a `debug` level rather than outright removed. If they're gone entirely, it can make for more needle-in-haystack problems for those who _do_ want to get to a code path level of detail.

I'm all for making sure they require the highest level of verbosity, though. Most of the time users do not need them.

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,292 @@
...
+
+#### FR3 - Output should be captured to a persistent log file
+
+If the users terminal closes or content scrolls off their buffer, they need to
+be able to get the details from the cluster creation. This is especially
+critical when something goes wrong and someone else needs to take a look at the
+output to help troubleshoot the failure.
+
+* Cluster creation output should be captured to
+  `~/.config/tanzu/tkg/bootstrap-logs/${CLUSTER_NAME}.log`
```
garrying, line 214: Wondering: Would it make sense for this to be user configurable?

stmcginnis: That's a good question. We could use an environment variable or some other persistent configuration to control where these logs are located.

For consistency sake, I would like to start out with a standardized path though. That way any debugging documentation we write (or even debug scripts) doesn't need to call out conditional path lookups. It will be fairly easy to add the capability later if we find there is a strong desire for having it configurable.

That said, please do let me know if this is a strong desire. ;)

docs/designs/2730-visibility.md
stmcginnis: Yeah, that makes sense.

garrying:
LGTM!

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+today:
+
+![current](imgs-2730/current.png)
+
+to a more structured and visually easier to parse format similar to the `kind`
+output:
+
+![kind](imgs-2730/kind.png)
+
+or the planned output for the proposed `unmanaged-cluster` command:
```
jpmcb, line 64: The _current_ output of `unamanged-cluster`? 😬

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+
+✅ Cluster created
+```
+
+Output is minimal, and can be skipped altogether if there isn't anything useful
+for the end user to know.
+
+#### FR3 - Output should be captured to a persistent log file
+
+If the users terminal closes or content scrolls off their buffer, they need to
```
nrb, line 209: Because of the scrolling issue, I think it would be helpful to output a message telling the user where the log files live.

That could potentially happen as an error handler, so we say something like `Bootstrap failed, see logs in {log_dir}`

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+be able to get the details from the cluster creation. This is especially
+critical when something goes wrong and someone else needs to take a look at the
+output to help troubleshoot the failure.
+
+* Cluster creation output should be captured to
+  `~/.config/tanzu/tkg/bootstrap-logs/${CLUSTER_NAME}.log`
+* The log should capture everything, regardless of verbosity level set by the
+  end user for console output
+* Additional debug command output, such as tailing CAPI controller logs on
+  failure, should be captured in the bootstrap log
```
nrb, line 219: This may be too deep, but something to think about - how difficult is it going to be to stream multiple sources into this logfile without contention?

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+🌐 Installing CNI
+   antrea.tanzu.vmware.com:0.13.3+vmware.1-tkg.1
+
+✅ Cluster created
+```
+
+Output is minimal, and can be skipped altogether if there isn't anything useful
+for the end user to know.
+
+#### FR3 - Output should be captured to a persistent log file
```
joshrosso, line 207: I think this is by far the most important. I'd love to see it go up to `FR1`, as inclusion of things like emojis are nice, but don't really matter.

Can you give detail on how this will be captured? What pods are you going to persist logs from?

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
+# Improved Bootstrap Visibility
```
joshrosso, line 1: What's unclear to be is whether this is for **management-cluster** bootstraping, **workload cluster** bootstrapping or both.

I think it should be for both. Would love details on what exactly we're going to scrape up or persist to provide the user details.

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### FR3 - Output should be captured to a persistent log file
+
+If the users terminal closes or content scrolls off their buffer, they need to
+be able to get the details from the cluster creation. This is especially
+critical when something goes wrong and someone else needs to take a look at the
+output to help troubleshoot the failure.
+
+* Cluster creation output should be captured to
+  `~/.config/tanzu/tkg/bootstrap-logs/${CLUSTER_NAME}.log`
```
jpmcb, line 215: `~/.config/.../*.log` feels weird to me. Would it make more sense to take advantage of most systems existing log dump directories? Usually under something like `/var/log/...` or in a user directory `/home/my-user/log/...`?

docs/designs/2730-visibility.md
joshrosso: oops just realized at the end:

> The main target for these changes are management-cluster and cluster create commands

I think we should call this out earlier. And I'd love some implementation details on what/how you're planning to slurp things up. I expect the `cluster` interaction to be more complicated.

docs/designs/2730-visibility.md
jpmcb: I wondered something similar. Maybe they would need to be timestamped files or with the intended cluster name?

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+* The main target for these changes are `management-cluster` and `cluster`
+  create commands
+* Output will be different between these two as there are different details
+  that are more relevant for one versus the other
+* The same output principles should be considered for things like `delete`
+  commands and any others that perform a multi-step process to accomplish
+
+### Security Model
+
+N/A
```
jpmcb, line 262: Is there any way to ensure we aren't logging secrets or sensitive data to the screen here? Or is that not of too much concern?

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+  `~/.config/tanzu/tkg/bootstrap-logs/${CLUSTER_NAME}.log`
+* The log should capture everything, regardless of verbosity level set by the
+  end user for console output
+* Additional debug command output, such as tailing CAPI controller logs on
+  failure, should be captured in the bootstrap log
+* To prevent hidden leaked filesystem space consumption, the bootstrap log
+  should be deleted as part of the `management-cluster delete` or `cluster
+  delete` operations
+
+#### FR4 - Debug capture on failed cluster deployment
```
joshrosso, line 224: now that we're persisting logs, should we also remind them of the bootstrap log where they can find failures?

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,293 @@
...
+UI developers should also be included in reviews to make sure there are no
+concerns from their perspective on the way changes may impact the UI.
+
+## Compatibility
+
+N/A
+
+## Alternatives
+
+An alternative - and really the current approach - would be to leave the output
```
dvonthenen, line 278: Just a general comment. Another alternative is tooling to grab relevant "stuff"... but that is also proving to be equally difficult to ramp up on. Maybe another proposal to implement better tools.

docs/designs/2730-visibility.md
jpmcb, line 207: Something I've wondered about for awhile, that is out of scope for this design, is I wonder if it'd be possible to log to a stream or socket that could push it's logs out to a service like fluentd. I wonder if there is a desire from customers to have their bootstrapping logs persisted to a observability solution.

One scenario I can imagine is clusters being bootstrapped via a pipeline and customers having a service that stores logs and can display them (like grafana's Loki). Then they can have easy access to those logs where they are natively pushed from the bootstrapper.

joshrosso: perhaps for future consideration^

docs/designs/2730-visibility.md
joshrosso: philosophically, we deft feel /var/log is more appropriate, however, for consistency sake today, we're going to stick with logs in ~/.config/tanzu/*.

docs/designs/2730-visibility.md
joshrosso: TODO: callout where the log message came from (ie which controller produced the logs)

docs/designs/2730-visibility.md
joshrosso: expand on we're relying on upstream sources not logging sensitive info.

docs/designs/2730-visibility.md
stmcginnis, line 215: I'm going to change this to `logs` instead of `bootstrap-logs` since I think this part will be useful for more than just bootstrap logging. For example, if a cluster is deployed, then later scaled, then upgraded, then deleted and the deletion failed... this log should capture all of that activity with the cluster.

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,391 @@
...
+
+#### Cluster API Controller Logs
+
+One common scenario we have found to work when troubleshooting deployment failures
+is to [look at the `capX-controller-manager` POD logs](https://tanzucommunityedition.io/docs/latest/tsg-bootstrap/#troubleshooting-manually).
+Looking for Error level logs (log lines that start with an `E`) will often indicate
+what is failing in reconcilation that is preventing the cluster from being
+properly created.
+
+Again, with having access to the `kubectl` for the cluster, we are able to
```
joshrosso, line 342:
```suggestion
Again, with having access to the kubeconfig for the cluster, we are able to
```

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,391 @@
...
+* Limited output for each grouping
+  * Raise log level of existing trace messages so they only print out if the
+    verbosity level is set high enough
+  * Lead with indentation so output is visually obvious as something underneath
+    the top level group
+  * Review the messages that are printed by default so they are very clear and
+    easy for someone to understand from the context of where they are printed in
+    the overall output
+
+#### Cluster API Details
```
joshrosso, line 280: how to you plan to expose these? and/or where do you plan to store these details?

_would be good to put this in the proposal if it's not already called out somewhere_

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,391 @@
...
+One common scenario we have found to work when troubleshooting deployment failures
+is to [look at the `capX-controller-manager` POD logs](https://tanzucommunityedition.io/docs/latest/tsg-bootstrap/#troubleshooting-manually).
+Looking for Error level logs (log lines that start with an `E`) will often indicate
+what is failing in reconcilation that is preventing the cluster from being
+properly created.
+
+Again, with having access to the `kubectl` for the cluster, we are able to
+automate this manual process, to some degree, by extracting these error logs
+and capturing them in the bootstrap log for easier troubleshooting.
+
```
joshrosso, line 345: Can we include the machine health check controller as something we gather bootstrap logs from?

stmcginnis: Good call!

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,391 @@
...
+* The same output principles should be considered for things like `delete`
+  commands and any others that perform a multi-step process to accomplish
+
+#### Process Steps
+
+The main change in output will be the selection of current log messages that are
+emitted. The things we will want to change for this process output will be:
+
+* Print high level steps
+  * Should lead with emoji if the console supports it
```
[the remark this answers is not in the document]
stmcginnis: I believe there was recent work in tanzu-framework to pick up a library that handles this. I hesitate to call that out explicitly in here, but if that has not actually beed addressed, I think making changes to improve that along with these other changes would be good, and likely not too much more effort compared to the other changes.

docs/designs/2730-visibility.md
stmcginnis: Updated!

docs/designs/2730-visibility.md
stmcginnis: Reordered. Details added to the "Implementation Details" section.

docs/designs/2730-visibility.md
stmcginnis: Added paragraph clearly stating these are things to capture to the debug log.

docs/designs/2730-visibility.md
stmcginnis: Oops :)

joshrosso (changes requested):
@stmcginnis once you address the linter issues, we're ready to approve/merge.

docs/designs/2730-visibility.md
```diff
@@ -0,0 +1,407 @@
...
+Saving file.
+Installing xyz
+Waiting for x...
+```
+
+All output is at the same level and there is no differentiation or
+relevance of each log line. The new output should have clear high level steps:
+
+```txt
+☑️  Determining cluster settings
```
swalner-vmware, line 179: I think it would be helpful to see "Determining cluster settings (step 1/7)" in the output so the user can get a sense of where they are in the process

stmcginnis: I do like this, but I wonder if we will always know up front how many steps will be in the process. I think some things might end up being added or removed dynamically based on the requested settings.

Would you be OK if we leave this out of the proposal for now, then as we solidify the implementation if we find this is actually something we can determine, add it as an implementation detail?

joshrosso (approved):

PR #2776, opened 2022-01-06
Restore existing standalone-cluster plugin

stmcginnis:

When we started working on the replacement for the standalone-cluster
plugin, we made the assumption that it would be an in-place replacement.
Based on that, the new implementation was done by modifying the existing
code.

We now need to support a deprecation period for standalone-cluster, with
the new implementation being made available as the unmanaged-cluster
plugin. In our feature development branch we have renamed the modified
standalone-cluster plugin directory to its new name of
unamanged-cluster. This restores the existing standalone-cluster plugin
source so when we merge the feature branch to the main branch we don't
lose the standalone-cluster plugin.

```release-note
NONE
```


jpmcb (approved):
👍🏼

nrb (approved):

PR #2782, opened 2022-01-07
WIP - Doc - updates for creating workload

### LINKED ISSUE #2698: [Docs] [](/docs/latest/workload-clusters/) Workload cluster setup instructions polishing
## Description

1. Duplicate links:
As shown below, the docs contain links to the workload cluster templates for each IaaS, and the links in the paragraph and the bullets are the same. Feels like duplicated info that could be condensed.

Page: https://tanzucommunityedition.io/docs/latest/workload-clusters

Excerpt:
```
To deploy a workload cluster, you create a configuration file. You then run the tanzu cluster create command, specifying the configuration file in the --file option. To see an example of a workload cluster configuration file template, see AWS Workload Cluster Template, Azure Workload Cluster Template, or vSphere Workload Cluster Template.

For specific configuration parameters, see:

AWS Workload Cluster Template
vSphere Workload Cluster Template
Microsoft Azure Workload Cluster Template
```

2. What config file to key off of:
Also, right after the reader is referred to these templates, they are instructed not to use them at all:
```
Copy the configuration file: When you deploy a workload cluster, most of the configuration for the cluster is the same as the configuration of the management cluster. The easiest way to obtain an initial configuration file for a workload cluster is to make a copy of the management cluster configuration file and to update it.
```

3. Clarify wording:
The section below talks about deploying workload clusters to namespaces in the management cluster. While this makes sense, there is an implicit assumption that the reader understands this refers to cluster.cluster.x-k8s.io CRD that will precipitate the provisioning of an actual workload cluster via ClusterAPI. If the reader is not clear on this, they may be confused by the phrasing of "deploying clusters to namespaces".

```
Create namespaces: To help you to organize and manage your development projects, you can optionally divide the management cluster into Kubernetes namespaces. You can then use Tanzu CLI to deploy workload clusters to specific namespaces in your management cluster. For example, you might want to create different types of workload clusters in dedicated namespaces. If you do not create additional namespaces, Tanzu Community Edition creates all workload clusters in the default namespace. Complete the following steps:
```

4. Typo:
On this page it says `kubectl create -foo`, which I think is supposed to be `kubectl create ns foo`
Right after that it says `kubectl get namespaces --show-labels` but doesn't mention what to look for in the labels (aka why not just `kubectl get namespaces` - what are the docs calling our attention to in the labels?)

Step 8—"Optional: Deploy a workload cluster in a specific namespace"— is interesting because it says that you must set env var NAMESPACE to install the cluster to the desired namespace. Does that mean that targeting the desired namespace in kube context is not enough, meaning that tanzu CLI will not pick up that information from the context?

Also:
I can see the value of copying the management cluster config file, but in the simplest form, it seems unnecessary since the only mandatory setting (CLUSTER_NAME) is superseded by a command line argument. If that is the only thing that is changing, you could just use the management cluster config file. Or if you want to change the namespace as well, the following would suffice:
`NAMESPACE=foo tanzu cluster create foo-workload-cluster --file /Users/somebody/.config/tanzu/tkg/clusterconfigs/abc1234567.yaml` (using the management cluster config file, not a copy of it).
This can also be used with --dry-run to generate declarative configs per cluster.
I'm not advocating that this is advisable over creating a values yaml file per workload cluster, just that the section could start with an example of the simplest command and then explain why/when it is advisable to actually have a values file per cluster.

Lastly - not sure about the importance of the following two but perhaps worth mentioning:
I wonder if there's value in relating the results of `tanzu get cluster xxx` to `kubectl get cluster xxx` or `kubectl get cluster,KubeadmControlPlane,AWSCluster` and any other resources created.
Also, a screenshot perhaps of EC2 instances in AWS console?

## Location

https://tanzucommunityedition.io/docs/latest/workload-clusters

kcoriordan:
Updates the [Deploy a workload cluster](https://tanzucommunityedition.io/docs/latest/workload-clusters/) topic with review comments from #2698

```release-note

```

Fixes: #2698

Run `hugo server `as per instructions in [readme](https://github.com/vmware-tanzu/community-edition/blob/main/docs/README.md)


kcoriordan:
closing and opening new PR on new fork

PR #2791, opened 2022-01-10
unmanaged-cluster: remove usage of grey log text

joshrosso:
This commit removes the usage of grey log text in favor of no text
stylization in bootstrapping.

This approach introduces ColorNone rather than trying to stylize under
a white value. This is because many modern terminal emulators let you
overwrite the meaning of white. This will result in the terminal's
default "no style". ColorBrightBlack has been removed since it is no
longer used; it can be re-introduces should it be used in the future.

This usage of they grey text was intended to focus the eye on
Event/Eventf commands. However, based on terminal stylization, we found
that this darker text had a high-variance in visibility. Atop that, we
should do more research in the realm of accessibility before dimming out
such text as a default value.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/log/log.go
+++ b/cli/cmd/plugin/unmanaged-cluster/log/log.go
@@ -16,11 +16,10 @@ import (
 )
 
 const (
-	ColorRed         = "\033[31m"
-	ColorBlue        = "\033[34m"
-	ColorLightGreen  = "\033[32m"
-	ColorBrightBlack = "\033[90m"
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Looks great to me, but will give @jpmcb a chance to review.

jpmcb (approved):

stmcginnis:
Looks like some linting issues to address before we merge.

joshrosso:
@stmcginnis, thanks -- lint errors addressed.

PR #2376, opened 2021-10-27
Introduce new `unmanaged-cluster` model

joshrosso:
## What this PR does / why we need it

See [this page to test this PR and understand details around its design](https://github.com/vmware-tanzu/community-edition/blob/standalone-overhaul/cli/cmd/plugin/standalone-cluster/README.md).

Please [add all feedback around this proposal to the proposal issue](https://github.com/vmware-tanzu/community-edition/issues/2266).

Please keep this PR exclusive to feedback about code.

Fixes: https://github.com/vmware-tanzu/community-edition/issues/2266

PR #2792, opened 2022-01-10
Set unmanaged-cluster aliases to agreed list

stmcginnis:

Based on [conversations](https://github.com/vmware-tanzu/community-edition/pull/2376#discussion_r781311340), we've decided to set the unmanaged-cluster
aliases to:

- um
- uc
- unmanaged

This is a minor fix to update that list of aliases that are provided on
plugin registration.


jpmcb (approved):
LGTM! 🚀

PR #2795, opened 2022-01-10
Updates deprecation notice in `standalone-cluster`

jpmcb:
Now, the standalone cluster deprecation notice mentions the new unmanaged-cluster plugin which will live side by side
with standalone

Gives simple example of how to use `unmanaged-cluster`

```release-note
NONE
```

N/a

Build with `make build-cli-plugins-loca` and tested the binaries for formatting in terminal. Example:

Screen Shot 2022-01-10 at 11 57 39 AM

N/a

### DIFF
--- a/cli/cmd/plugin/standalone-cluster/create.go
+++ b/cli/cmd/plugin/standalone-cluster/create.go
@@ -49,12 +49,14 @@ func init() {
 }
 
 func create(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


davidvonthenen (approved):

stmcginnis (approved):
Looks good! I do wonder if we should try to make it narrower to make sure it's still easy to read on smaller terminals, but that can be a follow up if we decide whether that's actually an issue or not.

PR #2800, opened 2022-01-10
Adds `Version` to unmanaged-cluster plugin descriptor

jpmcb:
Adds `Version` to the plugin descriptor for `unmanaged-cluster` using the `buildinfo.Version` compile time flag from `tanzu-framework` v0.10.0

```release-note
NONE
```

N/a

Now able to build from the `standalone-overhaul` branch using `make build-cli-plugins-local`

N/a


cli/cmd/plugin/unmanaged-cluster/main.go
```diff
@@ -19,6 +19,14 @@ var descriptor = cliv1alpha1.PluginDescriptor{
...
 	Description: description,
 	Group:       cliv1alpha1.RunCmdGroup,
+
+	// TODO: jpmcb - when tanzu framework has bumped their version of controller-runtime,
+	// we can use the newest version of the PluginDescriptor from v0.10.0+ which uses the buildinfo.Version.
+	// Right now, since we are accessing the controller-runtime package, there is a build conflict
+	// with v0.7.0 of controller runtime (which tanzu-framework uses). See the following github issue:
+	// https://github.com/vmware-tanzu/tanzu-framework/issues/444
+	// For now, we can just use a raw string for this build instead of a dynamic buildtime env-var.
+	Version: "v0.10.0",
```
stmcginnis, line 29: I wonder if it would be better to use an ldflag to do something like `-X cli/cmd/plugin/unmanaged-cluster/main.PluginVersion=${BUILD_VERSION}` in the Makefile. Assuming the addition of a `PluginVersion` variable that would then be used here to set the `PluginDescriptor.Version` field. Then this will just be set automatically based on the build being done.

jpmcb: That's definitely a possibility - we'd be introducing and owning some of that build time complexity, but I like that and I think would be much more maintainable over the long term.

I'll take a stab at it in this PR tomorrow

cli/cmd/plugin/unmanaged-cluster/go.mod
```diff
@@ -10,13 +10,14 @@ require (
...
 	github.com/vmware-tanzu/carvel-vendir v0.23.0
-	github.com/vmware-tanzu/tanzu-framework v1.4.0-pre-alpha-2.0.20210915174701-14fe0fdf4f0b
+	github.com/vmware-tanzu/tanzu-framework v0.10.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/api v0.22.2
 	k8s.io/apiextensions-apiserver v0.21.2
 	k8s.io/apimachinery v0.22.2
 	k8s.io/client-go v0.22.2
 	k8s.io/kube-aggregator v0.19.2
+	sigs.k8s.io/controller-runtime v0.9.0 // indirect
```
jpmcb, line 20: @joshrosso - Here's the crummy thing about getting this to work, seems like it has to import controller-runtime as a indirect dependency.

Looks like we're doing the same in the
- diagnostics plugin: 
- conformance plugin: https://github.com/vmware-tanzu/community-edition/blob/main/cli/cmd/plugin/conformance/go.mod#L10
- https://github.com/vmware-tanzu/community-edition/blob/main/cli/cmd/plugin/diagnostics/go.mod#L10

---

Incantation steps I took to get this to upgrade correctly and build

1. Update `tanzu-framework` version using `go get -d`
2. Ran `go mod download` and `go mod tidy` for good measure
3. Build the binary successfully with `go build -o build.out`
4. On running the binary, got the same error on the version descriptor being empty
5. Added the code to include the version via `buildinfo.Version` and a bit of logic to have a default version when non is given (like we do in `conformance` and `diagnostics`
6. I then noticed the indirect dependencies (which I _think_ I may have deleted when I originally created this PR ....)

joshrosso: Thanks for investigating this @jpmcb!

Yeah, it's really unfortunate, it appears our usage of [NewPlugin](https://github.com/vmware-tanzu/tanzu-framework/blob/a3ba874d842b4d6abc1fc8486a6f4de2cbb64eaa/pkg/v1/cli/command/plugin/plugin.go#L18-L32) brings in the following dependencies:

```sh
$ go mod why sigs.k8s.io/controller-runtime
# sigs.k8s.io/controller-runtime
github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster
github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin
github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli
github.com/vmware-tanzu/tanzu-framework/apis/config/v1alpha1
sigs.k8s.io/controller-runtime
```

By making the following changes to our `main.go`, the dependency is no longer.

```diff
 import (
        "os"

+       "github.com/spf13/cobra"
        "github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/cmd"
        cliv1alpha1 "github.com/vmware-tanzu/tanzu-framework/apis/cli/v1alpha1"
        "github.com/vmware-tanzu/tanzu-framework/pkg/v1/buildinfo"
-       "github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/command/plugin"
-       "github.com/vmware-tanzu/tanzu-framework/pkg/v1/tkg/log"
 )

 var description = `Deploy and manage single-node, static, Tanzu clusters.`
@@ -40,15 +39,12 @@ func main() {
        }

        // plugin!
-       p, err := plugin.NewPlugin(&descriptor)
-       if err != nil {
-               log.Fatal(err, "unable to initialize new plugin")
-       }
+       p := cobra.Command{}

-       p.Cmd.PersistentFlags().Int32VarP(&logLevel, "verbose", "v", 0, "Number for the log level verbosity(0-9)")
-       p.Cmd.PersistentFlags().StringVar(&logFile, "log-file", "", "Log file path")
+       p.PersistentFlags().Int32VarP(&logLevel, "verbose", "v", 0, "Number for the log level verbosity(0-9)")
+       p.PersistentFlags().StringVar(&logFile, "log-file", "", "Log file path")

-       p.AddCommands(
+       p.AddCommand(
                cmd.ConfigureCmd,
                cmd.CreateCmd,
                cmd.DeleteCmd,
```

validated via:

```
go mod why sigs.k8s.io/controller-runtime
# sigs.k8s.io/controller-runtime
(main module does not need package sigs.k8s.io/controller-runtime)
```

I feel we should really abstain from using this package until Framework can decouple it such that it does not bring in unreasonable dependencies. Interlock of dependencies in the k8s space is already extremely hard as things increment, and will be made hard if a dependency we're only using to register a plugin mandates a specific version of controller-runtime (eg in the future we can hit ever more conflicts than client-go/controller-runtime).

Looking at the callstack of NewPlugin, we should be able to re-implement this with moderate pain. Let's talk more about it tomorrow.

Thanks again!

joshrosso: @jpmcb I dug into the dependency tree this morning and it's really complicated to bring into an `internal` package.

Let's move forward with the usage of `// indirect`.

Would you mind adding a comment about that line in the go.mod stating we should remove it once https://github.com/vmware-tanzu/community-edition/issues/2811 is resolved?

jpmcb:
@joshrosso & @stmcginnis - this should be ready for review now!

nrb (approved):

nrb:
Code looks good, but failing Go style checks.

jpmcb:
> but failing Go style checks

I think that's something we'll have to address on the `standalone-overhaul` branch since we recently brought it up to speed from what was on `main` and the corresponding new linters we've enabled in the last few months

joshrosso (approved):

PR #2719, opened 2021-12-13
Create OpenAPIV3 schema for secret gen controller package and generate package with the schema

shivaani0505:
## What this PR does / why we need it
Create OpenAPIV3 schema for secret gen controller package and generate package with the schema

- [X] Add make target and corresponding script for generating open API schema for package 
- [X] Modify make target for make push-package and corresponding script for verifying the correct schema exists in package before imgpkg  bundle is pushed
- [X] Add overlay for package for openAPI schema to remove package-template.yaml
- [X] Modify the doc located at docs/site/content/docs/latest/designs/package-process.md to add the required steps
- [X] Add logic for storing the generated contents for testing schema in artifacts 
artifacts_directory

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
Create OpenAPIV3 schema for secret gen controller package and generate package with the schema
```

## Which issue(s) this PR fixes

Fixes: #2718

## Describe testing done for PR
Tested for both type of packages which contain sample values and one which don't.
Assumption: sample-values directory exists under bundle directory
Sample package tested: **secretgen-controller**
**Step 1:** Use make target for generating openAPI v3 schema

`make generate-openapischema-package PACKAGE=secretgen-controller VERSION=0.7.1`

**Step 2:**

`make push-package PACKAGE=secretgen-controller VERSION=0.7.1 TAG=0.7.1 `

Once the package is pushed successfully, copy the SHA and paste it in package.yaml at the appropriate location
Use the generated files for creating a PR.

## Special notes for your reviewer

**Note:** Doc to be updated with above make targets once initial review of scripts is done

All the above TKRs refer to version 0.7.1 of secretgen-controller , because all the current TKRs refer to that version

PR #2815, opened 2022-01-11
Move the sample value directory out of package bundle

shivaani0505:
Move the sample values directory out of package bundle.
This prevents these values being pushed as a part of image package bundle

Fixes: (https://github.com/vmware-tanzu/community-edition/pull/2719#discussion_r781689045)

Tested with and without sample values


davidvonthenen (approved):

blc1996 (approved):
LGTM

PR #2822, opened 2022-01-11
Introduce colors library for `unmanaged-cluster` logging

jpmcb:
- Added default colors to warn, error, info, etc. If a color is already set (aka, it's not the `color.Reset` attribute) it will use that color instead.

- This keeps us from having to manage escape codes ourselves and also gives us the advantage of bailing out on the use of colors when `NO_COLOR` is set in the env.

- This also replaces "successful" info messages with the `Faded` attribute. So during the progress animation, when the reconciliation or status is done, it replaces those messages with faded messages indicating success to the user.

A few screen-shots (notice the green towards the bottom and the faded text):

iTerm Dark theme
Screen Shot 2022-01-13 at 10 38 39 AM

iTerm Lite theme
Screen Shot 2022-01-13 at 10 39 55 AM

```release-note
NONE
```

N/a

See screenshots

Happy to test this with more themes etc.

The dependencies introduced here are also very minimal.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/log/log.go
+++ b/cli/cmd/plugin/unmanaged-cluster/log/log.go
@@ -13,14 +13,11 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/fatih/color"
 )
 
 const (
-	ColorNone       = "\033[0m"
-	ColorRed        = "\033[31m"
-	ColorBlue       = "\033[34m"
-	ColorLightGreen = "\033[32m"
-
 	// The following are a set of emoji codes that can be used
 	// with the Event and Eventf logging methods in this package.
 	// They should all take up 2 terminal columns.
@@ -47,8 +44,8 @@ type CMDLogger struct {
 	logLevel int
 	// instances of indentation (" ") to prepend to a long message
 	indent int
-	// color to log the message as
-	color string
+	// logColor defines the color to log the message as define by fatih/color Attributes
+	logColor color.Attribute
 	// output controls where log messages are sent
 	output io.Writer
 }
@@ -103,7 +100,7 @@ type Logger interface {
 	V(level int) Logger
 	// Style provides indentation and colorization of log messages. The indent argument specifies the amount of " "
 	// characters to prepend to the message. The color should be specified using color constants in this package.
-	Style(indent int, color string) Logger
+	Style(indent int, c color.Attribute) Logger
 	// AddLogFile adds a file name to log all activity to.
 	AddLogFile(filePath string)
 }
@@ -198,13 +195,7 @@ func (l *CMDLogger) Error(message string) {
 		return
 	}
 
-	// default to red when color is not set
-	if l.color == "" {
-		l.color = ColorRed
-		message = processStyle(l, message)
-	} else {
-		message = processStyle(l, message)
-	}
+	message = processStyle(l, message)
 	fmt.Print(message)
 }
 
@@ -213,13 +204,6 @@ func (l *CMDLogger) Errorf(message string, args ...interface{}) {
 		return
 	}
 
-	// default to red when color is not set
-	if l.color == "" {
-		l.color = ColorRed
-		message = processStyle(l, message)
-	} else {
-		message = processStyle(l, message)
-	}
 	message = processStyle(l, message)
 	fmt.Fprintf(l.output, message, args...)
 }
@@ -359,7 +343,7 @@ func (l *CMDLogger) V(level int) Logger {
 	}
 }
 
-func (l *CMDLogger) Style(indent int, color string) Logger {
+func (l *CMDLogger) Style(indent int, c color.Attribute) Logger {
 	// if tty is disable, don't return a style-capable logger
 	if !l.tty {
 		return l
@@ -369,7 +353,7 @@ func (l *CMDLogger) Style(indent int, color string) Logger {
 		level:    l.level,
 		logLevel: l.logLevel,
 		indent:   indent,
-		color:    color,
+		logColor: c,
 		output:   l.output,
 	}
 }
@@ -394,9 +378,8 @@ func processStyle(l *CMDLogger, message string) string {
 	}
 
 	// apply color value to entire message
-	if l.color != "" {
-		message = l.color + message
-		message += ColorNone
+	if l.logColor != 0 {
+		message = color.New(l.logColor).Sprint(message)
 	}
 
 	return message
--- a/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
+++ b/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
 	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/config"
 
 	v1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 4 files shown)


dvonthenen:
will this work on windows? I remember having issues with something on choco.

cli/cmd/plugin/unmanaged-cluster/log/log.go
```diff
@@ -13,13 +13,12 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/fatih/color"
 )
 
 const (
-	ColorNone       = "\033[0m"
-	ColorRed        = "\033[31m"
```
stmcginnis, line 20: Maybe I'm leaning towards over-engineering here, but would it make more sense to keep our Color constants here, but set their values to the library color values? That way if we every decide to replace `fatih/color` with something else, its usage isn't sprinkled all over our code base/

joshrosso:
> will this work on windows? I remember having issues with something on choco.

It should. We do OS detection and when using a non-compatible terminal, we disable all style.

joshrosso:
I noticed there's a `faint` attribute. Would that work for giving our indented text a bit of a fade?

Not a blocker, but curious since that's the issue that got us here.

cli/cmd/plugin/unmanaged-cluster/log/log.go
joshrosso: While I don't feel too strongly either way, I do lean towards this as it would contain the `fatih/color` import to this package. Granted a lot of other packages will surely import this one :smiley: .

jpmcb: I don't think this makes sense.

The interface that `fatih/color` gives us on individual `Color.Attributes` is a number of APIs (like `Sprint`, `Fprintf`, etc.) that write to IO writers, output a string, etc.

So if we went with something like
```go
const (
    ColorRed = Color.FgRed
    ....
)
```

further down, we would have to access the `fatih/color` Color.Attribute methods _anyways_ with something like
```
message := ColorRed.Sprint()
```

So the `ColorRed` indirection seems meaningless beyond us getting to define a single word variable; there would be no easy way to drop in and replace this library with a different one.

stmcginnis: Ah, I see what you mean. We would have to add a full abstraction layer over the top of `fatih/color` to isolate it. Not sure if that's worth it then. Probably fine to just use the import everywhere it's needed, and in the unlikely case that we decide to move to another library in the future just take care of the updates everywhere.

jpmcb: It seems like a very lite weight library and doesn't come with many dependencies that will get in our way.

Thinking abit more about this, we do sort of already have a layer of abstraction in the `log` package where we have the various logging methods (`Warnf`, `Infof`, etc. etc.) and we could easily ad colors to that if end users of the log library didn't want to import the color library and use `log.Style` to dynamically style the colors of their logs.

jpmcb: So maybe a _different_ approach I could take with this is to modify the various APIs in the `log` package we provide to use colors. And since we can be assured that they won't break in various different environments (like non-tty terminals, different color palettes, `NO_COLOR` scenarios, and even windows), end users of the library can use those without having to provide their own colors via `log.Style`

jpmcb:
@joshrosso @stmcginnis - Updated the description with the new screenshots. Thoughts?

joshrosso:
### completed text "faint" :heavy_check_mark: 

image

Making a note here that we decided to **not** encapsulate `github.com/fatih/color` in the `log` package. The trade-off here is that:

* pro: 
    * users can directory access this package and all its capabilities
    * we don't need to play a cat-and-mouse game of opening up abstractiosn as 

* con:
    * users of our lib need to import a 3rd party library and interact with it, rather than have const's available in the `log` package.

cli/cmd/plugin/unmanaged-cluster/log/log.go
```diff
@@ -180,66 +179,89 @@ func (l *CMDLogger) Warn(message string) {
...
 	fmt.Fprintf(l.output, message, args...)
+	l.logColor = color.Reset
 }
 
 func (l *CMDLogger) Info(message string) {
 	if l.logLevel > l.level {
 		return
 	}
 
+	// default to faint when color is not set
```
joshrosso, line 242: I love the progression now of how things go from "light" to "faint" -- however, I'm not too crazy about this default.

I wouldn't expect defaults or usage of ColorNone to use `faint`, which **is** a color.

For example:

```
log.Style(0, logger.ColorNone).Infof("View available packages:\n")
```

Produces faint, which is surprising to me.

Is there anyway to invert this behavior such that you need to specify faint, rather than it being the default?

jpmcb:
> users of our lib need to import a 3rd party library and interact with it, rather than have const's available in the log package.

This PR also adds default colors to `warn`, `error`, `info` so that if users really don't care about colors, they can at least get the default colors associated with those methods.

Maybe in the future, if users need, we can expand this with `success` method that prints green, etc. etc.

cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
```diff
@@ -238,11 +239,11 @@ func (t *TanzuUnmanaged) Deploy(scConfig *config.UnmanagedClusterConfig) error {
 	log.Eventf(logger.ControllerEmoji, "kubectl context set to %s\n\n", scConfig.ClusterName)
 	// provide user example commands to run
 	log.Style(0, logger.ColorNone).Infof("View available packages:\n")
```
joshrosso, line 241: with these changes, can the call to `Style` be removed?

cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
```diff
@@ -238,11 +239,11 @@ func (t *TanzuUnmanaged) Deploy(scConfig *config.UnmanagedClusterConfig) error {
...
-	log.Style(outputIndent, logger.ColorLightGreen).Infof("tanzu package available list\n")
+	log.Style(outputIndent, color.FgGreen).Infof("tanzu package available list\n")
 	log.Style(0, logger.ColorNone).Infof("View running pods:\n")
```
joshrosso, line 243: with these changes, can the call to `Style` be removed?

cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
```diff
@@ -238,11 +239,11 @@ func (t *TanzuUnmanaged) Deploy(scConfig *config.UnmanagedClusterConfig) error {
...
-	log.Style(outputIndent, logger.ColorLightGreen).Infof("kubectl get po -A\n")
+	log.Style(outputIndent, color.FgGreen).Infof("kubectl get po -A\n")
 	log.Style(0, logger.ColorNone).Infof("Delete this cluster:\n")
```
joshrosso, line 245: with these changes, can the call to `Style` be removed?

joshrosso (changes requested):
Requested some changes (or help with my understanding)

Also would like to see @garrying validate this before we merge. Thanks!

Overall looks **really great**.

cli/cmd/plugin/unmanaged-cluster/log/log.go
jpmcb: > `log.Style(0, logger.ColorNone).Infof("View available packages:\n")`
> Produces faint, which is surprising to me.

Ah - that is surprising! I hadn't considered the `log.Style(...).{Method}` workflow when adding defaults.

Yes, I think the alternative is to invert the behavior, otherwise, we'll be overriding people's color choice. Good catch!

garrying:
Noticed that `Download(ed/ing) to:` is a different line color in the TKR retrieval step. Not sure if it's intentional:
Screen Shot 2022-01-13 at 4 06 33 PM

`Downloading to:` repeated for me on a few occasions. Maybe something to do with resolving the request? 
Screen Shot 2022-01-13 at 4 08 00 PM

Cycled through a few themes. The color output looks great!

jpmcb:
@garrying good catch! Just pushed up a fix, give that a try and see how that renders for you.

joshrosso:
### TTY and style still disabled in command prompt :heavy_check_mark: 

image

### TTY and style still disable in powershell :heavy_check_mark: 

image

### TTY and style works great in windows terminal :heavy_check_mark: 

image

joshrosso (approved):
not automerging in case you want validation from @garrying.

garrying:
> @garrying good catch! Just pushed up a fix, give that a try and see how that renders for you.

`Downloaded to:` is on faint now. `Download to:` rendering is unchanged. How it looks for me:
Screen Shot 2022-01-13 at 6 26 35 PM

Totally cool to defer if this is ready to merge!

joshrosso:
Thanks @garrying.

This is an issue, but I believe independent of introducing this color library.

I've opened https://github.com/vmware-tanzu/community-edition/issues/2852 to track this bug.

PR #2829, opened 2022-01-12
Address lint issues

stmcginnis:

With the merge into this development branch from main, we've picked up
the latest golangci-lint settings. These point out a few issues that
were not found with the older version of the linters. We've also been
ignoring a few things, but since we've switched from golint to revive
for some checks they were being highlighted again.

To address new issues and to stop ignoring others, this makes various
minor updates to avoid linting complaints.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
+++ b/cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
@@ -36,18 +36,16 @@ const (
 	kappControllerDeploymentName = "kapp-controller"
 )
 
-// KappClient is a client for interfacing with kapp-controller.
-//nolint:golint
-type KappClient struct {
+// Client is a client for interfacing with kapp-controller.
+type Client struct {
 	dynClient  dynamic.Interface
 	clientSet  kubernetes.Interface
 	restMapper meta.RESTMapper
 	scheme     *runtime.Scheme
 }
 
-// KappInstallOpts contains information about how to install kapp-controller.
-//nolint:golint
-type KappInstallOpts struct {
+// InstallOpts contains information about how to install kapp-controller.
+type InstallOpts struct {
 	// MergedManifests represents the final (serialized) data for installing kapp-controller and its ancillary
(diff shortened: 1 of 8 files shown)


cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
```diff
@@ -62,20 +60,19 @@ type KappInstallOpts struct {
...
+	Install(opts InstallOpts) (*v1.Deployment, error)
 	// Status retrieves the pod status for kapp-controller. It expects to be passed the namespace and name for the
 	// kapp-controller Deployment object. If it cannot talk to the cluster, that status is reported. If the
 	// pod cannot be resolved, a status of not created is reported. Otherwise, the exact status message is returned.
 	Status(ns, name string) string
 }
 
 // New instantiates a new KappManager.
-func New(kubeconfigBytes []byte) (KappManager, error) {
+func New(kubeconfigBytes []byte) (Manager, error) {
```
jpmcb, line 75: I'm _alittle_ worried about overloading the term `Manager` throughout the codebase here

So far, it seems like we have a `Manager` for the `cluster`, `kapp`, `kubeconfig`, `tanzu`, etc which _may_ be confusing as a developer working in multiple packages of the library.

But i suppose it _does_ make more sense from an API end user perspective to call something as `*.Manager` with it's package name like `kapp.Manager` instead of the more verbose `{package}.{package}Manager`

🦆

jpmcb (approved):
Glad we can get rid of the `strconv` import in the `log` package! makes much more sense. Overall this makes sense to me. Thanks Sean!

cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
joshrosso: > But i suppose it does make more sense from an API end user perspective

I hold this perspective. Similar to how go recommends using `New()` in a package (e.g. dog) rather than `dog.NewDog` since the package new effectively the prefix.

joshrosso (approved):

PR #2839, opened 2022-01-13
unmanaged-cluster: add line break after error output

joshrosso:

This inserts a line break when an error is returned to the calling
client (cmd packag). It improves readability as PS1 of the terminal will
not start at the end of the error message.

Prior to this change, outputs like the following occurred.

image

image

With these changes, we now see an appropriate break, improving readability.

image

This is a minor enough change where-in if @jpmcb or @stmcginnis do not have any suggestions, it can be approved and merged.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
@@ -106,7 +106,7 @@ func create(cmd *cobra.Command, args []string) error {
 	tm := tanzu.New(log)
 	err = tm.Deploy(clusterConfig)
 	if err != nil {
-		log.Error(err.Error())
+		log.Errorf("%s\n", err.Error())
 		return nil
 	}
 
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
@@ -42,7 +42,7 @@ func list(cmd *cobra.Command, args []string) error {
 	tClient := tanzu.New(log)
 	clusters, err := tClient.List()
 	if err != nil {
-		return fmt.Errorf("unable to list clusters. Error: %s", err.Error())
+		return fmt.Errorf("unable to list clusters. Error: %s\n", err.Error())
 	}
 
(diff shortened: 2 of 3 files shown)


cli/cmd/plugin/unmanaged-cluster/cmd/list.go
```diff
@@ -42,7 +42,7 @@ func list(cmd *cobra.Command, args []string) error {
 	tClient := tanzu.New(log)
 	clusters, err := tClient.List()
 	if err != nil {
-		return fmt.Errorf("unable to list clusters. Error: %s", err.Error())
+		return fmt.Errorf("unable to list clusters. Error: %s\n", err.Error())
```
stmcginnis, line 45: Should the line break be in the error message itself? It seems like it would be better to have that controlled by whatever is actually logging the error.

jpmcb: The `list` func her _is_ the top most controlling function that gets passed to `cobra`, so the error formatting should happen at this level.

cli/cmd/plugin/unmanaged-cluster/cmd/create.go
```diff
@@ -106,7 +106,7 @@ func create(cmd *cobra.Command, args []string) error {
 	tm := tanzu.New(log)
 	err = tm.Deploy(clusterConfig)
 	if err != nil {
-		log.Error(err.Error())
+		log.Errorf("%s\n", err.Error())
```
jpmcb, line 109: An alternative here would be to change the `log.Error` method to:

```diff
+ fmt.Println(message)
- fmt.Print(message)
```

but I'm also good with your change

cli/cmd/plugin/unmanaged-cluster/cmd/list.go
stmcginnis: Ah, missed that part!

cli/cmd/plugin/unmanaged-cluster/cmd/create.go
joshrosso: let's revisit this option as an enhancement to the log package.

PR #2841, opened 2022-01-13
Fork output writer to avoid bold table headers

stmcginnis:
We had been using the OutputWriter from tanzu-framework. In the release
we are using, the header values for table columns in that version would
bold the text, resulting in garbage characters when doing things like
`watch` to wait for conditions in the output. This has since been fixed,
but is not available in a tanzu-framework release we can use yet.

Until we are able to bump up to a version of tanzu-framework that has
the fix, this temporarily duplicates that code locally. This can be
removed in a future release.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
@@ -8,9 +8,9 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/internal/hack"
 	logger "github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/log"
 	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/tanzu"
-	"github.com/vmware-tanzu/tanzu-framework/pkg/v1/cli/component"
 )
 
 const listDesc = `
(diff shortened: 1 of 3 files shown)


cli/cmd/plugin/unmanaged-cluster/log/output.go
```diff
@@ -0,0 +1,231 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
joshrosso, line 1: @stmcginnis can you put this in `unmanaged-cluster/internal/hack`. I want to ensure:

1. The package name really stands out as a workaround.
2. The package is not importable and APIs/datatypes/public functions cannot be coded against.

This will add an extra import in the `cmd` package, but this is a good thing IMO.

Let me know if this is not possible.

joshrosso:
## `ls` no longer produces bold column headers :heavy_check_mark: 

```
watch ./tanzu-unmanaged-cluster-linux_amd64 ls
```

```
watch 

Every 2.0s: ./tanzu-unmanaged-cluster-linux_amd64 ls                                         u: Thu Jan 13 17:57:00 2022

  NAME   PROVIDER
  doggo  kind
```

joshrosso (changes requested):
Once we move this to the internal package, requested [here](https://github.com/vmware-tanzu/community-edition/pull/2841/files#r784199792), this is ready to merge!

joshrosso (approved):
Thanks for moving the package. Re-compiled off `d26dc9d6e770b3a90bdb023f120dd1b3d7076a52` and verified the bold is fixed!

image

joshrosso:
@stmcginnis would you fix the small merge conflict then self-merge?

thanks!

PR #2850, opened 2022-01-13
Change tty control flag to `--tty-disable`

stmcginnis:
Bool flags for cobra commands are defaulted to true when present, and
the syntax of needing an = to set the value to something other than that
isn't the most intuitive.

Rather than have our tty control flag default to true, requiring the
user to explicitly set the value to switch it to false, this simplifies
things by changing the flag to `--tty-disable`. That makes its usage a
little simpler in that when it is actually used, the intent of using it
is to disable the TTY formatting in our output. So now it can just be
used as a simple flag.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
@@ -34,7 +34,7 @@ func init() {
 	ConfigureCmd.Flags().StringVarP(&co.cni, "cni", "c", "", "The CNI to deploy. Default is 'antrea'")
 	ConfigureCmd.Flags().StringVar(&co.podcidr, "pod-cidr", "", "The CIDR to use for Pod IP addresses. Default and format is '10.244.0.0/16'")
 	ConfigureCmd.Flags().StringVar(&co.servicecidr, "service-cidr", "", "The CIDR to use for Service IP addresses. Default and format is '10.96.0.0/16'")
-	ConfigureCmd.Flags().BoolVar(&co.tty, "tty", true, "Specify whether terminal is tty; set to false to disable styled output.")
+	ConfigureCmd.Flags().Bool("tty-disable", false, "Disable log stylization and emojis")
 }
 
 func configure(cmd *cobra.Command, args []string) error {
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
@@ -5,8 +5,6 @@
 package cmd
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 
 	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/config"
@@ -22,7 +20,6 @@ type createUnmanagedOpts struct {
 	cni                    string
 	podcidr                string
 	servicecidr            string
-	tty                    bool
 	portMapping            []string
 }
 
@@ -48,6 +45,7 @@ var CreateCmd = &cobra.Command{
 	Short: "Create an unmanaged Tanzu cluster",
 	Long:  createDesc,
 	RunE:  create,
+	Args:  cobra.MaximumNArgs(1),
 }
 
 var co = createUnmanagedOpts{}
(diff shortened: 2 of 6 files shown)


joshrosso (changes requested):
@stmcginnis I'm finding `--tty false` to no longer error out, but also no longer be respected.

image

Thoughts?

stmcginnis:
> Thoughts?

Yep. With the way you provided this as `--tty false` and no cluster name, cobra is interpreting this as "set tty to true" and "cluster name is 'false'. It's a syntax problem with how cobra handles that switch. Unless you specify `--tty=false`, it will not work the way you are expecting.

I left a comment on Slack, but wonder since the use of this flag will usually be to disable tty, maybe we should invert things and rename it to `--no-tty`.

joshrosso:
> 

Ooops, yeah that makes sense. Didn't realize I left out the cluster name (arg)!

However, i'm a little surprised, so cobra **cannot** evaluate `--tty false`? It needs to be `tty=false`?

stmcginnis:
>However, i'm a little surprised, so cobra cannot evaluate --tty false? It needs to be tty=false?

That is my understanding. Maybe @jpmcb can confirm, but I had to walk through in a debugger to confirm that `--tty false` was setting the tty variable to `true` and `false` was one of the `args` being passed through to the cmd.

joshrosso:
Perhaps it would be more clear and best practice to introduce an argumentless flag. 

For example: `--tty=false` to `--tty-disable`.

Let's get @jpmcb input as to whether this is a good practice. To me, `--tty-disable` seems more clear.

joshrosso (changes requested):
preventing merge while we has these UX details out.

jpmcb:
Sean is correct; this is a symptom of how `spf13/pflag` is implemented which is what Cobra uses for persistent flags.

`spf13/pflag` is doing a best effort here with what it's given as a "default" value:

- If it finds `--tty`, it's going to treat this as a switch flag and set this to true (which is already the default value). Note here that it won't attempt to interpret any next "word" as an argument to this flag since it's condition has already been satisfied as a boolean flag. 
- If it finds _no `--tty` flag, it's going to use the default value we've set which is `true`.
- If it finds `-tty={value}` it's going to attempt to use that value as the new flag bool.

Again, in the case where it is seeing just `--tty`, it's going to interpret that as just a boolean switch, take the argument after and use that as the argument to the create command.

There's actually a good little chunk about what's happening here and the possibility to change the default behavior given no flag is set the pflag library: https://github.com/spf13/pflag#setting-no-option-default-values-for-flags

I agree with Josh, I think the best thing to do here is set a `--tty-disable` flag that acts as an inverse switch to what we were doing and turns TTY off.

joshrosso:
Thanks for the details @jpmcb.

@stmcginnis can we move forward with, `--tty-disable`. I prefer it over `--no-tty` because i'd like the noun `tty` to come before the verb `disable`.

stmcginnis:
Updated, but now I see there are merge conflicts. Rebase on its way.

joshrosso:
thanks! I'll review and test this shortly.

joshrosso:
### `--tty-disable` is now available :heavy_check_mark: 

```
josh@u:~$ ./tanzu-unmanaged-cluster-linux_amd64 create --help

Create an unmanaged Tanzu cluster. This sets up a Kubernetes cluster and installs
Tanzu packages. Once the environment is bootstrapped, your kubectl context is
automatically switched, enabling you to begin using the kubectl and tanzu CLIs.

No configuration is required for this command. However, you can setup a config
file by running tanzu unmanaged-cluster configure <cluster name>. This config file can
then be referenced using the -f flag.

When create is called, it respects the following precedence for all configuration:

1. flags (most respected)
2. environment variables
3. config file
4. defaults (least respected)

Usage:
  tanzu unmanaged-cluster create <cluster name> [flags]

Flags:
  -c, --cni string            The CNI to deploy; default is antrea
  -f, --config string         A config file describing how to create the Tanzu environment
  -h, --help                  help for create
      --pod-cidr string       The CIDR for Pod IP allocation; default is 10.244.0.0/16
  -p, --port-map strings      Ports to map between container node and the host (format: '80:80/tcp' or just '80')
      --provider string       The infrastructure provider for cluster creation; default is kind
      --service-cidr string   The CIDR for Service IP allocation; default is 10.96.0.0/16
      --skip-preflight        Skip the preflight checks; default is false
  -t, --tkr string            The URL to the image containing a Tanzu Kubernetes release
      --tty-disable           Disable log stylization and emojis

Global Flags:
      --log-file string   Log file path
  -v, --verbose int32     Number for the log level verbosity(0-9)
```

### default create uses stylization :heavy_check_mark: 

image

### `--tty-disable` removes stylization :heavy_check_mark: 

image

### TTY autodetect still works (validated in cmd prompt, windows) :heavy_check_mark: 

image

joshrosso (approved):

PR #2865, opened 2022-01-18
Bump go version to 1.17

stmcginnis:

The tanzu-framework repo has bumped its version of go to 1.17 [0].
This follows suite and updates the go version in all go.mod and
github workflows to specify go 1.17.

[0] https://github.com/vmware-tanzu/tanzu-framework/commit/b2816e1ae30f3d8cf0c201eda05fcd89ea6c5924

```release-note
Go version has been raised to go 1.17.
```


ankeesler (approved):

christianang (approved):

12345lcr (approved):

Makefile
```diff
@@ -217,7 +217,7 @@ release-docker: ### builds and produces the release packaging/tarball for TCE in
 		-w /go/src/community-edition \
 		-v ${PWD}:/go/src/community-edition \
 		-v /tmp:/tmp \
-		golang:1.16.2 \
+		golang:1.17.6 \
```
karuppiah7890, line 220: Should we just use `golang:1.17` instead of fixing to a particular version? I mean, it's not bad to get the latest golang patch version from `1.17.x`, but yeah, releasing TCE with `release-docker` multiple times could produce different binaries with different behavior if done that way which is a con

karuppiah7890 (approved):
LGTM! Thanks a lot for doing this 😁

Also, the first thing that came to mind related to this was #1487 , but I guess that issue is still tricky because we still use an older version of golangci-lint and plan to do so to avoid unexpected addition of linters

https://github.com/vmware-tanzu/community-edition/blob/db51111f367e85e3f9e35b39af09d8df96785b6e/.golangci.yaml#L140

Gotta check if there are issues in golangci-lint in my local when I use `go v1.17.6` with `golangci-lint v1.38.0`

EDIT - Ah, but in terms of tooling, we still get latest I guess? And only in runtime the golangci-lint version is used? Not sure how that works together - just linters from older version? Idk

https://github.com/vmware-tanzu/community-edition/runs/4856667757?check_suite_focus=true#step:4:9 shows v1.43.0

and we get the latest version here for local and CI -

https://github.com/vmware-tanzu/community-edition/blob/db51111f367e85e3f9e35b39af09d8df96785b6e/hack/tools/Makefile#L58-L60

Makefile
jpmcb: I think pinning to a patch and having reproducible builds is probably more desirable here vs using a docker base image with a floating patch version

hack/builder/go.mod
```diff
@@ -1,13 +1,110 @@
...
 // replace github.com/vmware-tanzu/tanzu-framework => ../../../tanzu-framework
 
 require (
-	github.com/mattn/go-runewidth v0.0.9 // indirect
 	github.com/spf13/cobra v1.2.1
 	github.com/vmware-tanzu/tanzu-framework v0.10.0
+)
+
+require (
+	cloud.google.com/go v0.81.0 // indirect
```
jpmcb, line 14: Hmm curious why there are suddenly all the new indirect dependencies throughout? Did something in 1.17 change with how go mod resolves dependencies?

stmcginnis: Yep, there are changes in the module graph handling. Take a look at the "Pruned module graphs in go 1.17 modules" section in the 1.17 release notes here: https://go.dev/doc/go1.17

jpmcb: That's sweet! Good to know!! (and very glad they have made the module graph hopefully less of a pain)

davidvonthenen (approved):
seems good to me

jpmcb (approved):

PR #2879, opened 2022-01-20
Document adding additional RBAC for external-dns

christianang:

This changes the wording for a known issue with the external-dns package for sources that don't have proper RBAC setup. This makes the text more generally applicable to any non-default source that the package does not handle, but is a source a user may want to use.

```release-note
NONE
```

Fixes: n/a

none


mcwumbly (approved):
lgtm!

PR #2884, opened 2022-01-20
[E2E] add `BUILD_VERSION` to vSphere E2E test script invocations and allow manual trigger of E2E test workflows

### LINKED ISSUE #2883: vSphere E2E test failing due `BUILD_VERSION` not being set
We have two pipelines failing with this error

https://github.com/vmware-tanzu/community-edition/actions/runs/1721486069

https://github.com/vmware-tanzu/community-edition/actions/runs/1721486071

The issue is due to the fact that we now download and use daily build by default for vSphere E2E tests and if it's not present only then we build TCE using `make release`. And for using the daily build, we need the `BUILD_VERSION` parameter

This is due to the change that happened here #2712 which also had a similar issue with `BUILD_VERSION` with the publish-daily-build.sh script but that was fixed in #2745 but it's not used in vSphere E2E test script invocations, and all this time the vSphere E2E tests were disabled due to failures that weren't debugged / fixed, now that it's enabled we noticed this issue just now

karuppiah7890:

Adds `BUILD_VERSION` to vSphere E2E test script invocations, and allow manual trigger of E2E test workflows for cluster deployment tests and also diagnostics plugin E2E test

```release-note
NONE
```

Fixes: #2883 

None

None

### DIFF
--- a/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
+++ b/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
@@ -1,6 +1,7 @@
 name: E2E Test - AWS Management and Workload Cluster
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
--- a/.github/workflows/e2e-aws-standalone-cluster.yaml
+++ b/.github/workflows/e2e-aws-standalone-cluster.yaml
@@ -1,6 +1,7 @@
 name: E2E Test - AWS Standalone Cluster
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
--- a/.github/workflows/e2e-azure-management-and-workload-cluster.yaml
+++ b/.github/workflows/e2e-azure-management-and-workload-cluster.yaml
@@ -1,6 +1,7 @@
 name: E2E Test - Azure Management and Workload Cluster
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
--- a/.github/workflows/e2e-azure-standalone-cluster.yaml
+++ b/.github/workflows/e2e-azure-standalone-cluster.yaml
@@ -1,6 +1,7 @@
 name: E2E Test - Azure Standalone Cluster
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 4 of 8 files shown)


.github/workflows/e2e-aws-management-and-workload-cluster.yaml
```diff
@@ -1,6 +1,7 @@
 name: E2E Test - AWS Management and Workload Cluster
 
 on:
+  workflow_dispatch:
```
dvonthenen, line 4: I think you are missing something here like and event?

dvonthenen: Maybe some additional context, the github action to trigger this event in the daily build uses the `repository_dispatch`.

karuppiah7890: Ah okay, I missed to add details about this in the PR description, my bad. I added this to enable manual trigger of the pipeline . The manual trigger can help anyone trigger from web or CLI manually, so it's not based on any other event like `daily-build`, but one can introduce inputs for manual triggers but in this case I don't see any inputs. More info - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch , https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

karuppiah7890: I wanted to add this manual trigger feature for all pipelines, so that one can trigger it without other events (like push, etc)

But in some workflows, like vSphere one, there's a `repository_dispatch` where we can trigger by simulating / emitting that event using the GitHub API without the official daily build triggering it with the https://github.com/peter-evans/repository-dispatch action

karuppiah7890: Let me know if that makes sense

karuppiah7890: It will look something like this for example (my fork where I have complete access) - 

Screenshot 2022-01-27 at 5 08 08 PM

dvonthenen: ok gotcha makes sense!

davidvonthenen (approved):

PR #2887, opened 2022-01-20
Enables dependabot version upgrade

jpmcb:
This yaml config should enable dependabot version upgrades for our many go modules. This will open bot PRs for upgradeable dependencies in our go mods.

As documented here: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-dependabot-version-updates

```release-note
Enables dependabot PRs
```

N/a

Will have to wait and see it work in GitHub 👀 

N/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,210 @@
+# Dependabot config for go mod version updgrades
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "hack/builder"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "tanzu-framework"
+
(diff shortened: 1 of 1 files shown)


nrb:
Should we exclude tanzu-framework, since we know we're not going to be actively updating to every new version?

jpmcb:
> Should we exclude tanzu-framework

That's a good point. Especially since we update those manually for our releases. Looking at a way to ignore those now

stmcginnis (approved):
This looks good to me. Will wait to merge in case anyone else wants to take a look, but I think we can just address any issues in follow up patches.

nrb:
Yeah, let's merge and see what shakes out :)

PR #2877, opened 2022-01-19
Adds GitHub action stale bot

jpmcb:
## What this PR does / why we need it
Adds `stale` label to PRs and issues after 30 days
and closes them after 15 days.

Avoids the stale label for PRs with milestones attached.

Ref GitHub stale action: https://github.com/actions/stale

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
Added github actions stale bot to label / close stale issues and PRs
```

## Which issue(s) this PR fixes
Fixes: #2876

## Describe testing done for PR
N/a

## Special notes for your reviewer
n/a

PR #2900, opened 2022-01-21
Introduce "never-stale" label exemption to stalebot

jpmcb:
Since stalebot will try to make stale and close issues/prs
that correlate with tanzu-framework tracking issues,
this introduces the use of "never-stale" to ignore those issues.

```release-note
stale-bot now ignores issues tracking tanzu-framework issues
```

N/a - followup to https://github.com/vmware-tanzu/community-edition/pull/2877

N/a wait and see what the action does

N/a

### DIFF
--- a/.github/workflows/stale-bot.yaml
+++ b/.github/workflows/stale-bot.yaml
@@ -19,3 +19,5 @@ jobs:
           days-before-pr-close: 30
           # Don't add stale label to PRs / issues with milestones "upcoming" attached.
           exempt-milestones: "upcoming"
+          # Don't add stale label to PRs / issues with this label
+          exempt-issue-labels: "never-stale"


stmcginnis:
Upstream uses `lifecycle/frozen`. Maybe we should do the same?

jpmcb:
I'm up for that! I have no preference! 

@joshrosso thoughts on the label name here?

joshrosso:
Since we don't use the `lifecycle/` prefix today, I recommend we don't introduce it. However, I'm open to it if ya'll wanna introduce and define the value of `lifecycle/` in this project.

.github/workflows/stale-bot.yaml
```diff
@@ -19,3 +19,5 @@ jobs:
           days-before-pr-close: 30
           # Don't add stale label to PRs / issues with milestones "upcoming" attached.
           exempt-milestones: "upcoming"
+          # Don't add stale label to PRs / issues that are associated with TF
```
joshrosso, line 22:
```suggestion
          # Don't add stale label to PRs / issues with this label
```

jpmcb:
I'm not passionate either way; maybe in the future, as the need arises to introduce `lifecycle/` labels, we can refactor this to include that label?

joshrosso (approved):

joshrosso:
I think `never-stale` is good.

PR #2904, opened 2022-01-21
`unmanaged-cluster` log package: Non-format log methods print new-line

### LINKED ISSUE #2873: `unmanaged-cluster` non-format logging methods should include new lines
## Feature Request

Currently, within the `unmanaged-cluster` logging package, users must include a new line themselves when logging messages:

Here's an example from the cluster create command:

https://github.com/vmware-tanzu/community-edition/blob/f1e98b04554d7ae3ea3daff95a81a75c74e5d6d8/cli/cmd/plugin/unmanaged-cluster/cmd/create.go#L108

As an enhancement, we should make this interface alittle nicer for users and include methods that will print newlines automtaically without the need to use the `format` logging methods.

So in this example, the user would just use something like:
```
log.Error(err.Error())
```
and the current `log.Error` method will also print the new line.

We could just expect people to use the `format` option of the logging methods we provide.

## Additional context

Discussed here: https://github.com/vmware-tanzu/community-edition/pull/2839#discussion_r784175060

jpmcb:
Refactors the non-format style methods in the logging package
to print a newline at the end of their message.

This simplifies the logging API where before, code lines like:
```go
log.Errorf("%s\n", myError)
```

can now be simplified to:
```go
log.Error(myError)
```

Users can still use the format style alternatives
to provide a format string and a newline at the end if they prefer.

```release-note
unmanaged-cluster log package non-format methods now print newline
```

Fixes: #2873

Built the plugin binaries and all still looks good!

image

@stmcginnis I also removed this line in the `cmd/list.go`:
```diff
t.Render()
-log.Info("")
```
and I couldn't for the life of me figure out what it was supposed to be doing since new lines were not added before this PR. So my assumption is it was something left over

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
@@ -105,7 +105,7 @@ func create(cmd *cobra.Command, args []string) error {
 	tm := tanzu.New(log)
 	err = tm.Deploy(clusterConfig)
 	if err != nil {
-		log.Errorf("%s\n", err.Error())
+		log.Error(err.Error())
 		return nil
 	}
 
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/list.go
@@ -53,7 +53,6 @@ func list(cmd *cobra.Command, args []string) error {
 		t.AddRow(c.Name, c.Provider)
 	}
 	t.Render()
-	log.Info("")
 
 	return nil
 }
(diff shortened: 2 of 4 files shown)


stmcginnis:
Awesome, testing this now.

> and I couldn't for the life of me figure out what it was supposed to be doing since new lines were not added before this PR. So my assumption is it was something left over.

This was actually added based on your feedback to another PR that when you ran a command with `-o json`, there was no break between the end of the json output and your terminal prompt. I'll double check how things look with this change.

jpmcb:
> there was no break between the end of the json output and your terminal prompt. I'll double check how things look with this change.

Oh that's right. Very weird, I can no longer reproduce it ...

Here's a screen grab in a native terminal:
Screenshot from 2022-01-21 12-29-29

And a grab from my tmux session:
image

maybe a red herring?

stmcginnis (approved):
Built and tested locally. All output looks good.

Issue #2848, opened 2022-01-13
Introduce an Application Toolkit Package


kartiklunkad26:
# Proposal: Introduce an Application Toolkit Package

# Summary
We are introducing an Application Toolkit Meta-Package to simplify the installation of all the packages that are important for a developer to create and iterate on Tanzu workloads. 

# Problem Statement

The introduction of carvel packaging technology has significantly improved the installation experience of tools on Kubernetes. We've also seen a wide adoption of the packaging technology for several tools on TCE. 

Installing each of the individual packages manually can be cumbersome and error-prone. 

# Proposal

**Meta-Package**

A meta-package is a set of components (Carvel packages). Once that meta-package is defined, a profile can be implemented based on the need to enable an intent-based installation experience of some or all of the specified packages.

Further details regarding the Metapackage as a concept has been introduced in the [Design Doc](https://github.com/vmware-tanzu/community-edition/pull/2911)

We are introducing an **App-Platform** (name can still change) metapackage which will initially consists of the below packages. This will expand in the future. 

**Phase 1 Packages**

| Packages                |
| ----------------------- |
| Contour                 |
| Cert-manager            |
| Source Controller       |
| knative-serving         |
| kpack                   |
| Cartographer            |

**User Experience**

Here is the flow of the installation experience using the developer profile of Application Toolkit Meta-Package 

1. Listing all the packages as they relate to Application Toolkit (there can be additional non-Application Toolkit packages here)  

`tanzu package available list`

```
NAME                                     DISPLAY-NAME             SHORT-DESCRIPTION
  
app-toolkit.tanzu.vmware.com             Application Toolkit             Package to install a set of Application Toolkit components
contour.tanzu.vmware.com                 Contour                  Used as a Ingress Controller
cert-manager.tanzu.vmware.com            Cert-Manager             Used to install certificates
source-controller.tanzu.vmware.com       Source Controller        Used to watch the git repositories
knative-serving.tanzu.vmware.com         Knative Serving          Used to convert images to Knative services
kpack.tanzu.vmware.com                   Kpack                    Used to convert source code into images
cert-injection.webhook.tanzu.vmware.com  Cert-Injection Webhook   Used to inect certificates into the running containers
cartographer.tanzu.vmware.com            Cartographer             Used to create software supply-chains
ootb-basic-supply-chain.tanzu.vmware.com OOTB Basic Supply Chain  Used to install a basic supply chain going from git repository to running Knative Service. 
```

[...]

Issue #322, opened 2021-03-15
Install a cluster with a set of addons and offer profiles that combine and configure multiple packages


jorgemoralespou:
Would it be possible to bootstrap cluster with some addons already applied/installed? Kind of a plan where I can do `tanzu cluster create --plan=appstack` or ` tanzu cluster create --plan=serverless -v domain.name=mydomain.com` that will get me a cluster with all the appstack or serverless required addons installed?

PR #2911, opened 2022-01-21
Design Doc for Metapackage Proposal

kartiklunkad26:

This PR provides the Design Doc for the Metapackage technology which is the basis for the profile-based installation experience in #2848

Provides design doc for https://github.com/vmware-tanzu/community-edition/issues/2848


jorgemoralespou:
One comment I would make to this, having probably done the first metapackage and then iterated over it, and then looking at the TAP ones, is that I would make configuration more user friendly and usecase oriented, rather than package oriented and then have a proper way to translate this "user friendly" configuration to the real package configuration under the hood.

Sane defaults and enumerated configuration (configuration that packages a combined set of configuration values and that is limited and predefined) should be preferred, IMHO. An example would be to have something like `provider: [aws,docker,azure,vsphere] and based on one of those 4 values set the best sane defaults for all other packages. When you look for example to external-dns you can easily create the optimal  configuration for the user based on this instead of telling him to provide a lot of information. Just compare these 2 packages configuration [my own external-dns](https://github.com/failk8s-packages/external-dns-package#global) and [tce's bundled external-dns](https://github.com/vmware-tanzu/community-edition/tree/main/addons/packages/external-dns/0.8.0#external-dns-configuration). The latter is package oriented and imposes the user to understand how external-dns is configured. The former is user oriented, and imposes the user understand where he's using it, but not how the package is configured.

jorgemoralespou:
Reading the proposal, will leave here some more comments:
- I would not use `Profile` as a concept for metapackages. It's very confusing and doesn't really tell what the user is doing. Not sure a better word, as to me this would be `Package` or `bundle` as a set of "canned packages that are shipped together when the XXX is selected"
- Always provide a list of packages to add or remove to a Profile, so that it can be tweaked (augmented/reduced)
- Provide a first class citizenship to `Profiles` (or whatever that end up being called) so that I can list the profiles a metapackage provides, and potentially a description. Currently this is not possible, but to me this is key, so if Carvel and or Package plugins can't be augmented, then instead of using profiles, create different metapackages per profile. (vendir can help de-duplicate code)
- As such, I would want to have a values-file being created based on the profile I will install, so that I don't see config that is not relevant to such profile.
- With regards to this section "Risks and Mitigations: UX has already been reviewed as this has been adopted by other projects using carvel technology.". The fact that the pattern has been adopted doesn't prove it's well implemented or optimal. I could argue many of the things that are needed and not easy. UX needs quite a big enhancement.
- Versioning of a metapackage with multiple profiles becomes quite complicated, as every profile follows same version scheme.

jorgemoralespou:
All, in all, I'm in favor of doing bundles/metapackages or whatever we want to create as can be seen by this issue https://github.com/vmware-tanzu/community-edition/issues/322, but I'm personally not keen of the current implementation followed by TAP as I think can be improved. 

I would consider this proposal the place to describe what we want rather than a proposal of what we can do. TCE should be the place where we can innovate and come up with the right way of doing things later downstream.

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+
+This metapackage is essentially a package of packages wherein it references a Bill of Materials that includes existing packages in the repository. Being able to install, update & remove these packages in concert with each other simplifies the complexity significantly. 
+
+This is a method in practice in other projects leveraging Carvel tooling. 
+
+## Motivation
+
+The motivation for introducing this is to reduce the complexity that comes up with installing multiple carvel packages at a time. This will allow newcomers to the project rapidly establish a working technology stack.
+
+* Describe why the change is important and the benefits to users.
```
jpmcb, line 17: Was this part of the template? I think this line needs to be removed.

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+The proposal will describe what a metapackage is and how it will be implemented. 
+
+The proposal can be leveraged for various user scenarios. One example user scenario is described below. 
+
+**Terminology**
+
+* Metapackage - concept to bring two or more packages together. 
+* Profile - Enabling users to decide by a config value which part of the metapackage the user wants to install. 
+
+
```
joshrosso, line 39: nit: please remove these excess line-breaks

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Example Scenario
+
+The below scenario is one use-case of the metapackage technology. This is **not** the only way to use it. 
+
```
joshrosso, line 44: nit: please remove excess line breaks

docs/designs/2848-metapackage-proposal.md
stmcginnis, line 17: Remove this line?

docs/designs/2848-metapackage-proposal.md
jpmcb, line 39: Nit: too many extra lines here

jpmcb: The markdown linter may have also failed on these MD formatings

docs/designs/2848-metapackage-proposal.md
nrb, line 17: This line seems left over from the template.

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+## Compatibility
+
+* tanzu cli with package plugin
+* kapp controller version
+* TCE-repo package repository
+
+## Alternatives
+
+The alternative today is to install each package individually with it's own values file and it's own upgrade planning.
```
stmcginnis, line 149: Super nit I only mention if you do other more important updates:

s/it's/its/

joshrosso:
> I would consider this proposal the place to describe what we want rather than a proposal of what we can do. TCE should be the place where we can innovate and come up with the right way of doing things later downstream.

@kartiklunkad26 I second this sentiment.

So overall very much +1. However, a metapackage is essentially a package and **we can accept** these today. I'm wondering if the proposal can reflect what you're specifically looking to bring to the table as a metapackage?

Otherwise, if the question is, will TCE accept metapackage contributions? The answer is yes!

We even mention it here:

https://github.com/vmware-tanzu/community-edition/blob/main/docs/packaging/considerations/versioning.md

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+| bar 2 | bar 2 |
+|       | bar 3 |
+
+2. Installing a profile
+* List version information for the package by running:
+
+```
+tanzu package available list foo.tanzu.vmware.com 
+```
+* Create a `foo-values.yaml` file. This configuration file will have configuration required to deploy foo and the subordinate packages.
```
jpmcb, line 78: I don't think this should be a single yaml file: these could end up being huge and I think breaking them apart will make them easier for users to grok.

For example, lets say there was some observability metapackage that was Grafana, Prometheues, and Fluent. That could easily run 1000 lines of yaml if it was all in one.

docs/designs/2848-metapackage-proposal.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Goals
+
+* Demonstrate the ability that multiple subordinate packages can be installed using a single metapackage. 
+* Be able to provide a single values.yaml file that can be used for all the subordinate package values.yaml.
+
+### Non-Goals/Future Work
+
+* Define each profile (and the packages it will have) that will use this metapackage technology. 
+
```
jpmcb, line 27: Can we call out a non-goal here as _not_ being able to coordinate packages when deployed?

For example, today, _i don't believe_ it's possible via Carvel for packages to introspect one another. So if I installed a prometheus, grafana observability profile, they wouldn't be able to introspect eachother intelligently to automatically deploy dashboards into grafana based on what's on the cluster.

nrb:
I'm a little confused by this proposal; it seems more like something for Carvel tooling...which is already supported. What is the benefit of having this in TCE?

docs/designs/2848-metapackage-proposal.md
joshrosso: ref the idea of dependency management issue that could trend towards something like this being possible.
https://github.com/vmware-tanzu/carvel-kapp-controller/issues/371

joshrosso:
As discussed at https://kubernetes.slack.com/archives/C02GY94A8KT/p1643999766789219, we're going to close this PR as we can accept meta packages as defined in this design doc. For us, they are largely just packages.

We'll focus on getting approval for what Tanzu Application Platform (TAP) will be brining in https://github.com/vmware-tanzu/community-edition/issues/2848.

Thanks!

Issue #2916, opened 2022-01-24
Folder ends with '[568vq]' will be ignored by git


lubronzhan:
## Bug Report

Currently, folder end with letters within`[568vq]` will be ignored, https://github.com/vmware-tanzu/community-edition/blob/main/.gitignore#L31-L32

## Expected Behavior
After creating 1.22.5 folder, `git status` should show the folder is added
```
mkdir 1.22.5
```
## Steps to Reproduce the Bug

```
mkdir 1.22.5
git status
```

## Environment Details

* Build version (`tanzu version`): 
* Deployment (Managed/Standalone cluster): 
* Infrastructure Provider (Docker/AWS/Azure/vSphere): 
* Operating System (client):

lubronzhan:
Closed due to https://github.com/vmware-tanzu/community-edition/pull/2917

PR #2917, opened 2022-01-24
Explicity ignore file end with [568vq].

lubronzhan:

It will block team to add packages with version like `*.5`

```release-note

```

Fixes: #https://github.com/vmware-tanzu/community-edition/issues/2916

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,9 +28,6 @@ bin
 addons/repos/generated
 
 # Architecture specific extensions/prefixes
-*.[568vq]
-[568vq].out
-
 *.cgo1.go
 *.cgo2.c
 _cgo_defun.c


jpmcb (changes requested):
I believe the `$` will glob for a literal dollar sign, not the end of the string:

Screen Shot 2022-01-24 at 11 32 51 AM

And this seems to also not be working in the gitignore:
```
$ echo "hello world" > test.q
$ git status
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        test.q

```

Again, I'm wondering if we just remove those two lines. Not sure what they are for

lubronzhan:
Thanks for finding the issue.
I think these are for cgo's output after I did some googling. Not many repos has these lines. Even k8s repo

lubronzhan:
Ok this can work
```
!*.[568vq]/
```
image

stmcginnis:
> I'm wondering if we can just remove both those lines since I'm really not sure what they do for us

I think so, and that would be my preference.

If these are for cgo, I think that's OK. We don't really support cgo (at least officially so far) so I don't think we need to do anything special to accomodate that.

lubronzhan:
Ok I just removed these two lines.

jpmcb (approved):

stmcginnis:
Thanks!

PR #2918, opened 2022-01-24
Introduces `ExistingClusterKubeconfig` option to `unmanaged-cluster`

### LINKED ISSUE #2890: `unmanaged-clusters` create should have a `--kubeconfig` flag
## Bug Report

Currently, you can "bring your own cluster" with unmanaged-clusters using the config:
```
$ tanzu unmanaged-clusters create -c my-config
```

where the config looks like:

```yaml
kubeconfig: /path/to/kubeconfig
```

But there is no option to provide just a file path to the command line

## Expected Behavior

I would expect something like this to work:
```
$ tanzu unmanaged-clusters create --kubeconfig /path/to/file
```

## Steps to Reproduce the Bug

Run `tanzu unmanaged-clusters create help` and see no option to provide a kubeconfig

jpmcb:
This PR introduces the `ExistingClusterKubeconfig` config option which is supported by the `--existing-cluster-kubeconfig` flag during `create`.

This enables users to bring their own clusters that already exist and already have a kubeconfig on the filesystem.

#### Why not just use `KubeconfigPath`?

This would have required a greater refactor since the `config.KubeconfigPath` is used to create and set the context of _new_ created kind clusters:

```go
clusterConfig := kindcluster.CreateWithkubeconfigPath(c.KubeconfigPath)

...

err = kindProvider.Create(c.ClusterName, clusterConfig, kindConfig)
```

This means that we would have been _overloading_ the use of `KubeconfigPath`: it would both be the path to the kubeconfig for some arbitrary created cluster or used in logic to determine if a user is attempting to provide an existing cluster.

Instead, introducing a _new option_ allows us to keep `Kubeconfig` just as it is: the file path to a kubeconfig.

Deeper in the implementation, we actually take advantage of this by setting the `Kubeconfig` to the what is provided by a user when using `ExistingClusterKubeconfig`. This enables existing program flow to stay in place without creating extra logic around a separate kubeconfig.

```release-note
introduce ExistingclusterKubeconfig option for unmanaged-cluster
```

Fixes: #2890

Existing functionality intact, still able to create cluster and target it:

Screen Shot 2022-01-24 at 11 09 30 AM

- create some arbitrary cluster. Used kind for this example:
```
$ kind create cluster --name already-exists
...
```
- Get the cluster's kubeconfig:
```
$ kind get kubeconfig --name already-exists > /tmp/kubeconfig.yaml
```

- Give `unmanaged-cluster` the kubeconfig as an option:

Screen Shot 2022-01-24 at 1 34 56 PM

Check the stats of the cluster:
```
❯ k get pods -A --kubeconfig kubeconfig.yaml
NAMESPACE            NAME                                                   READY   STATUS    RESTARTS   AGE
kube-system          coredns-558bd4d5db-j72jw                               1/1     Running   0          2m52s
kube-system          coredns-558bd4d5db-psrm9                               1/1     Running   0          2m52s
kube-system          etcd-already-exists-control-plane                      1/1     Running   0          2m55s
kube-system          kindnet-gx8v4                                          1/1     Running   0          2m52s
kube-system          kube-apiserver-already-exists-control-plane            1/1     Running   0          2m55s
kube-system          kube-controller-manager-already-exists-control-plane   1/1     Running   0          2m54s
kube-system          kube-proxy-8lgm2                                       1/1     Running   0          2m52s
kube-system          kube-scheduler-already-exists-control-plane            1/1     Running   0          2m55s
local-path-storage   local-path-provisioner-547f784dff-rfhv8                1/1     Running   0          2m52s
tkg-system           kapp-controller-f7dbbcb4f-nrpbz                        1/1     Running   0          89s

```

N/a

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/config/config.go
+++ b/cli/cmd/plugin/unmanaged-cluster/config/config.go
@@ -20,20 +20,21 @@ import (
 )
 
 const (
-	ClusterConfigFile = "ClusterConfigFile"
-	ClusterName       = "ClusterName"
-	Tty               = "Tty"
-	TKRLocation       = "TkrLocation"
-	Provider          = "Provider"
-	Cni               = "Cni"
-	PodCIDR           = "PodCidr"
-	ServiceCIDR       = "ServiceCidr"
-	configDir         = ".config"
-	tanzuConfigDir    = "tanzu"
-	yamlIndent        = 2
-	ProtocolTCP       = "tcp"
-	ProtocolUDP       = "udp"
-	ProtocolSCTP      = "sctp"
+	ClusterConfigFile         = "ClusterConfigFile"
+	ExistingClusterKubeconfig = "ExistingClusterKubeconfig"
+	ClusterName               = "ClusterName"
+	Tty                       = "Tty"
+	TKRLocation               = "TkrLocation"
+	Provider                  = "Provider"
+	Cni                       = "Cni"
+	PodCIDR                   = "PodCidr"
+	ServiceCIDR               = "ServiceCidr"
+	configDir                 = ".config"
+	tanzuConfigDir            = "tanzu"
+	yamlIndent                = 2
+	ProtocolTCP               = "tcp"
+	ProtocolUDP               = "udp"
+	ProtocolSCTP              = "sctp"
 )
 
 var defaultConfigValues = map[string]string{
@@ -60,8 +61,11 @@ type PortMap struct {
 type UnmanagedClusterConfig struct {
 	// ClusterName is the name of the cluster.
 	ClusterName string `yaml:"ClusterName"`
-	// KubeconfigPath is the serialized path to the kubeconfig to use.
+	// KubeconfigPath is the location where the Kubeconfig will be persisted
+	// after the cluster is created.
 	KubeconfigPath string `yaml:"KubeconfigPath"`
+	// ExistingClusterKubeconfig is the serialized path to the kubeconfig to use of an existing cluster.
+	ExistingClusterKubeconfig string `yaml:"ExistingClusterKubeconfig"`
 	// NodeImage is the host OS image to use for Kubernetes nodes.
 	// It is typically resolved, automatically, in the Taznu Kubernetes Release (TKR) BOM,
 	// but also can be overridden in configuration.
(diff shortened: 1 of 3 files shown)


joshrosso (changes requested):
exciting! putting this blocker in so i remember to review/validate today.

cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
```diff
@@ -597,6 +605,21 @@ func runClusterCreate(scConfig *config.UnmanagedClusterConfig) (*cluster.Kuberne
...
+	scConfig.KubeconfigPath = scConfig.ExistingClusterKubeconfig
+	noopManager := cluster.NewNoopClusterManager()
+
+	kc, err := noopManager.Create(scConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Style(outputIndent, color.FgYellow).Warn("Warning: an existing cluster may not be compatible with kapp controller and the default CNI.")
+	log.Style(outputIndent, color.FgYellow).Warn("         unmanaged-cluster will not lifecycle this cluster. Use with caution.\n")
```
joshrosso, line 618: **nit**: I like this warning, but would consider the following:

* remove the arbitrary indentation, or use a different value for `outputIndent`. TBH I don't think the indentation is needed at all with this one.
* remove "use with caution"
* I expect most users of this flag to put `--cni=none` so this warning won't be relevant for that case
* make the warning more about "components installed using this approach will need to be manually removed".

cli/cmd/plugin/unmanaged-cluster/config/config.go
```diff
@@ -60,8 +61,10 @@ type PortMap struct {
 type UnmanagedClusterConfig struct {
 	// ClusterName is the name of the cluster.
 	ClusterName string `yaml:"ClusterName"`
-	// KubeconfigPath is the serialized path to the kubeconfig to use.
+	// KubeconfigPath is the serialized path to the kubeconfig.
 	KubeconfigPath string `yaml:"KubeconfigPath"`
```
joshrosso, line 65: Initially, I had expected this to be removed in favor of `ExistingClusterKubeconfig`.

Any value in keeping it?

jpmcb: > ExistingClusterKubeconfig is almost a sort of "subset" of KubeconfigPath

Hmmm maybe it can be done with just one then. Might be worth chatting over further

joshrosso (changes requested):
Validated this works great against Rancher Desktop.

I also [added an update](https://github.com/vmware-tanzu/community-edition/pull/2941) to support `--cni=none`, which I forsee being common for some of these existing cluster deployments.

Ready to approve once we remove the old `KubeconfigPath` field.

joshrosso (approved):

PR #2929, opened 2022-01-25
Introduce terminal-width aware animated logging for `unmanaged-cluster`

### LINKED ISSUE #2852: unmanaged-cluster: Downloading TKR re-prints log message rather than showing progress `.` characters
## Bug Report

From a comment in another PR (https://github.com/vmware-tanzu/community-edition/pull/2822#issuecomment-1012618133)

image

## Expected Behavior

For a carriage return `\r` to replace this line, giving the visualization of progress.

## Steps to Reproduce the Bug

Remove the locally downloaded TKR then run a cluster create.

jpmcb:
Introduces "terminal-width" aware progress logging.

Using the file descriptor of the attached terminal, we can determine the width of the terminal using Go's standard library. There is a bug ([described in depth here](https://github.com/vmware-tanzu/community-edition/issues/2852#issuecomment-1020427005)) that keeps long lines in small terminals from being replaced.

So, if the progress line is too long, we rebuild a truncated version of it. While this does cut off some information, the next "success" log line is still full length giving the user all the necessary details. 

Shout out to @stmcginnis for finding a solution to this in the go std library!!

```release-note
fixed unmanaged-cluster animated logging wrapping during progress logging incorrectly
```

Fixes: #2852

Normal terminal size still works great:

Screen Shot 2022-01-25 at 9 32 35 AM

Small terminal now won't have the wrapping issue. Lines get replaced as expected:

Screen Shot 2022-01-25 at 9 33 16 AM

_*note*_: even though the full path here is getting truncated, the "success" line is still full length and replaces that line successfully giving all the necessary details to the user. Example:

Screen Shot 2022-01-25 at 9 41 11 AM

N/a

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/log/log.go
+++ b/cli/cmd/plugin/unmanaged-cluster/log/log.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/fatih/color"
+	"golang.org/x/term"
 )
 
 const (
@@ -43,6 +44,8 @@ type CMDLogger struct {
 	logLevel int
 	// instances of indentation (" ") to prepend to a long message
 	indent int
+	// termFd maps to the file descriptor of the attached terminal when the logger is initilized
+	termFd int
 	// logColor defines the color to log the message as define by fatih/color Attributes
 	logColor color.Attribute
 	// output controls where log messages are sent
@@ -106,10 +109,13 @@ type Logger interface {
 
 // NewLogger returns an instance of Logger, implemented via CMDLogger.
 func NewLogger(tty bool, level int) Logger {
+	fd := int(os.Stdout.Fd())
+
 	return &CMDLogger{
 		tty:    tty,
 		level:  level,
 		output: os.Stdout,
+		termFd: fd,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


joshrosso (changes requested):
Building off 6a7c336dcb0b78b14a9ae473c24d517bca8cdf22, I see:

image

jpmcb:
Hmmmm @joshrosso maybe an old binary? I just pushed again to https://github.com/vmware-tanzu/community-edition/commit/13945c2c903be47168456e90c4821649e4aa494b

Here's what I'm seeing on a small terminal:

Screen Shot 2022-01-25 at 10 03 15 AM

Screen Shot 2022-01-25 at 10 03 19 AM

stmcginnis:
Works right for me on iTerm2. I did need to make sure to `rm -fr ~/.config/tanzu` to make sure it had to download the BOM again.

image

Will try on my Linux machine to see if it matches what @joshrosso is seeing.

jpmcb:
Here's on my linux desktop, PopOS, zsh, using Alacrity terminal:

### Normal width

Screenshot from 2022-01-25 10-19-41

### Small width
Screenshot from 2022-01-25 10-20-26

stmcginnis:
Worked as expected for me with default Ubuntu terminal and Sakura terminal.

cli/cmd/plugin/unmanaged-cluster/log/log.go
```diff
@@ -106,10 +109,13 @@ type Logger interface {
 
 // NewLogger returns an instance of Logger, implemented via CMDLogger.
 func NewLogger(tty bool, level int) Logger {
+	fd := int(os.Stdout.Fd())
+
 	return &CMDLogger{
 		tty:    tty,
 		level:  level,
 		output: os.Stdout,
+		termFd: fd,
```
stmcginnis, line 118: Seems like we could get this as needed in the progressf call, but things are usually such a short lived process that I suppose it's more efficient to open it once and pass it around rather than getting it on every progress update.

stmcginnis (approved):

joshrosso (approved):

cli/cmd/plugin/unmanaged-cluster/log/log.go
jpmcb: True - I wanted to prevent a `syscall` every time we log progress (roughly once / second), so instead, storing it as part of the command logger struct made more sense. 

I also thought it might be something other APIs in the logging package want, but that's debatable.

Further, the likelihood of the terminal detaching and having it's file-descriptor change is unlikely.

stmcginnis (approved):

PR #2933, opened 2022-01-25
Docs: Updated curl for installing kubectl latest

garrying:
Signed-off-by: Garry Ing <ging@vmware.com>

Updates step 5 in Linux installation to reflect the above instructions of downloading and installing latest `kubectl`. Command is coming from [Install and Set Up kubectl on Linux](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/).

```release-note

```

Fixes: n/a

- `hugo server`
- Navigate to `/docs/latest/cli-installation/`
- Last step (5) in Procedure, Linux

### DIFF
--- a/docs/site/content/docs/assets/cli-install-linux.md
+++ b/docs/site/content/docs/assets/cli-install-linux.md
@@ -61,7 +61,7 @@
 1. You must download and install the latest version of `kubectl`.
 
     ```sh
-    curl -LO https://dl.k8s.io/release/v1.20.1/bin/linux/amd64/kubectl
+    curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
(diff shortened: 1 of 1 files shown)


vrabbi:
i think this change can be very problematic. kubectl has a support scheme defined in the kubernetes documentation that states:  
"[kubectl is supported within one minor version (older or newer) of kube-apiserver](https://kubernetes.io/releases/version-skew-policy/#kubectl)"  
currently that command would install a 1.23 kubectl which while it does work is not supported against any released TKR in TCE making it a problematic change i believe. the version  needs to be 1.20.x it seems with the current skew of 1.19-1.21 TKRs in TCE.

garrying:
TKR to kubectl version is interesting to know and something I wasn't aware of. Maybe this can be closed with a follow-on to update copy to make clear `1.20.x` for proper support.

dvonthenen:
Closing this based on this comment:
https://github.com/vmware-tanzu/community-edition/pull/2933#issuecomment-1022588606

PR #2960, opened 2022-01-27
package/external-dns: Remove empty serviceaccount value from e2e fixture

christianang:

This removes the empty serviceaccount from the external-dns e2e fixture. This is causing some downstream tests to fail because it picks up upstream tests immediately, but not the package. This won't cause any issues for current e2e test runs because no assertions were made for the service account. We intend to add this back once downstream is more robust.

```release-note
NONE
```

Fixes: n/a

Manually ran the e2e tests locally.


tylerschultz (approved):
LGTM

PR #2963, opened 2022-01-28
package/external-dns: Ensure external-dns e2e test deploys correct version of package

christianang:

Previously the external-dns test would assume there was only one version of the package on the cluster and that would be the correct version to test. This checks the versions of the available packages and picks the correct version to test, ensuring the correct version is tested and allowing multiple versions of the package to be available and tested at the same time.

This also updates ginkgo/gomega libraries and also adds the ginkgo cli as a dependency. This ensures that the go.sum contains the dependencies of the ginkgo cli and allows us to run `go run github.com/onsi/ginkgo/ginkgo` in the Makefile.

```release-note
NONE
```

n/a

Ran the external-dns e2e tests on a local cluster.


tylerschultz (dismissed):
LGTM

tylerschultz (approved):
LGTM

christianang:
@dvonthenen or @joshrosso once this PR is merged I think we can reenable the the external-dns unit test workflow. It adds the ginkgo cli as a dependency, which should stop the ci complaining about the go sum.

christianang:
@vmware-tanzu/tce-owners can I get a review on this? Thanks!

davidvonthenen (approved):

PR #3001, opened 2022-02-02
Notify community with daily developer build downloads

davidvonthenen:

Since we can't publish a message in the #`#tanzu-community-edition` slack channel, we have deleted the failed method to publish an RSS feed and instead replaced it with:

- dynamically modifying the README.md with the links for the latest daily build on all supported platforms
- post a thread in a newly created category in the repo's Discussion tab found here: https://github.com/vmware-tanzu/community-edition/discussions/categories/daily-development-builds

Note that on the next successful build, the placeholder text will be updated with each subsequent build's links.

Future TODO: Something that could be cool.... taking that Discussions thread that gets created and maybe using the release tool to provide a commit history between daily builds. Just saying...

```release-note
Notify community with daily developer build downloads
```

NA

Performed local testing of scripts

NA

### DIFF
--- a/hack/dailybuild/dailybuild.go
+++ b/hack/dailybuild/dailybuild.go
@@ -4,26 +4,35 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"crypto/tls"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"time"
 
 	"cloud.google.com/go/storage"
-	"github.com/gorilla/feeds"
 )
 
 const (
-	projectID  string = "vmware-cna"
-	bucketName string = "tce-cli-plugins-staging"
-	gcpFile    string = "gcptokenfile.json"
-	rssFile    string = "../../rss.xml"
+	regexReadmeLinuxAmd64   string = "\\[Linux AMD64.*\\]\\(.*\\)"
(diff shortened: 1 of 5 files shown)


hack/dailybuild/dailybuild.go
```diff
@@ -165,69 +175,105 @@ func uploadAll(tag, gcpToken string) (*Builds, error) {
...
 }
 
-func createRssFeed(builds *Builds) error {
-	// update rss
-	now, err := time.Parse(time.RFC3339, "2013-01-16T21:52:35-05:00")
+func updateReadme(builds *Builds) error {
+	readme, err := os.ReadFile(fullPathReadme)
 	if err != nil {
-		fmt.Printf("time.Parse failed: %v\n", err)
+		fmt.Printf("ReadFile returned error: %v\n", err)
```
stmcginnis, line 181: Would it be better to do something like `"Failed to parse README file at %s: %v\n", fullPathReadme, err`

stmcginnis (approved):
A small suggestion, but I think it would be fine as a future refinement. This looks good to me.

hack/dailybuild/dailybuild.go
dvonthenen: Will follow up on this for Monday. Going to merge this so we have a build tonight. Will fix this plus I need to delete the rss.xml at the root of the repo.

PR #3018, opened 2022-02-02
Add rebase rule to ako statefulset to prevent kapp-controller from removing its annotations.

HanFa:

AKO uses annotation to pass "cleanup finished" signal to the AKO-operator. We don't want this annotation to be override (removed) by the kapp-controller during its reconcilation.

```release-note
Add rebase rule to ako statefulset to prevent kapp-controller from removing its annotations.
```

Fixes: #

### DIFF
--- a/addons/packages/load-balancer-and-ingress-service/1.6.1/bundle/config/kapp-config.yaml
+++ b/addons/packages/load-balancer-and-ingress-service/1.6.1/bundle/config/kapp-config.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ako-kapp-config
+  labels:
+    kapp.k14s.io/config: ""
+data:
+  config.yml: |
+    apiVersion: kapp.k14s.io/v1alpha1
+    kind: Config
+    rebaseRules:
+    - paths:
+      - [metadata, annotations, AviObjectDeletionStatus]
+      type: copy
+      sources: [existing]
+      resourceMatchers:
+      - kindNamespaceNameMatcher: {kind: StatefulSet, namespace: avi-system, name: ako}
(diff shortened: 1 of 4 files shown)


addons/packages/load-balancer-and-ingress-service/1.6.1/bundle/config/kapp-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: ako-kapp-config
+  labels:
+    kapp.k14s.io/config: ""
+data:
+  config.yml: |
+    apiVersion: kapp.k14s.io/v1alpha1
+    kind: Config
+    rebaseRules:
+    - paths:
+      - [annotations]
```
XudongLiuHarold, line 13: I don't think this path is correct?

Shouldn't it be `[metadata,annotations]`

HanFa: oh wow, great catch

addons/packages/load-balancer-and-ingress-service/1.6.1/bundle/config/kapp-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: ako-kapp-config
+  labels:
+    kapp.k14s.io/config: ""
+data:
+  config.yml: |
+    apiVersion: kapp.k14s.io/v1alpha1
+    kind: Config
+    rebaseRules:
+    - paths:
+      - [metadata, annotations]
```
lubronzhan, line 13: Looks like you can also add annotation key name https://github.com/vmware-tanzu/carvel-kapp/blob/d3ee9a01b5f0d7d5632b6a157ea7d0338730d497/pkg/kapp/config/default.go#L159

HanFa: greeeeat catch! thanks @lubronzhan

lubronzhan (approved):
/lgtm

Issue #2231, opened 2021-10-13
Automated upgrade test for Chocolatey package as part of PR changes


karuppiah7890:
## Feature Request

Same as #2229 , but just for upgrade. Upgrade has been mentioned in #2229 but it cannot be done as part of #2229 as of now since TCE is not published to the Chocolatey Community Repository

The upgrade check would look something like
- Install from the public package repo containing TCE stable release
- Upgrade using the local package repo containing TCE package with the PR changes

I think this upgrade test will only work when there is a version change in the TCE Chocolatey package. In case of other changes (install script changes, uninstall script changes etc), we have to see if we can do this upgrade test. Or we can try to do some sort of reinstall if not upgrade, in such cases. We have to think this through though

## Describe alternatives you've considered

None, except for manual testing

PR #3055, opened 2022-02-07
Add script to update choco package TCE version after every release

### LINKED ISSUE #2228: Automate publishing of Chocolatey package to Chocolatey Community Repository
## Feature Request

Automate publishing of Chocolatey package to Chocolatey Community Repository. As of now, the TCE Chocolatey package is still under review - https://community.chocolatey.org/packages/tanzu-community-edition/ . Once the review is done and the package is published to the package repository, we can start using automation to publish the package for every stable release

This would mean two things -

One is - we need to change the Chocolatey package code for every stable release in an automated fashion. For example
  - Change version here - https://github.com/vmware-tanzu/community-edition/blob/cd2de42ff721d69439f69603d7c04d17553da1cd/hack/choco/tanzu-community-edition.nuspec#L6 
  - Change version here - https://github.com/vmware-tanzu/community-edition/blob/cd2de42ff721d69439f69603d7c04d17553da1cd/hack/choco/tools/chocolateyinstall.ps1#L5
  - Change SHA sum here - https://github.com/vmware-tanzu/community-edition/blob/cd2de42ff721d69439f69603d7c04d17553da1cd/hack/choco/tools/chocolateyinstall.ps1#L12

Once the changes are done and a PR could be created automatically and some automated tests could be run to check the changes and the PR could be merged to `main` automatically once the tests pass

The second is - we need to package and publish the Chocolatey package. This could be automated too by triggering a check in `main` every time Chocolatey package code changes and as part of this check, we can check if a new version (compared to public repository version) is present in the package code, if yes, we can automatically publish it, if not, we can leave out the publish step

Ideally the version change would be pushed by automation through a PR.

This is just one idea, we can think of other ways to do it too 👍 

None, except making changes to the package and publishing the package manually

aman556:
Signed-off-by: Aman Sharma <amansh@vmware.com>

This PR adds a script which will update the version of TCE in choco package after ever release.

Files in which TCE version is updating are:
- [hack/choco/tce-community-edition.nuspec](https://github.com/vmware-tanzu/community-edition/blob/main/hack/choco/tanzu-community-edition.nuspec)
- [hack/choco/tools/chocolateyinstall.ps1](https://github.com/vmware-tanzu/community-edition/blob/main/hack/choco/tools/chocolateyinstall.ps1)

```release-note

```

Fixes: [#2228](https://github.com/vmware-tanzu/community-edition/issues/2228)

Publish a release in forked branch after which GitHub workflow ran and update the TCE version in choco package files.
Action logs for release-choco-package workflow: [logs](https://github.com/aman556/community-edition/runs/5092964616?check_suite_focus=true) 

Please have a look on the above logs.

### DIFF
--- a/.github/workflows/release-choco-package.yaml
+++ b/.github/workflows/release-choco-package.yaml
@@ -1,0 +1,33 @@
+name: Release - Choco Package
+
+on:
+  release:
+    types:
+      - released
+
+jobs:
+  update-choco-package:
+    name: Update Choco Package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Config credentials
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        run: |
+          git config --global pull.rebase true
+          git config --global url."https://git:${GITHUB_TOKEN}@github.com".insteadOf "https://github.com"
+          git config --global user.name github-actions
+          git config --global user.email github-actions@github.com
+
+
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Update Choco Package
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_ACCESS_TOKEN }}
+        shell: bash
+        run: |
+          GITHUB_REF="${{ github.ref }}"
+          version=${GITHUB_REF/refs\/tags\//}
+          ./hack/choco/update-choco-package.sh ${version}
--- a/hack/choco/update-choco-package.sh
+++ b/hack/choco/update-choco-package.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+ 
+# Copyright 2022 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+ 
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+ 
+version="${1:?TCE version argument empty. Example usage: ./hack/choco/update-choco-package.sh v0.10.0}"
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is not set}"
+ 
+ 
+temp_dir=$(mktemp -d)
+ 
+pushd "${temp_dir}"
+
+TCE_REPO="https://github.com/vmware-tanzu/community-edition" 
+TCE_REPO_RELEASES_URL="https://github.com/vmware-tanzu/community-edition/releases"
+TCE_WINDOWS_TAR_BALL_FILE="tce-darwin-amd64-${version}.tar.gz"
+TCE_CHECKSUMS_FILE="tce-checksums.txt"
+ 
+echo "Checking if the necessary files exist for the TCE ${version} release"
+ 
+wget --spider -q \
+   "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_WINDOWS_TAR_BALL_FILE}" || {
+       echo "${TCE_WINDOWS_TAR_BALL_FILE} is not accessible in TCE ${version} release"
+       exit 1
+   }
+ 
+wget --spider -q "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_CHECKSUMS_FILE}" || {
+   echo "${TCE_CHECKSUMS_FILE} is not accessible in TCE ${version} release"
+   exit 1
+}
+ 
+# Use --depth 1 once https://github.com/cli/cli/issues/2979#issuecomment-780490392 get resolve
+git clone "${TCE_REPO}"
+
+cd community-edition
+ 
+PR_BRANCH="update-tce-to-${version}-${RANDOM}"
+ 
+# Random number in branch name in case there's already some branch for the version update,
+# though there shouldn't be one. There could be one if the other branch's PR tests failed and didn't merge
+git checkout -b "${PR_BRANCH}"
+
+
+# Replacing old version with the latest stable released version.
+# Using -i so that it works on Mac and Linux OS, so that it's useful for local development.
+sed -i -e "s/\(\$releaseVersion =\).*/\$releaseVersion = ""'${version}'""/g" hack/choco/tools/chocolateyinstall.ps1 
+rm -fv hack/choco/tools/chocolateyinstall.ps1-e
+
+version="${version:1}"
+sed -i -e "s/\(<version>\).*\(<\/version>\)/<version>""${version}""\<\/version>/g" hack/choco/tanzu-community-edition.nuspec
+rm -fv hack/choco/tanzu-community-edition.nuspec-e
+ 
+git add hack/choco/tools/chocolateyinstall.ps1
+git add hack/choco/tanzu-community-edition.nuspec
+ 
+git commit -m "auto-generated - update tce choco install scripts for version ${version}"
+ 
+git push origin "${PR_BRANCH}"
+ 
+gh pr create --repo ${TCE_REPO} --title "auto-generated - update tce choco install scripts for version ${version}" --body "auto-generated - update tce choco install scripts for version ${version}"
+ 
+gh pr merge --repo ${TCE_REPO} "${PR_BRANCH}" --squash --delete-branch --auto
+ 
+popd


hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,78 @@
...
+
+TCE_REPO="https://github.com/vmware-tanzu/community-edition/" 
+TCE_REPO_RELEASES_URL="https://github.com/vmware-tanzu/community-edition/releases"
+TCE_DARWIN_TAR_BALL_FILE="tce-darwin-amd64-${version}.tar.gz"
+TCE_LINUX_TAR_BALL_FILE="tce-linux-amd64-${version}.tar.gz"
+TCE_CHECKSUMS_FILE="tce-checksums.txt"
+ 
+echo "Checking if the necessary files exist for the TCE ${version} release"
+ 
+curl -f -I -L \
```
stmcginnis, line 27: Do we care about the darwin and linux release artifacts when we're dealing with choco?

.github/workflows/release-choco-package.yaml
```diff
@@ -0,0 +1,33 @@
+name: Release - Choco Package
+
+on:
+  release:
+    types:
+      - released
```
stmcginnis, line 6: This will only run when a full official release is done, not a pre-release, right?

Edit: That does appear to be the case according to https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+ 
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
```
stmcginnis, line 3: Copyright should be 2022 now.

stmcginnis (commented):
I think the release note should be removed from the PR description. Release notes are used by end users to see what a new release will provide. The fact that we added a script to update some packaging is an internal detail that is not relevant to whether someone decides to pick up a new release or not.

hack/choco/update-choco-package.sh
aman556: Thanks @stmcginnis for reviewing.
Sorry missed that completely that dealing with choco not brew.

.github/workflows/release-choco-package.yaml
aman556: Right once the full official release is there we will update the TCE version for choco not for the pre-release as in background choco is installing the latest official release.

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,71 @@
...
+       echo "${TCE_WINDOWS_TAR_BALL_FILE} is not accessible in TCE ${version} release"
+       exit 1
+   }
+ 
+wget "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_CHECKSUMS_FILE}" || {
+   echo "${TCE_CHECKSUMS_FILE} is not accessible in TCE ${version} release"
+   exit 1
+}
+ 
+git clone "${TCE_REPO}"
```
stmcginnis, line 37: Let's do `git clone --depth 1` since we don't need to pull down the full history.

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,71 @@
...
+ 
+PR_BRANCH="update-tce-to-${version}-${RANDOM}"
+ 
+# Random number in branch name in case there's already some branch for the version update,
+# though there shouldn't be one. There could be one if the other branch's PR tests failed and didn't merge
+git checkout -b "${PR_BRANCH}"
+ 
+# Replacing old version with the latest stable released version.
+# Using -i so that it works on Mac and Linux OS, so that it's useful for local development.
+sed -i -e "s/\(\$releaseVersion =\).*/\$releaseVersion = ""'${version}'""/g" hack/choco/tools/chocolateyinstall.ps1
```
stmcginnis, line 52: If we know this is always going to be running on our Linux GitHub Action environment, the `-i` can be dropped and you don't need to do the `rm` in the next line.

Otherwise, you could do detection like we do in https://github.com/vmware-tanzu/community-edition/blob/5b8537e2c64b902553b57956af0ba650b18b2387/hack/packages/create-package.sh#L27

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,71 @@
...
+pushd "${temp_dir}"
+
+TCE_REPO="https://github.com/vmware-tanzu/community-edition/" 
+TCE_REPO_RELEASES_URL="https://github.com/vmware-tanzu/community-edition/releases"
+TCE_WINDOWS_TAR_BALL_FILE="tce-darwin-amd64-${version}.tar.gz"
+TCE_CHECKSUMS_FILE="tce-checksums.txt"
+ 
+echo "Checking if the necessary files exist for the TCE ${version} release"
+ 
+curl -f -I -L \
```
stmcginnis, line 26: For both of these, we don't care about the contents of the files, just that they exist. I think this and line 32 could be updated to be:

```sh
wget --spider -q "${URL}" || {
```

hack/choco/update-choco-package.sh
nrb: Do we need to do a git clone since we're using the `checkout` action?

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,71 @@
...
+# though there shouldn't be one. There could be one if the other branch's PR tests failed and didn't merge
+git checkout -b "${PR_BRANCH}"
+ 
+# Replacing old version with the latest stable released version.
+# Using -i so that it works on Mac and Linux OS, so that it's useful for local development.
+sed -i -e "s/\(\$releaseVersion =\).*/\$releaseVersion = ""'${version}'""/g" hack/choco/tools/chocolateyinstall.ps1
+rm -fv hack/choco/tools/chocolateyinstall.ps1-e
+ 
+version="${version:1}"
+sed -i -e "s/\(<version>\).*\(<\/version>\)/<version>""${version}""\<\/version>/g" hack/choco/tanzu-community-edition.nuspec
```
nrb, line 56: A general question - have you tried running the ps1 script after it's been modified by this script? I'm wondering if the newline differences between unix and windows will arise due to this editing.

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,71 @@
...
+   echo "${TCE_CHECKSUMS_FILE} is not accessible in TCE ${version} release"
+   exit 1
+}
+ 
+git clone "${TCE_REPO}"
+
+cd community-edition
+
+# make sure we are on main branch before checking out
+git checkout main
```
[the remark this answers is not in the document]
aman556: Make sense I will update in next commit.

hack/choco/update-choco-package.sh
aman556: Hey @nrb thanks for reviewing this PR.
I tried this thing but it was giving error that no main branch is there will give an another try and update with logs here.

hack/choco/update-choco-package.sh
aman556: I haven’t tried but I am planning a upgrade test for [#2231](https://github.com/vmware-tanzu/community-edition/issues/2231)after this script in which i will test once this script will update the version in ps1 script.

hack/choco/update-choco-package.sh
aman556: Agree I will update this.

hack/choco/update-choco-package.sh
stmcginnis: If we use the code that we are already in, then we can just have the `checkout` action and remove the creation of a tmp directory to make this change in.

On the other hand, if we ever want to run this locally, then creating a new pristine temporary clone would help make sure there is nothing else dirty in the repo that would get picked up and committed below.

karuppiah7890: Let's go with pristine directory over the existing git repo from the `checkout` action?

nrb: Ok, I'm fine with the fresh clone.

hack/choco/update-choco-package.sh
aman556: Hello @stmcginnis I was trying the above code change and found that for linux too we need `-i` flag to change the same file so as for both OS mac/linux we need this then there is no need for detection. Please correct me if i interpret wrong yours above suggestion.

hack/choco/update-choco-package.sh
aman556: There is a bug that with `--depth 1` gh pr create will  fail.
Issue comment link for complete context - https://github.com/cli/cli/issues/2979#issuecomment-780490392
So for now we have to avoid   `--depth 1`. 
What do you say @stmcginnis ?

stmcginnis: That's unfortunate, but at least for now the repo is still fairly new enough that pulling the entire history down isn't that much extra overhead. Good thing to fix in the future if the referenced issue is resolved.

If you are still making updates, can you add a comment above this line stating something to the effect that we want to use `--depth 1` but need to wait until a fix for that issue is available?

aman556: Sure I will add a comment above this line and reference the issue too.

aman556:
@stmcginnis and @nrb Please let me know if any other corrections or I have to squash the commits.

stmcginnis (approved):
Looks good! Thanks for addressing all of the comments. I'll wait a bit to merge to give Nolan a chance to re-review if he wants, but we should be able to merge this a little later today.

nrb (approved):
@aman556 If you could file an issue for tracking the `git clone --depth 1` problem, that would be awesome. Otherwise the code looks good, so I'll merge.

hack/choco/update-choco-package.sh
```diff
@@ -0,0 +1,69 @@
...
+       echo "${TCE_WINDOWS_TAR_BALL_FILE} is not accessible in TCE ${version} release"
+       exit 1
+   }
+ 
+wget --spider -q "${TCE_REPO_RELEASES_URL}/download/${version}/${TCE_CHECKSUMS_FILE}" || {
+   echo "${TCE_CHECKSUMS_FILE} is not accessible in TCE ${version} release"
+   exit 1
+}
+ 
+# Use --depth 1 once https://github.com/cli/cli/issues/2979#issuecomment-780490392 get resolve
```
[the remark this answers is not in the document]
aman556: Sure @nrb I will  file one thanks.

PR #3065, opened 2022-02-07
Release notes between daily builds

davidvonthenen:

This will provide release note type changes between daily builds when posting to the Discussion boards.

```release-note
Release notes between daily builds
```

NA

Tested the release generation locally and posted a couple of tests to the Discussion board

NA


hack/dailybuild/publish-daily-build.sh
```diff
@@ -39,6 +40,17 @@ set -x
...
+
+echo "Generating release notes..."
+set +x
+GITHUB_TOKEN="${GITHUB_TOKEN}" release-notes \
+  --org vmware-tanzu --repo tce --branch main \
+  --start-sha "${PREVIOUS_DAILY_HASH}" --end-sha "${ACTUAL_COMMIT_SHA}" \
+  --required-author "" --go-template go-template:./daily.template --output daily-notes.txt
+set -x
+
+echo "${ACTUAL_COMMIT_SHA}" | tee ./PREVIOUS_DAILY_HASH
```
nrb, line 53: Is there a risk that the release-notes program at L47 errors, but we still write the hash out and cause a mismatch?

Put another way - should we confirm the release notes were generated and committed correctly before updating the 'previous hash'?

dvonthenen: As long as the hashes are valid, a file will get generated. To test this scenario, I made it such that the start and end SHA were the same. This yields a blank file. We read this empty file and don't append any release notes.

Worst case scenario (which I don't really foresee), if the hash isn't valid, I think we should update the hash in the file to move past this error and see if this gets corrected in the next daily build.

nrb (approved):

PR #3078, opened 2022-02-08
Doc - Add links to Cluster API infrastructure providers

### LINKED ISSUE #2990: Add references to Cloud Infrastructure Providers implementation repositories
## Description

TCE used Cloud Infrastructure Providers, but it does not reference where respective repositories are located. User may have hard time locating those implementations.

## Location

[Support Matrix](https://tanzucommunityedition.io/docs/latest/support-matrix/#infrastructure-providers-target-platforms)

## Additional Context

Modify `Support Matrix` by adding links to repositories of Cloud Infrastructure providers: 

for exampe: [vSphere](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/)

kcoriordan:

Add links to Cluster API Providers in Support Matrix in doc

```none

```

Fixes: #2990 

Run `hugo server `as per instructions in [readme](https://github.com/vmware-tanzu/community-edition/blob/main/docs/README.md)

Is this the best way to link these or do we need to elaborate further somewhere else in doc, we don't mention cluster api anywhere else in doc We could add a [glossary ](http://localhost:1313/docs/latest/glossary/#c) entry with a description and links  and then link to this glossary item from support matrix - probably over thinking this!

### DIFF
--- a/docs/site/content/docs/latest/support-matrix.md
+++ b/docs/site/content/docs/latest/support-matrix.md
@@ -18,9 +18,9 @@ hardware/software configurations is required on your local machine.
 After you install Tanzu Community Edition on your local machine, you can use the Tanzu CLI to deploy a cluster to **one** of the following infrastructure providers:
 | Cloud Infrastructure Provider    |Local Infrastructure Provider|
 |:------------------------ |:------------------------ |
-|Amazon Web Services (AWS) |Docker |
-|Microsoft Azure| |
-|vSphere| |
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Looks great to me.

> We could add a [glossary ](http://localhost:1313/docs/latest/glossary/#c)entry with a description and links and then link to this glossary item from support matrix - probably over thinking this!

That is a good idea! I think we could go there, but this is a good start. So let's merge this, and if we decide later that it adds value to change it to link to a glossary where we can expand on what these means, we can do that as a follow up.

Thanks!

PR #3079, opened 2022-02-08
add Contour 1.20.1 package

### LINKED ISSUE #696: Contour addon: validate values.yaml values
## Feature Request

We should add validation functions to verify that the user has provided valid values.yaml values. These would be implemented in a .star file within the Contour addon dir.

N/A

## Additional context

N/A

cc @vmware-tanzu/contour-owners

### LINKED ISSUE #697: Contour addon: set Envoy service annotations based on infra provider & params
## Feature Request

The Envoy service requires certain annotations based on the infrastructure provider, provider-specific params, and other things like the proxy protocol setting. We should add values.yaml values and logic to intelligently derive and apply the appropriate set of annotations.

N/A

## Additional context

Need to verify the supported set of infrastructure providers - vsphere, aws, azure (?), anything else?

cc @vmware-tanzu/contour-owners

skriss:
Adds Contour 1.20.1 package

Also adds more validation and defaulting of config values, as well as unit tests.

```release-note
Add Contour 1.20.1 package
```

Fixes #696 
Fixes #697 

WIP

Individual commits should help review by separating out copy-paste steps from modifications.


addons/packages/contour/1.20.0/bundle/config/values.yaml
```diff
@@ -0,0 +1,94 @@
...
+  #! Envoy service settings.
+  service:
+    #! The type of Kubernetes service to provision for Envoy. If not specified,
+    #! will default to "NodePort" for docker and vsphere and "LoadBalancer" for others.
+    type: ""
+
+    #! The desired load balancer IP.  If type is not "LoadBalancer", this field is ignored.
+    #! It is up to the cloud provider whether to honor this request.  If null, then load balancer IP
+    #! will be assigned by the cloud provider.
+    loadBalancerIP: null
```
sunjayBhatia, line 37: just curious, any reason this is null vs empty string?

sunjayBhatia: or i guess since `loadBalancerIP` was added first, why empty string vs null haha, seems like the same thing ultimately?

skriss: Yeah i think ultimately either works as a "zero value", but let me double-check and make consistent.

skriss: Double-checked this and they function equivalently; I switched all to use "" or 0 to mean "not specified" for consistency.

addons/packages/contour/1.20.0/bundle/config/contour.star
```diff
@@ -0,0 +1,120 @@
...
+  end
+  
+  data.values.envoy.logLevel in ("trace", "debug", "info", "warning", "warn", "error", "critical", "off") or assert.fail("envoy.logLevel must be one of trace|debug|info|warning/warn|error|critical|off")
+  
+  data.values.envoy.terminationGracePeriodSeconds or assert.fail("envoy.terminationGracePeriodSeconds must be provided")
+end
+
+def validate_envoy_service():
+  if data.values.envoy.service.type:
+    # TODO should we relax this constraint? Not sure why we need to disallow ClusterIP for vsphere.
```
skriss, line 103: I'm leaning towards removing this restriction of no ClusterIP services for vsphere. I'm not aware of why it's there, and I feel like a user will bump into it at some point and we'll have no good reason not to allow it. It still defaults to "NodePort" for vsphere, so only will use ClusterIP if the user explicitly specifies it.

addons/packages/contour/1.20.0/package.yaml
```diff
@@ -0,0 +1,141 @@
...
+  version: 1.20.0
+  releaseNotes: "contour 1.20.0 https://github.com/projectcontour/contour/releases/tag/v1.20.0"
+  releasedAt: 2022-01-31T00:00:00Z
+  licenses:
+    - "Apache 2.0"
+  template:
+    spec:
+      fetch:
+        - imgpkgBundle:
+            image: projects.registry.vmware.com/tce/contour@sha256:TODO
```
skriss, line 16: TODO

skriss:
Testing looking good. May end up holding this until we do a Contour 1.20.1, if that comes soon.

sunjayBhatia (dismissed):
LGTM, in the same boat about the ClusterIP thing, not sure why that is there. 👍🏽 to hold if we end up doing a patch release

addons/packages/contour/1.20.0/bundle/config/contour.star
skriss: Dropped this restriction.

skriss:
> LGTM, in the same boat about the ClusterIP thing, not sure why that is there. 👍🏽 to hold if we end up doing a patch release

Dropped the vsphere/clusterIP restriction, now will just hold off for a bit to wait for the patch release.

skriss:
Got this updated for 1.20.1, and added a bunch of unit tests, should be ready for a final review.

addons/packages/contour/1.20.1/test/unittest/contour_test.go
```diff
@@ -0,0 +1,729 @@
...
+		})
+
+		It("defaults the Envoy service type to LoadBalancer", func() {
+			Expect(err).NotTo(HaveOccurred())
+
+			docs := findDocsContainingLines(output, "kind: Service", "name: envoy")
+			Expect(docs).To(HaveLen(1), "Envoy service not found")
+
+			service := unmarshalService(docs[0])
+			Expect(service.Spec.Type).To(Equal(corev1.ServiceTypeLoadBalancer))
```
skriss, line 200: I opted against checking the entire ~10k lines of YAML for each test case as I don't think that's manageable on an ongoing basis since they'd all have to be updated regularly as things upstream change. I am thinking about adding a small number of test cases that do check the full YAML, though, as a guard against any weird regressions in the overlays. Can be done in a separate PR since this one's big enough already, but interested in others' thoughts on the approach here.

stealthybox:
Nice updates to the Contour package!

I'm going to really appreciate the infrastructure provider option for the AWS NLB.
Looking forward to this package release because it contains the TLSDelegation fix for Ingress v1.

addons/packages/contour/1.20.1/test/unittest/contour_test.go
stealthybox: At first, I didn't like the testing approach because it felt very string dependent instead of YAML native.
After reading through it more and understanding the test-helpers, I think it's pretty precise and minimizes the need to Unmarshal a bunch of YAML which should be nice and fast.

I'm not understanding what you're saying about checking the 10k lines of YAML for each test case.
As long as ytt is returning successfully, that's a pretty good check for the YAML build as a whole with the input values.

I agree with the TODO's to assert the outputs for each infra provider.

Running the tests is a good experience!

Reviewing them all in a single, long file in GitHub is unpleasant, but I'm guessing future diffs will be easier to look through.
I can't think of a way to split up the different sections of tests into separate files without redefining more `Describe` blocks or using more functions and creating unnecessary indirection.

It might be helpful to move the Makefile up a directory and also add the `ytt`/`kbld`/`imgpkg push` commands related to pinning image versions, checking ytt builds, publishing, and other package maintenance needs.

addons/packages/contour/1.20.0/test/unittest/contour_test.go
```diff
@@ -0,0 +1,729 @@
...
+)
+
+var _ = Describe("Contour Ytt Templates", func() {
+	var (
+		values string
+		output string
+		err    error
+
+		configDir = filepath.Join(repo.RootDir(), "addons/packages/contour/1.20.0/bundle/config")
+	)
```
stealthybox: Maybe this could be a per-case `testData` struct instead to avoid bugs in the test runner.

addons/packages/contour/1.20.1/test/unittest/contour_test.go
skriss: > I'm not understanding what you're saying about checking the 10k lines of YAML for each test case.
As long as ytt is returning successfully, that's a pretty good check for the YAML build as a whole with the input values.

I have the tests verifying specific properties of the YAML (e.g. the Envoy service type, the Contour deployment args, etc) rather than having an entire 10k line of "expected YAML" for each test case. That leaves the door open for overlay logic to be inadvertently altering some other portion of the YAML output, without these tests detecting it. If you look at e.g. [external-dns tests](https://github.com/vmware-tanzu/community-edition/tree/main/addons/packages/external-dns/0.10.0/test/unittest/fixtures/expected), they use files containing the entire expected YAML output for each test case. However, their YAML is orders of magnitude shorter than ours.

> Reviewing them all in a single, long file in GitHub is unpleasant, but I'm guessing future diffs will be easier to look through.

Yeah, future reviews should be easier since there should be minimal changes/additions going forward.

> It might be helpful to move the Makefile up a directory and also add the ytt/kbld/imgpkg push commands related to pinning image versions, checking ytt builds, publishing, and other package maintenance needs.
It would make the learning curve easier for users who have a sudden need to fork or contribute fixes.

I'm pretty sure this Makefile needs to be where it is to integrate with repo-wide tooling for running package tests. We could look at adding another one for those other tasks you're talking about, although much of that is already covered in the repo-level Makefile (e.g. `make vendir-sync-package`, `make lock-package-images`, `make push-package`).

addons/packages/contour/1.20.0/test/unittest/contour_test.go
skriss: I think these need to stay as-is because of how Ginkgo works, with references to these closure variables in individual specs as well as BeforeEach, JustBeforeEach, etc. It'll handle parallelization properly if we ever need to go that route (see https://onsi.github.io/ginkgo/#spec-parallelization). You can run `ginkgo -p unittest/` to try it out.

skriss:
cc @sunjayBhatia @youngnick this should be ready for a final review whenever you get a chance. I can pull the unit tests into a separate PR if you prefer but you'll still have to review 'em :)

addons/packages/contour/1.20.1/test/unittest/contour_test.go
sunjayBhatia: yeah maybe we can do something around pulling the CRD part of the expected YAML out into some common helper etc. (or something to that effect) so we don't have to keep repeating them if/when we do add holistic tests, they should be static regardless of config changes and are the longest section line-wise of our base YAML

skriss: Yeah that's a good idea, they probably are also the YAML that tends to change the most upstream on a release-to-release basis (not that other YAMLs can't change, just that they're less frequent so maybe less of a maintenance burden if we do want to fully validate them).

sunjayBhatia:
could be something with my local setup, but had to run `go get github.com/onsi/ginkgo/ginkgo@v1.16.4` for this diff:

```
diff --git a/addons/packages/contour/1.20.1/test/go.sum b/addons/packages/contour/1.20.1/test/go.sum
index c7cfec0e..e93c1c17 100644
--- a/addons/packages/contour/1.20.1/test/go.sum
+++ b/addons/packages/contour/1.20.1/test/go.sum
@@ -29,6 +29,7 @@ github.com/go-openapi/jsonpointer v0.19.3/go.mod h1:Pl9vOtqEWErmShwVjC8pYs9cog34
 github.com/go-openapi/jsonpointer v0.19.5/go.mod h1:Pl9vOtqEWErmShwVjC8pYs9cog34VGT37dQOVbmoatg=
 github.com/go-openapi/jsonreference v0.19.3/go.mod h1:rjx6GuL8TTa9VaixXglHmQmIL98+wF9xc8zWvFonSJ8=
 github.com/go-openapi/swag v0.19.5/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh66Z9tfKk=
+github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 h1:p104kn46Q8WdvHunIJ9dAyjPVtrBPhSr3KT2yUst43I=
 github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0/go.mod h1:fyg7847qk6SyHyPtNmDHnmrv/HOrqktSC+C9fM+CJOE=
 github.com/gogo/protobuf v1.3.2 h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
 github.com/gogo/protobuf v1.3.2/go.mod h1:P1XiOD3dCwIKUDQYPy72D8LYyHL2YPYrpS2s69NZV8Q=
@@ -183,6 +184,7 @@ golang.org/x/tools v0.0.0-20200505023115-26f46d2f7ef8/go.mod h1:EkVYQZoAsY45+roY
 golang.org/x/tools v0.0.0-20200619180055-7c47624df98f/go.mod h1:EkVYQZoAsY45+roYkvgYkIh4xh/qjgUK9TdY2XT94GE=
 golang.org/x/tools v0.0.0-20201224043029-2b0845dc783e/go.mod h1:emZCQorbCU4vsT4fOWvOPXz4eW1wZW4PmDk9uLelYpA=
 golang.org/x/tools v0.0.0-20210106214847-113979e3529a/go.mod h1:emZCQorbCU4vsT4fOWvOPXz4eW1wZW4PmDk9uLelYpA=
+golang.org/x/tools v0.1.5 h1:ouewzE6p+/VEB31YYnTbEJdi8pFqKp4P4n85vwo3DHA=
 golang.org/x/tools v0.1.5/go.mod h1:o0xws9oXOQQZyjljx8fwUC0k7L1pTE6eaCbjGeHmOkk=
 golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
```

sunjayBhatia (dismissed):
Overall test coverage looks quite nice, made some changes locally to overlays and the expected tests failed 👍🏽 

still an open idea/question around the more holistic YAML test(s) but can do that in a separate change IMO

skriss:
Hmm yeah I can repro that, I copied most of that from another package but let me get it fixed. I think we might need a tools.go file.

skriss:
OK got that fixed and updated to ginkgo v2 while I was there, can confirm the go.sum is up to date and tests run via `go test ./...` or `make test` (via ginkgo).

sunjayBhatia (dismissed):
checked the ginkgo/tools changes and looks good 👍🏽

addons/packages/contour/1.20.0/test/unittest/contour_test.go
stealthybox: Oh thanks!
`-p` does make a 2x difference on my little laptop
Should we add it the Makefile target, so it always run it that way?

skriss: Sure, doesn't hurt.

skriss:
Made the last round of updates including pushing the package to Harbor and capturing the SHA, should be ready to merge now.

sunjayBhatia (approved):

jorgemoralespou:
Hey, @skriss IMHO it'll be ideal that the namespace if changed would also be replaced in the provide example contour config in the ConfigMap ([here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L77), [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L80) and [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L84)) and in the sample schema in the CRD ([here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L309), [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L581), [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L1233) and [here](https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/bundle/config/upstream/contour.yaml#L1508)), so that people using these as example have the right namespace that was configured on the cluster.

PR #3105, opened 2022-02-10
Dump daily build changelog for debug purposes

davidvonthenen:

We should dump the changelog in the GH action log for debug purposes. Last night 2/10, the GH Discussion post did not get created and I suspect that it was because the changelog had a special character that was messing up the JSON. We should dump the changelog for debugging and even just for visibility purposes.

Good news is that we anticipated failures would happen and the hash was updated and committed to the repo; therefore won't mess up tonight's (or future) daily build because it can just advance to the next delta.

```release-note
Dump daily build changelog for debug purposes
```

NA

NA

NA


stmcginnis (approved):
Good call!

PR #3107, opened 2022-02-10
Package docs should warn about comments in custom values.yaml file #2909

### LINKED ISSUE #2909: Package Management doc issue at: /docs/latest/package-management
## Description

This page directs you to github where there are {Package}.yaml files used for examples. If you use one of those files, and try to deploy with it, you get an error becausee there are #@ lines in the file and it can't deploy correctly. Users will need to know to remove those lines or replace the file with a usable yaml file.

"expected to not find annotations inside data values file (hint: remove comments starting with '#@'

seemiller:

Minor update to package docs to instruct removing comments from custom values.yaml files.

```release-note
Minor update to package docs to instruct removing comments from custom values.yaml files.
```

Fixes: #2909


joshrosso (changes requested):
We should likely move away from recommending this approach altogether and have folks lookup the schema to form their own values file.

Can we update this to provide that guidance instead? Or is there a technical limitation making looking up the package source still compelling?

seemiller:
@joshrosso I'm good with re-working docs to reference the schema approach. However, only 6 of the current 26 packages are using schemas. So would you consider approving this issue on the basis of it supporting existing packages without schemas and we create a new issue to update docs instructing the use of schemas going forward? Or I'm happy to use this issue if you prefer.

joshrosso:
gotcha, yes

joshrosso (approved):

PR #3109, opened 2022-02-10
Move back to production harbor for compatibility file

joshrosso:
For @dvonthenen to merge when we're ready.


stmcginnis (approved):
Looks good!

PR #3110, opened 2022-02-10
Only build platforms needed in action checks

stmcginnis:

We have several "check" jobs that run on new PRs. These jobs are
currently running `make release` to get the binary to use to perform
those checks.

The GitHub Actions run on a Linux AMD64 platform, but by default the
`make release` target builds for all supported platforms. We end up
waiting a long time while these unnecessary platform targets are built,
to then just pull out the linux-amd64 binary to perform our tests.

This updates the `make release` targets called in these actions to only
build the linux-amd64 binary that we need, saving some time by not
waiting on platform targets we just throw away.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/check-main.yaml
+++ b/.github/workflows/check-main.yaml
@@ -51,7 +51,7 @@ jobs:
         shell: bash
         run: |
           make ensure-deps
-          make release
+          make release ENVS=linux-amd64
       - name: Install
         shell: bash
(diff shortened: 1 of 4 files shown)


stmcginnis:
It looks like this shaves at least ~20 minutes off of total job run time:

image

jpmcb (approved):

PR #3112, opened 2022-02-10
OCI Plugin Repo 

joshrosso:
This sets up OCI repos for our plugins.

WIP.


johnSchnake:
@joshrosso ping me if I can help with this. I keep having packaging conversations w.r.t. Sonobuoy (the conformance plugin) and I'd like to just be involved more with how it is/will be implemented here. I'm considering doing something like this for the Sonobuoy repo itself, so if landing something there would be helpful, let me know.

PR #3123, opened 2022-02-11
Add Pinniped 0.12.1 package

sabbey37:
Add Pinniped 0.12.1 package to bring in bug fix release for an incompatibility between the Pinniped CLI and Chrome v98+ that causes browser-based logins to fail.

```release-note
Add 0.12.1 Pinniped package to bring in bug fix release for an incompatibility between the Pinniped CLI and Chrome v98+ that causes browser-based logins to fail.
```

None

Verified change with vSphere, AWS, and Azure environments.


addons/packages/pinniped/0.12.1/bundle/.imgpkg/images.yml
```diff
@@ -0,0 +1,25 @@
...
+    kbld.carvel.dev/origins: |
+      - resolved:
+          tag: v2.30.2_vmware.1
+          url: projects-stg.registry.vmware.com/tkg/dex:v2.30.2_vmware.1
+  image: projects-stg.registry.vmware.com/tkg/dex@sha256:2b0212eb1a6032c4512c00d8a5d9e98976ad8a0925938027ee01b925e5c4484a
+- annotations:
+    kbld.carvel.dev/id: projects.registry.vmware.com/tce/tanzu_core/addons/tkg-pinniped-post-deploy:v1.3.1
+    kbld.carvel.dev/origins: |
+      - resolved:
+          tag: v0.17.0-dev-31-gf9c9c28e
```
sabbey37, line 22: This is the latest image that's been pushed via Harbor, but not sure if we want to use a dev version.

ankeesler (approved):

addons/packages/pinniped/0.12.1/bundle/.imgpkg/images.yml
seemiller: Looks like a force push removed the images.yml file. The package needs that to be complete. I assume that file will be restored when the image version is sorted out?

sabbey37: @seemiller correct, thanks for noting that!

ankeesler (approved):

PR #3139, opened 2022-02-14
Implements allowing users to set number of nodes for unmanaged-clusters

jpmcb:
## What this PR does / why we need it
Users can now provide `nodes` in configuration,or  the `--nodes` / `-n` flag to set the number of nodes.

The first node is always a control plane. Other nodes are automatically set to works for the `kind` provider.

_Note:_ This does not support scaling.

## Details for the Release Notes (PLEASE PROVIDE)
```release-note
Can set number of nodes for unmanaged-cluster
```

## Which issue(s) this PR fixes
Fixes: #3121

## Describe testing done for PR
```
$ tanzu unmanaged-cluster create -n 3 meow
...

$ kubectl get pods -A
❯ k get pods -A
NAMESPACE            NAME                                         READY   STATUS    RESTARTS   AGE
kube-system          antrea-agent-7dcdm                           2/2     Running   0          43s
kube-system          antrea-agent-gvkqg                           2/2     Running   0          43s
kube-system          antrea-agent-vt86x                           2/2     Running   0          43s
kube-system          antrea-controller-c9cc8588b-4rsr9            1/1     Running   0          42s
kube-system          coredns-558bd4d5db-6nsz7                     1/1     Running   0          2m14s
kube-system          coredns-558bd4d5db-lrx2x                     1/1     Running   0          2m14s
kube-system          etcd-meow-control-plane                      1/1     Running   0          2m18s
kube-system          kube-apiserver-meow-control-plane            1/1     Running   0          2m18s
kube-system          kube-controller-manager-meow-control-plane   1/1     Running   0          2m18s
kube-system          kube-proxy-j6lcx                             1/1     Running   0          2m2s
kube-system          kube-proxy-m8rwg                             1/1     Running   0          2m15s
kube-system          kube-proxy-ntcbp                             1/1     Running   0          2m2s
kube-system          kube-scheduler-meow-control-plane            1/1     Running   0          2m18s
local-path-storage   local-path-provisioner-547f784dff-7hs8w      1/1     Running   0          2m14s
tkg-system           kapp-controller-f7dbbcb4f-jvzq5              1/1     Running   0          104s

$ kubectl get ndoes
NAME                 STATUS   ROLES                  AGE     VERSION
meow-control-plane   Ready    control-plane,master   2m39s   v1.21.5
meow-worker          Ready    <none>                 2m9s    v1.21.5
meow-worker2         Ready    <none>                 2m9s    v1.21.5

```

PR #3150, opened 2022-02-15
Implements setting number of nodes for control-plane and workers

### LINKED ISSUE #3145: Enhance cluster node count capability when creating clusters
With the introduction of:

* https://github.com/vmware-tanzu/community-edition/pull/3139

Users can now specify multi-node clusters. However, the current model operates under an assumption that you're going to specify a "number of nodes". Which may be too generic/limiting for future use cases and especially for future providers.

While a little more complex, I think we need to update the API to be:

* `workerNodeCount:`
* `controlPlaneNodeCount:`

It is up to the cluster provider(s) to understand and operate on these, but we should be passing them down as separate values.

One of the primary complexities here is how to express single node clusters. Single node clusters mean there is no distinction between a worker and controlplane node. On a technical level, it's just a control plane node with its taint removed, which would normally prevent scheduling workloads.

Open to ideas on how to express this. For one, single node should be the **default**. And we can make some contractual conventions such as `controlPlaneNodeCount: 1` and `workerNodeCount: 0` is assumed to be a single node cluster. However, this, again, should be a convention implemented in the provider.

jpmcb:

This commit enables `unmanaged-clusters` to support _both_
setting control-plane nodes _and_ worker nodes, not just an
arbitrary number of nodes (where 1 automatically
is set to a control plane)

```release-note
unmanaged-cluster can set both/either control-plane nodes and worker nodes
```

Fixes: #3145

```shell
# default behavior preserved ...
$ tanzu unmanaged-cluster create test1
...

$ kubectl get nodes
NAME                  STATUS   ROLES                  AGE     VERSION
test1-control-plane   Ready    control-plane,master   48s     v1.21.5

$ tanzu unmanaged-cluster delete test1
...

# can set any number of control plane and worker nodes
$ tanzu unmanaged-cluster create life --control-plane-node-count 2 --worker-node-count 2
...

$ kubectl get pods -A
NAMESPACE            NAME                                          READY   STATUS    RESTARTS   AGE
kube-system          antrea-agent-ft87c                            2/2     Running   0          3m40s
kube-system          antrea-agent-gpsbd                            2/2     Running   0          3m40s
kube-system          antrea-agent-qks7d                            2/2     Running   0          3m40s
kube-system          antrea-agent-szp6x                            2/2     Running   0          3m40s
kube-system          antrea-controller-67846968bb-5b5zq            1/1     Running   0          3m40s
kube-system          coredns-558bd4d5db-cgn4q                      1/1     Running   0          6m32s
kube-system          coredns-558bd4d5db-kwzg4                      1/1     Running   0          6m32s
kube-system          etcd-life-control-plane                       1/1     Running   0          6m31s
kube-system          etcd-life-control-plane2                      1/1     Running   0          6m16s
kube-system          kube-apiserver-life-control-plane             1/1     Running   0          6m31s
kube-system          kube-apiserver-life-control-plane2            1/1     Running   0          6m17s
kube-system          kube-controller-manager-life-control-plane    1/1     Running   1          6m31s
kube-system          kube-controller-manager-life-control-plane2   1/1     Running   0          6m18s
kube-system          kube-proxy-87wmg                              1/1     Running   0          5m25s
kube-system          kube-proxy-c5zbg                              1/1     Running   0          5m25s
kube-system          kube-proxy-dv5s7                              1/1     Running   0          6m33s
kube-system          kube-proxy-vpp4g                              1/1     Running   0          6m18s
kube-system          kube-scheduler-life-control-plane             1/1     Running   1          6m31s
kube-system          kube-scheduler-life-control-plane2            1/1     Running   0          6m17s
local-path-storage   local-path-provisioner-547f784dff-2qvs2       1/1     Running   0          6m32s
tkg-system           kapp-controller-f7dbbcb4f-szzlf               1/1     Running   0          5m4s


$ kubectl get nodes
NAME                  STATUS   ROLES                  AGE     VERSION
life-control-plane    Ready    control-plane,master   6m48s   v1.21.5
life-control-plane2   Ready    control-plane,master   6m22s   v1.21.5
life-worker           Ready    <none>                 5m29s   v1.21.5
life-worker2          Ready    <none>                 5m29s   v1.21.5

```

---

Note: Per Sean's comments, this does _not_ support users deploying multiple control planes with no worker nodes. It will error out. After this merges, I will create an issue to support removing the taint from the control plane nodes if there are no worker nodes present.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
@@ -149,25 +149,52 @@ func kindConfigFromClusterConfig(c *config.UnmanagedClusterConfig) ([]byte, erro
 }
 
 func setNumberOfNodes(c *config.UnmanagedClusterConfig) ([]kindconfig.Node, error) {
-	// Get and check number from config
-	num, err := strconv.Atoi(c.NumberOfNodes)
+	// Get and check control plane count from config
+	cpnc, err := strconv.Atoi(c.ControlPlaneNodeCount)
 	if err != nil {
 		return nil, err
 	}
 
-	if num < 1 {
-		return nil, fmt.Errorf("cannot have less than 1 node")
+	if cpnc < 1 {
+		return nil, fmt.Errorf("cannot have less than 1 control plane node")
+	}
+
+	// Get and check worker count from config
+	wnc, err := strconv.Atoi(c.WorkerNodeCount)
+	if err != nil {
+		return nil, err
+	}
+
+	if wnc < 0 {
+		return nil, fmt.Errorf("cannot have less than 0 worker nodes")
+	}
+
+	// TODO (jpmcb): in single node clusters, control-plane nodes act as worker
+	// nodes, in which case the taint will be removed.
+	//
+	// But, in the case where there are no worker nodes but multiple control plane nodes,
+	// the taint is _not_ automatically removed so workloads will not be scheduled to the control plane nodes.
+	//
+	// In the future, the taint should be removed in order to support workloads being scheduled
+	// on control-plane nodes without worker nodes.
+	// https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#control-plane-node-isolation
+	if cpnc > 1 && wnc == 0 {
+		return nil, fmt.Errorf("multiple control plane nodes require at least one worker node for workload placement")
 	}
 
 	nodes := []kindconfig.Node{}
 
-	for i := 1; i <= num; i++ {
-		n := kindconfig.Node{}
+	for i := 1; i <= cpnc; i++ {
+		n := kindconfig.Node{
+			Role: kindconfig.ControlPlaneRole,
+		}
 
-		// If there are more than 1 nodes, then set those as "worker" nodes.
-		// The first node by default will be set to the control-plane
-		if i != 1 {
-			n.Role = "worker"
+		nodes = append(nodes, n)
+	}
+
+	for i := 1; i <= wnc; i++ {
+		n := kindconfig.Node{
+			Role: kindconfig.WorkerRole,
 		}
 
 		nodes = append(nodes, n)
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
@@ -22,7 +22,8 @@ type createUnmanagedOpts struct {
 	podcidr                   string
 	servicecidr               string
 	portMapping               []string
-	numNodes                  string
+	numContPlanes             string
+	numWorkers                string
 }
 
 const createDesc = `
@@ -63,7 +64,8 @@ func init() {
 	CreateCmd.Flags().StringSliceVarP(&co.portMapping, "port-map", "p", []string{}, "Ports to map between container node and the host (format: '80:80/tcp' or just '80')")
 	CreateCmd.Flags().Bool("tty-disable", false, "Disable log stylization and emojis")
 	CreateCmd.Flags().BoolVar(&co.skipPreflightChecks, "skip-preflight", false, "Skip the preflight checks; default is false")
-	CreateCmd.Flags().StringVarP(&co.numNodes, "nodes", "n", "", "The number of nodes to deploy; default is 1 control-plane")
+	CreateCmd.Flags().StringVar(&co.numContPlanes, "control-plane-node-count", "", "The number of control plane nodes to deploy; default is 1")
+	CreateCmd.Flags().StringVar(&co.numWorkers, "worker-node-count", "", "The number of worker nodes to deploy; default is 0")
 }
 
 func create(cmd *cobra.Command, args []string) error {
@@ -87,7 +89,8 @@ func create(cmd *cobra.Command, args []string) error {
 		config.Cni:                       co.cni,
 		config.PodCIDR:                   co.podcidr,
 		config.ServiceCIDR:               co.servicecidr,
-		config.NumberOfNodes:             co.numNodes,
+		config.ControlPlaneNodeCount:     co.numContPlanes,
+		config.WorkerNodeCount:           co.numWorkers,
 	}
 	clusterConfig, err := config.InitializeConfiguration(configArgs)
 	if err != nil {
--- a/cli/cmd/plugin/unmanaged-cluster/config/config_test.go
+++ b/cli/cmd/plugin/unmanaged-cluster/config/config_test.go
@@ -13,15 +13,16 @@ import (
 )
 
 var emptyConfig = map[string]string{
-	ClusterConfigFile: "",
-	ClusterName:       "",
-	Tty:               "",
-	TKRLocation:       "",
-	Provider:          "",
-	Cni:               "",
-	PodCIDR:           "",
-	ServiceCIDR:       "",
-	NumberOfNodes:     "",
+	ClusterConfigFile:     "",
+	ClusterName:           "",
+	Tty:                   "",
+	TKRLocation:           "",
+	Provider:              "",
+	Cni:                   "",
+	PodCIDR:               "",
+	ServiceCIDR:           "",
+	ControlPlaneNodeCount: "",
+	WorkerNodeCount:       "",
 }
 
 func TestInitializeConfigurationNoName(t *testing.T) {
@@ -62,8 +63,12 @@ func TestInitializeConfigurationDefaults(t *testing.T) {
 		t.Errorf("expected default ServiceCidr, was: %q", config.ServiceCidr)
 	}
 
-	if config.NumberOfNodes != defaultConfigValues[NumberOfNodes] {
-		t.Errorf("expected default NumberOfNodes, was: %q", config.NumberOfNodes)
+	if config.ControlPlaneNodeCount != defaultConfigValues[ControlPlaneNodeCount] {
+		t.Errorf("expected default ControlPlaneNodeCount, was: %q", config.ControlPlaneNodeCount)
+	}
+
+	if config.WorkerNodeCount != defaultConfigValues[WorkerNodeCount] {
+		t.Errorf("expected default WorkerNodeCount, was: %q", config.ControlPlaneNodeCount)
 	}
 }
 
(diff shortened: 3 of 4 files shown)


cli/cmd/plugin/unmanaged-cluster/cmd/create.go
```diff
@@ -63,7 +64,8 @@ func init() {
 	CreateCmd.Flags().StringSliceVarP(&co.portMapping, "port-map", "p", []string{}, "Ports to map between container node and the host (format: '80:80/tcp' or just '80')")
 	CreateCmd.Flags().Bool("tty-disable", false, "Disable log stylization and emojis")
 	CreateCmd.Flags().BoolVar(&co.skipPreflightChecks, "skip-preflight", false, "Skip the preflight checks; default is false")
-	CreateCmd.Flags().StringVarP(&co.numNodes, "nodes", "n", "", "The number of nodes to deploy; default is 1 control-plane")
+	CreateCmd.Flags().StringVarP(&co.numContPlanes, "control-plane-node-count", "", "", "The number of control plane nodes to deploy; default is 1")
```
stmcginnis, line 67: Should these be `StringVar` instead of `StringVarP`? Or still determining what to use for the one-letter shorthand?

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -149,25 +149,39 @@ func kindConfigFromClusterConfig(c *config.UnmanagedClusterConfig) ([]byte, erro
...
+	// Get and check control plane count from config
+	cpnc, err := strconv.Atoi(c.ControlPlaneNodeCount)
 	if err != nil {
 		return nil, err
 	}
 
-	if num < 1 {
-		return nil, fmt.Errorf("cannot have less than 1 node")
+	if cpnc < 1 {
+		return nil, fmt.Errorf("cannot have less than 1 control plane nodes")
```
stmcginnis, line 159: grammar nit: s/nodes/node/

cli/cmd/plugin/unmanaged-cluster/config/config_test.go
```diff
@@ -203,8 +218,12 @@ func TestInitializeConfigurationFromConfigFile(t *testing.T) {
 		t.Errorf("expected ServiceCidr to be set to '9.9.9.0/24', was: %q", config.ServiceCidr)
 	}
 
-	if config.NumberOfNodes != "99" {
-		t.Errorf("expected NumberOfNodes to be set to 'bongos', was: %q", config.NumberOfNodes)
+	if config.ControlPlaneNodeCount != "99" {
+		t.Errorf("expected ControlPlaneNodeCount to be set to 'bongos', was: %q", config.ControlPlaneNodeCount)
```
stmcginnis, line 222: Error string on both of these looks like it was copied from the CNI assertion?

stmcginnis (commented):
I'm getting a hang if I do `unmanaged-cluster create --control-plane-nodes 2 foo` when it gets to installing kapp-controller. Not sure how that would be related to this though.

We don't have full docs yet for `unamanged-cluster` but one thing we may want to note is that if there are more than 1 control plane nodes requested, there will also be a haproxy load balancer container deployed.

Other than the kapp-controller thing, I've built locally and confirmed multiple nodes are created for worker and control plane node requests.

joshrosso (changes requested):
just adding myself as a reviewer.

Thanks for knocking this out!

stmcginnis:
OK, a little more digging on the kapp-controller hang. I see this:

```
Events:
  Type     Reason            Age    From               Message
  ----     ------            ----   ----               -------
  Warning  FailedScheduling  9m49s  default-scheduler  0/2 nodes are available: 2 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
  Warning  FailedScheduling  9m48s  default-scheduler  0/2 nodes are available: 2 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.
```

Looks like we may need to add some more logic around things where if there are no worker nodes requested, but more than one control plane node, we tell it not to taint the nodes.

From single node:

```
$ k get node --show-labels
NAME                STATUS   ROLES                  AGE     VERSION
moo-control-plane   Ready    control-plane,master   2m51s   v1.21.5
~
$ k describe node moo-control-plane | grep Taint
Taints:             <none>
```

From `--control-plan-node-count 2` but `--worker-node-count 0`:

```
$ kubectl --kubeconfig /home/smcginnis/.config/tanzu/tkg/unmanaged/moo/kube.conf get nodes
NAME                 STATUS     ROLES                  AGE    VERSION
moo-control-plane    NotReady   control-plane,master   108s   v1.21.5
moo-control-plane2   NotReady   control-plane,master   83s    v1.21.5
~
$ kubectl --kubeconfig /home/smcginnis/.config/tanzu/tkg/unmanaged/moo/kube.conf describe node moo-control-plane | grep Taint
Taints:             node-role.kubernetes.io/master:NoSchedule
```

stmcginnis:
I can't find a way to change those taints via kind config. The note is pretty clear on behavior here:

https://github.com/kubernetes-sigs/kind/blob/e3864c0be8cc3b17380d7d98d7f2632212ef79b6/pkg/apis/config/v1alpha4/types.go#L154-L157

Perhaps we should start with only allowing `--worker-node-count` and leave `--control-plane-node-count` hard coded to 1 for now?

cli/cmd/plugin/unmanaged-cluster/cmd/create.go
jpmcb: Good catch - these probably don't need shorthands and can just be `StringVar`

cli/cmd/plugin/unmanaged-cluster/config/config_test.go
jpmcb: Guilty as charged 😂

jpmcb:
> Perhaps we should start with only allowing --worker-node-count and leave --control-plane-node-count hard coded to 1 for now?

That was the behavior from https://github.com/vmware-tanzu/community-edition/pull/3139 where `--nodes` would denote the number of _total_ nodes deployed, including 1 control plane by default and the rest workers.

Maybe, the `kind` provider needs some extra logic where if
```
worker-nodes == 0 && control-plane nodes > 1:
  error "cannot deploy multiple control plane nodes with no workers. See [doc link] for more info"
```
?

I think it's still useful to expose the interface of setting control-plane nodes _and_ worker nodes, but the `kind` provider needs to handle the case where it can't deploy multiple CPs without a worker.

stmcginnis:
Maybe we could get around this for now in our code with (pseudo):

```
if controlplanes > 1 and workernodes <1:
    error("multiple control plane nodes require at least one worker node for workload placement")
```

Otherwise, we do have the kubeconfig and we could get the list of nodes and iterate through the control plane nodes and untaint them. That's fairly simple, but could have a little more involved implementation.

jpmcb:
+1 @stmcginnis  - very similar pseudo-code! 😂 

But yes, I agree, we could easily untaint them. Maybe something to tackle outside of this PR? If so, why don't we go with the error case you came up with and  can create a followup issue.

stmcginnis:
> very similar

Oops, at first quick read I thought you were suggesting that was something `kind` itself should handle. So yeah, in total agreement - I think it makes sense for our `kind` provider to add that check for now. :D

jpmcb:
Updated! 🚀

cli/cmd/plugin/unmanaged-cluster/cmd/create.go
```diff
@@ -63,7 +64,8 @@ func init() {
 	CreateCmd.Flags().StringSliceVarP(&co.portMapping, "port-map", "p", []string{}, "Ports to map between container node and the host (format: '80:80/tcp' or just '80')")
 	CreateCmd.Flags().Bool("tty-disable", false, "Disable log stylization and emojis")
 	CreateCmd.Flags().BoolVar(&co.skipPreflightChecks, "skip-preflight", false, "Skip the preflight checks; default is false")
-	CreateCmd.Flags().StringVarP(&co.numNodes, "nodes", "n", "", "The number of nodes to deploy; default is 1 control-plane")
+	CreateCmd.Flags().StringVar(&co.numContPlanes, "control-plane-node-count", "", "The number of control plane nodes to deploy; default is 1. NOTE: Multiple control-plane nodes without a worker node is not supported")
```
joshrosso, line 67: nit: I think we should remove `NOTE: Multiple control-plane nodes without a worker node is not supported`, namely because this is a decision for the provider implementation take make.

stmcginnis: Maybe add "when using the 'kind' provider? It seems like we should have some mention of it if we are going to reject the request later on.

Or we could just set worker node count to 1 if not specified when control plane node count is more than 1.

jpmcb: Very true - great catch Josh!

jpmcb: Whoops - didn't refresh to see your comment Sean. Went ahead and removed it.

Setting the control-plane node to 1 regardless of the users input feels weird to me. But we should have something that communicates what's happening.

Here's an alternative approach I've considered as well:

```shell
# Create a cluster with 2 control plane nodes
# Installation will stale (not error out) and print warning message that user needs to remove taints manually
$ 

...

🚀 Creating cluster test
   Base image downloaded
   Cluster created
   To troubleshoot, use:

   kubectl ${COMMAND} --kubeconfig /home/jmcb/.config/tanzu/tkg/unmanaged/test/kube.conf

⚠️ Warning! Kind based clusters with multiple control planes 
   and no worker nodes will NOT have workloads scheduled to them
   You must remove taints manually in order for installation to continue:

   kubectl --kubeconfig ~/.config/tanzu/tkg/unmanaged/test/kube.conf taint nodes --all node-role.kubernetes.io/master-
   
📧 Installing kapp-controller
   kapp-controller status: Pending...

----

# In a separate terminal, a user can remove the taints using the following command
$ ❯ kubectl --kubeconfig ~/.config/tanzu/tkg/unmanaged/test/kube.conf taint nodes --all node role.kubernetes.io/master-
node/test-control-plane untainted
node/test-control-plane2 untainted

-----

# Installation resumes now that kapp controller can be scheduled onto the control planes
📧 Installing kapp-controller
   kapp-controller status: Running

📧 Installing package repositories
   Core package repo status: Reconcile succeeded

...
```

This workflow worked locally after removing the block that errors if there are no worker nodes to schedule to.

joshrosso: I'm happy to revisit approaches outside this PR.

However, for now, I believe surfacing whether control plane / worker values are valid should be done in the cluster package, via providers. It's something providers can also surface in their pre-flight checks. I'm not too crazy about having conditionals in our flag help text.

joshrosso (approved):
works great :+1:

joshrosso:
Let's wait for @stmcginnis to approve/request comments before merging.

2 is better than 1 :wink: .

stmcginnis (approved):
Good!

jorgemoralespou:
@jpmcb has this been tested with multiple nodes exposing ports?

e.g. `tanzu unmanaged-cluster create life --control-plane-node-count 2 --worker-node-count 2 -p 80:80 -p 443:443`

jpmcb:
@jorgemoralespou - 👀 good catch: this needed a new issue., created here: https://github.com/vmware-tanzu/community-edition/issues/3293

PR #3160, opened 2022-02-16
TCE UI POC - Add a React seed project

miclettej:
Adds a React v17 seed project which includes:
Some basic routing
Import and consumption of Clarity CDS Core and CDS/React
Placeholder Header component
 - Setting pattern for global SCSS variables

Fixes: # 3068

React App builds and runs successfully, able to display Clarity Core components and CSS

Screen Shot 2022-02-16 at 10 53 11 AM


stmcginnis:
Looks like a great start @miclettej!

I'm wondering if we should move this to be under the `ui` plugin so all material is contained there. Though it is nice having it very visible from the root. I guess part of it depends on how we ultimately plan on packing it all up to be included in the resulting binary we deliver.

miclettej:
> 

Ah yes, let me synch with the remote branch and have a look at the UI plugin. Ultimately we will want the UI's compiled 'dist' folder in the plugin, but may want to keep the source code elsewhere. Let's all discuss in our synch up this afternoon.

If you're okay with this location for now, we can move it easily in the near future.

PR #3161, opened 2022-02-16
Standardize adopters logos

stmcginnis:
The current logos included were quite large, so with the rendered view
they took up a lot of the screen. This updates our two existing logos to
be the same, smaller, width so they are a more reasonable size in the
rendered view.

This also adds a white background to the ITG logo so it is easier to
view on systems with dark mode enabled.


jpmcb (approved):
Went over to your branch and viewed your `ADOPTERS.md` and LGTM in the rendered view!! 👍🏼

PR #3162, opened 2022-02-16
Fix the upload of GCP artifacts to bucket

davidvonthenen:

After refactoring of TCE to use v0.16.0 a number of things relating to the upload of TF and TCE artifacts needed to be updated and fixed.

Probably the most significant change in this PR is relocation of the build of the distribution and discovery from `compile` to `install-plugins` step. The reason why is we don't need to create the distribution/discovery until we plan on installing the components.

```release-note
Fix the upload of GCP artifacts to bucket
```

NA

Made sure the new workflow `make release-buckets` works. Along with:
`make release`
`make install-all-tanzu-cli-plugins`

NA

### DIFF
--- a/.github/workflows/build-staging.yaml
+++ b/.github/workflows/build-staging.yaml
@@ -54,7 +54,9 @@ jobs:
           make get-deps
       - name: Build
         shell: bash
-        run: make build-tce-cli-plugins
+        run: |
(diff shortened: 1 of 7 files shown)


stmcginnis (approved):
Looks fine to me.

PR #3184, opened 2022-02-17
Add start of basic server

stmcginnis:
This adds basic HTTP serving for static files and the start of an API
for the UI to interact with the system.

This is exploratory for now. May want to see if we want to reuse/refactor
the existing code in https://github.com/vmware-tanzu/tanzu-framework/tree/main/pkg/v1/tkg/web/server
for our purposes.

### DIFF
--- a/cli/cmd/plugin/ui/api/base.go
+++ b/cli/cmd/plugin/ui/api/base.go
@@ -1,0 +1,19 @@
+// Copyright 2022 VMware Tanzu Community Edition contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// SupportedAPIVersions handles requests to get the supported API versions by the server.
+func SupportedAPIVersions(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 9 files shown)


jpmcb (commented):
Very cool! Seems to work!

Are there extra steps to get it to serve the test page?

I built the binary with `make build-tce-cli-plugins` and ran the binary from the artifacts dir:
```
❯ ./tanzu-ui-linux_amd64
Serving from /home/jmcb/workspace/github.com/vmware-tanzu/tce/artifacts/ui/v0.10.0-dev.5/web
http://0.0.0.0:8080/ui/ browser:

```

but I'm getting 404 on the browser and it didn't automagically open my browser

garrying:
Works for me as well! Seems like the built binary is trying to serve static files from an adjacent `web` folder that is not getting created in the artifacts dir: https://github.com/vmware-tanzu/community-edition/blob/1065d9b5098db562117ad8b6c21686ea06fb5bcb/cli/cmd/plugin/ui/main.go#L44

miclettej (approved):
This is great to get us up and running without having to bring over all of the integration of swagger generated routes and such from the current kickstart UI server. I think it makes sense to move over specific pieces of that server implementation as needed instead of doing it all at once.

PR #3195, opened 2022-02-19
Update cli-install-linux.md

minuq:
Matching the cli-install-linux.md with cli-install-mac.md, changing into the \<DOWNLOAD-DIR\> before using tar to extract files.

```release-note
NONE
```

Fixes: #


minuq:
Not that important, but also not stale i guess

stmcginnis:
Sorry for lack of feedback on this, but I think the general consensus is the one line command is preferred, so we don't want to separate this out into two steps that require changing directories. Thank you for working on this, but I'm going to close it without merging.

PR #3211, opened 2022-02-22
[E2E] Use self hosted AWS EC2 runners for Azure and AWS E2E tests

### LINKED ISSUE #3203: Use self hosted runners for AWS and Azure E2E tests
We currently use GitHub hosted runners which build Tanzu Framework and TCE slowly due to less CPUs and RAM. Some of the workflows use self hosted runners (on demand AWS EC2 instances) which seem to have faster build time due to more CPUs and RAM. So it's better to use those for other E2E tests too which do a build

karuppiah7890:

Use self hosted AWS EC2 runners for Azure and AWS E2E tests

```release-note
NONE
```

Fixes: #3203 

Skipping as I'm yet to get the AMI image (or way to produce it) we need to run the self hosted runners. I would need the AMI image in my AWS account and then have to run the AWS service (available in TCE repo) with my AWS account credentials in it's config and with webhooks settings setup in my fork repo

-

### DIFF
--- a/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
+++ b/.github/workflows/e2e-aws-management-and-workload-cluster.yaml
@@ -14,11 +14,27 @@ on:
       - "**"
 
 jobs:
+  setup-runner:
+    name: Start self-hosted EC2 runner
+    # Only run this job if we're in the main repo, not a fork.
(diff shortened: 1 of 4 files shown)


davidvonthenen (approved):
this looks good to me! this should be good to go!

Issue #3066, opened 2022-02-07
UI Plugin: Add a Go HTTP server to serve static UI content


miclettej:
The Tanzu management cluster plugin has a Go HTTP server that is used for serving static UI and routing HTTP REST calls. A version of this implementation should be added to the TCE UI plugin, to enable development and proving decisions on the UI plugin's technology stack.

The user should be able to reference the UI plugin and execute a command which starts the HTTP server and launches a browser window.

E.G. `tanzu ui init` or `tanzu ui`

PR #3219, opened 2022-02-22
TCE UI - Use Go embed to include and serve UI static files in plugin

miclettej:
Uses Go embed to include and serve contents of the /web/tanzu-ui/build folder, containing all static UI assets
Removes any previous Go bindata implementation

Note: router implementation requires that static Javascript/CSS files included by index.html have a path prefix of 'ui.' I was able to work around this by setting PUBLIC_URL=/ui in the react app but we should find a better implementation

Fixes: #3066 

To test running the UI, you must first build the UI with the makefile target `ui-build` found in `community-edition/cli/cmd/plugin/ui/Makefile`

Once you have built the UI you can run `go run main.go` from `community-edition/cli/cmd/plugin/ui`

### DIFF
--- a/cli/cmd/plugin/ui/Makefile
+++ b/cli/cmd/plugin/ui/Makefile
@@ -8,7 +8,6 @@ GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 
 # Directories
 UI_DIR := web/tanzu-ui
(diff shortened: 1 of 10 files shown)


stmcginnis (approved):
Looks good enough to me! ;)

PR #3255, opened 2022-02-24
Deploy aviinfra setting only in workload cluster to fix ako upgrade issue

XudongLiuHarold:

Because we leverage AVIInfraSetting CRD in ako operator to do the VIP network separation, we need to avoid that crd deployed in ako package when in TKG management cluster.

```release-note

```

Fixes: #


HanFa (approved):
lgtm

PR #3258, opened 2022-02-24
Add cert-injection-webhook package

### LINKED ISSUE #2689: Proposal to add Cert Injection Webhook as a Tanzu package in the TCE Package Repo
## Feature Request

The [cert injection webhook](https://github.com/vmware-tanzu/cert-injection-webhook) is a controller and webhook configuration that can inject trusted ca certificates into pods that need to communicate with services that have self signed certificates. It can be used with other deployments like kpack to allow builds to push images to self signed registries. The addition of the cert injection webhook package would allow kpack to leverage it, but its consumption wouldn't  have to just be limited to kpack. There is another proposal (#2462 ) to add kpack to TCE.

One alternative would be to include the cert injection webhook as part of the kpack package. We decided against this because there may be cases where you don't have kpack installed, but want to leverage the cert injection webhook, such as a cluster that is running application workloads that communicate with a self signed service. Alternatively, If a user doesn't have a need for self signed cert support, they do not have to deploy an additional controller.

## Additional context

kpack proposal - #2462

tomkennedy513:
This PR adds the [Cert Injection Webhook](https://github.com/vmware-tanzu/cert-injection-webhook) as an installable package

```release-note
Add Cert Injection Webhook v0.1.0 as an installable package
```

Fixes: #2689 

Ran smoke tests against deployed package

### DIFF
--- a/addons/packages/cert-injection-webhook/0.1.0/README.md
+++ b/addons/packages/cert-injection-webhook/0.1.0/README.md
@@ -1,0 +1,74 @@
+# Cert Injection Webhook
+
+The Cert Injection Webhook for Kubernetes extends kubernetes with a webhook that injects
+CA certificates and proxy environment variables into pods. The webhook uses certificates and
+environment variables defined in configmaps and injects them into pods with the desired labels or annotations.
+
+## Components
+
+* cert-injection-webhook
+
+## Supported Providers
+
+The following table shows the providers this package can work with.
+
+| AWS | Azure | vSphere | Docker |
+|-----|-------|---------|--------|
+| ✅   | ✅     | ✅       | ✅      |
+
+## Configuration
+
+The following configuration values can be set to customize the cert-injection-webhook
+installation.
+
+| Value          | Required/Optional                        | Description                                                                                                   |
+|----------------|------------------------------------------|---------------------------------------------------------------------------------------------------------------|
+| `ca_cert_data` | Optional                                 | CA cert data to inject into pod trust store                                                                   |
+| `labels`       | Required if annotations are not provided | Array of labels that will be used to match on pods that will have certs and proxy environment injected        |
+| `annotations`  | Required if labels are not provided      | Array of annotations that will be used to match on pods that will have certs and proxy environment injected   |
+| `http_proxy`   | Optional                                 | The HTTP proxy to inject into pod environment                                                                 |
+| `https_proxy`  | Optional                                 | The HTTPS proxy to inject into pod environment                                                                |
+| `no_proxy`     | Optional                                 | A comma-separated list of hostnames, IP addresses, or IP ranges in CIDR format to inject into pod environment |
+
+## Installation
+
+### Package Installation steps
+
+1. Create a `cert-injection-webhook-config-values.yaml` with the labels or annotations (or both) that you would like to use.
+   Any pod that matches one of these labels or annotations will have the provided cert injected. For example:
+
+   ```yaml
+   ---
+   labels:
+   - kpack.io/build
+   annotations:
+   - some-annotation
+   ca_cert_data: |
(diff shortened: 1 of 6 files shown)


jorgemoralespou:
Does this package inject a certificate or a CA? If the latter, maybe name should be revisited. If the former, is there a way then to integrate with cert-manager so that if the cluster has cert-manager configured with a CAIssuer this is used?

jorgemoralespou:
I would also recommend to update the README and package metadata longDescription to give a bit more context on why this is needed, as currently it seems it's only for kpack. Is that the case? Why does kpack require the CAcert to be injected into the pod? Can't it use the CAcert chain that is on the node?

jorgemoralespou:
I've been looking at this package, and it would be nice to integrate it with cert-manager.

This cert-injection-webhook expects that the user provides the ca_cert_data. For platforms where [cert-manager](https://cert-manager.io/) is installed, a user could potentially have a SelfSigned cluster issuer (or really any Issuer) and would want to reuse the CA in those, instead of manually copying from place to place. For that, cert-manager provides a [CA injector](https://cert-manager.io/docs/concepts/ca-injector/) that injects the CA in webhooks, this way, a user could deploy cert-injection-webhook with a reference to the cert-manager's Issuer, which is something that can be easier for end users.

This would give a config like:
```
useCertManager: true
issuer:
  namespace: example
  name: selfsignedissuer
ca_cert_data: <NOTHING>
...
```

tomkennedy513:
> Why does kpack require the CAcert to be injected into the pod? Can't it use the CAcert chain that is on the node? 

kpack would require the CA cert to talk to registry that is using a self signed certificate which we see happen a lot. We could theoretically mount the certs from the node, but it's unclear to me exactly what that would look like. Additionally, I'm not sure we would expect the registry cert to exist on the node for the kpack use case.

tomkennedy513:
> I've been looking at this package, and it would be nice to integrate it with cert-manager.
> 
> This cert-injection-webhook expects that the user provides the ca_cert_data. For platforms where [cert-manager](https://cert-manager.io/) is installed, a user could potentially have a SelfSigned cluster issuer (or really any Issuer) and would want to reuse the CA in those, instead of manually copying from place to place. For that, cert-manager provides a [CA injector](https://cert-manager.io/docs/concepts/ca-injector/) that injects the CA in webhooks, this way, a user could deploy cert-injection-webhook with a reference to the cert-manager's Issuer, which is something that can be easier for end users.
> 
> This would give a config like:
> 
> ```
> useCertManager: true
> issuer:
>   namespace: example
>   name: selfsignedissuer
> ca_cert_data: <NOTHING>
> ...
> ```

This is definitely something that could be nice to add in the future. It would likely not be applicable in the kpack use case because we usually see those registries deployed off cluster so the cert would not be generated by cert-manager, but this might be valuable for other types of workloads. Would you be open to creating an issue in the repo for this?

jorgemoralespou:
> Would you be open to creating an issue in the repo for this?

@tomkennedy513 should this be in the cert-injection-webhook repo or in the package repo?

tomkennedy513:
> > Would you be open to creating an issue in the repo for this?
> 
> @tomkennedy513 should this be in the cert-injection-webhook repo or in the package repo?

We actually don't have a separate repo for the webhook package so https://github.com/vmware-tanzu/cert-injection-webhook is perfect

seemiller (changes requested):
@tomkennedy513 Could you remove the `addons/repos/main.yaml` file from the PR? We will make this change when we are ready to publish the package in in our package repository.

tomkennedy513:
> @tomkennedy513 Could you remove the `addons/repos/main.yaml` file from the PR? We will make this change when we are ready to publish the package in in our package repository.

@seemiller done

seemiller (approved):

PR #3259, opened 2022-02-24
Fix file access issue in kapp-controller package

blc1996:

Add write access to `/etc/pki/tls/certs/ca-bundle.crt` so kapp-controller will have no issue in writing custom certificates into it.

```release-note
Fix a bug that prevents kapp-controller using custom certificates.
```

Fixes: #

Tested package in a local environment.


blc1996:
Move to #3282

PR #3266, opened 2022-02-26
Fix install.sh sudo failure handling

stmcginnis:

Error handling was added to exit if we are unable to sudo when trying to
install the tanzu binary into a system path. This had an error though
where the "exit 1" would be executed outside of the "or", so whether
sudo worked or not it would still exit with a failure.

```release-note
NONE
```

Ran `make release`. Then `cd release/tce-linux-amd64-v0.11.0-dev.1/` to get the full release bundle.
From there, ran `install.sh` and the first time through entered the wrong sudo password three times
to get it to fail. On sudo failure, it correctly echoed the error message and exited with a 1 return code.
Then ran `install.sh` again, this time giving the correct password. Install proceeded and no errors.

### DIFF
--- a/hack/release/install.sh
+++ b/hack/release/install.sh
@@ -32,6 +32,11 @@ case "${BUILD_OS}" in
 esac
 echo "${XDG_DATA_HOME}"
 
+handle_sudo_failure() {
+    echo "sudo access required to install to ${TANZU_BIN_PATH}"
+    exit 1
+}
+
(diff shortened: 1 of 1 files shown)


stmcginnis:
@dvonthenen @jpmcb - much more thorough testing this time. Should work as expected.

davidvonthenen (approved):

PR #3304, opened 2022-03-02
`unmanaged-cluster` kind provider can consume `ProviderConfiguration`

jpmcb:
- Enables the `kind` provider in `unmanaged-cluster` able to consume the `ProviderConfiguration`
- Make `ProviderConfiguration` serialized for kind provider

```release-note
unmanaged-cluster kind provider can consume ProviderConfigurations
```

N/a - Related to several issues where users are asking for features in the `kind` provider that can be configured in the kind configuration

Created the following configuration file:
```yaml
ClusterName: test
KubeconfigPath: ""
ExistingClusterKubeconfig: ""
NodeImage: ""
Provider: kind

ProviderConfiguration: 
  rawKindConfig: |
    kind: Cluster
    apiVersion: kind.x-k8s.io/v1alpha4
    nodes:
    - role: control-plane
      extraPortMappings:
      - containerPort: 888
        hostPort: 888
        listenAddress: "127.0.0.1"
        protocol: TCP
    - role: worker
    - role: worker

Cni: antrea
CniConfiguration: {}
PodCidr: 10.244.0.0/16
ServiceCidr: 10.96.0.0/16
TkrLocation: projects.registry.vmware.com/tce/tkr:v1.21.5
SkipPreflight: false
```

Then deployed a new cluster with the above file:
```
$ tanzu unmanaged-cluster create -f test.yaml asdf
```

And I can successfully see the node configurations applied in the kind configuration:
```
$  k get nodes
NAME                 STATUS   ROLES                  AGE   VERSION
asdf-control-plane   Ready    control-plane,master   49m   v1.21.1
asdf-worker          Ready    <none>                 48m   v1.21.1
asdf-worker2         Ready    <none>                 48m   v1.21.1
```

And the port forwarding worked:
```
$ docker ps
CONTAINER ID   IMAGE                  COMMAND                  CREATED          STATUS          PORTS                                               NAMES
4b6172c08433   kindest/node:v1.21.1   "/usr/local/bin/entr…"   50 minutes ago   Up 50 minutes                                                       asdf-worker2
820246ef2e11   kindest/node:v1.21.1   "/usr/local/bin/entr…"   50 minutes ago   Up 50 minutes                                                       asdf-worker
2775c4b4f3fe   kindest/node:v1.21.1   "/usr/local/bin/entr…"   50 minutes ago   Up 50 minutes   127.0.0.1:888->888/tcp, 127.0.0.1:38469->6443/tcp   asdf-control-plane

```

A warning is logged informing the user that all other configurations will be ignored:
image

---

Existing operations work:
```
$ tanzu unmanaged-cluster create test
```

Fails if kind config is the _wrong type_:
```
$ cat test.yaml
ClusterName: test
KubeconfigPath: ""
ExistingClusterKubeconfig: ""
NodeImage: ""
Provider: kind
ProviderConfiguration:
  rawKindConfig: 123
Cni: antrea
CniConfiguration: {}
PodCidr: 10.244.0.0/16
ServiceCidr: 10.96.0.0/16
TkrLocation: projects.registry.vmware.com/tce/tkr:v1.21.5
PortsToForward: []
SkipPreflight: false
ControlPlaneNodeCount: "1"
WorkerNodeCount: "0"

$ tanzu unmanaged-cluster create test -f test.yaml
...
 Creating cluster test-wrong-type
   Base image downloaded
   Creating cluster.             failed to create cluster, Error: unable to serialize kind config from given ProviderConfiguration. Error was: ProviderConfiguration.rawKindConfig wrong type, expected string
```

Fails if kind config is _not_ valid:
```
$ cat test.yaml
ClusterName: test
KubeconfigPath: ""
ExistingClusterKubeconfig: ""
NodeImage: ""
Provider: kind
ProviderConfiguration:
  rawKindConfig: |
    woof woof
Cni: antrea
CniConfiguration: {}
PodCidr: 10.244.0.0/16
ServiceCidr: 10.96.0.0/16
TkrLocation: projects.registry.vmware.com/tce/tkr:v1.21.5
PortsToForward: []
SkipPreflight: false
ControlPlaneNodeCount: "1"
WorkerNodeCount: "0"

$ tanzu unmanaged-cluster create test-wrong-config -f test.yaml
...
 Creating cluster test-wrong
   Base image downloaded
   Creating cluster.             failed to create cluster, Error: kind returned error: could not determine kind / apiVersion for config: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `woof woof` into encoding.typeMeta

```

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
@@ -50,18 +50,36 @@ type KindClusterManager struct {
 
 // Create will create a new kind cluster or return an error.
 func (kcm KindClusterManager) Create(c *config.UnmanagedClusterConfig) (*KubernetesCluster, error) {
+	var err error
+
 	kindProvider := kindcluster.NewProvider()
 	clusterConfig := kindcluster.CreateWithKubeconfigPath(c.KubeconfigPath)
 
-	// TODO(stmcginnis): Determine what we need to do for kind configuration
-	parsedKindConfig, err := kindConfigFromClusterConfig(c)
+	// Serlize unstructured data into a kindProviderConfig.
+	// Return any error from attempting to read the data
+	serializedProviderConfig, err := serializeKindProviderConfig(c.ProviderConfiguration)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to serialize kind config from given ProviderConfiguration. Error was: %s", err)
 	}
+
+	// If a user has provided something in the ProviderConfiguration,
+	// assume it is a full kind configuration and don't attempt to produce
+	// a configuration from the unmanaged-cluster config
+	parsedKindConfig := []byte(serializedProviderConfig.rawKindConfig)
+
+	// when the parsed kind config from the provider config is empty,
+	// create a kind config from ClusterConfig settings, using the given flags and options
+	if len(parsedKindConfig) < 1 {
+		parsedKindConfig, err = kindConfigFromClusterConfig(c)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate a viable kind config. Error was: %s", err)
+		}
+	}
+	
 	kindConfig := kindcluster.CreateWithRawConfig(parsedKindConfig)
 	err = kindProvider.Create(c.ClusterName, clusterConfig, kindConfig)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("kind returned error: %s", err)
 	}
 
 	// readkubeconfig in bytes
@@ -88,6 +106,26 @@ func (kcm KindClusterManager) Create(c *config.UnmanagedClusterConfig) (*Kuberne
 	}
 
 	return kc, nil
+}
+
+type kindProviderConfig struct {
+	rawKindConfig string
+}
+
+func serializeKindProviderConfig(pc map[string]interface{}) (kindProviderConfig, error) {
+	// Check if key exists. If not, return empty config and continue
+	if _, ok := pc["rawKindConfig"]; !ok {
+		return kindProviderConfig{}, nil
+	}
+
+	// Check if provided data is a string.
+	if _, ok := pc["rawKindConfig"].(string); !ok {
+		return kindProviderConfig{}, fmt.Errorf("ProviderConfiguration.rawKindConfig wrong type, expected string")
+	}
+
+	return kindProviderConfig{
+		pc["rawKindConfig"].(string),
+	}, nil
 }
 
 func kindConfigFromClusterConfig(c *config.UnmanagedClusterConfig) ([]byte, error) {
--- a/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
+++ b/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
@@ -134,6 +134,11 @@ func (t *UnmanagedCluster) Deploy(scConfig *config.UnmanagedClusterConfig) error
 	log.AddLogFile(bootstrapLogsFp)
 	log.Event(logger.FolderEmoji, "Created cluster directory")
 
+	// Log a warning if the user has given a ProviderConfiguration
+	if len(scConfig.ProviderConfiguration) != 0 {
+		log.Style(outputIndent, color.FgYellow).ReplaceLinef("Reading ProviderConfiguration from config file. All other provider specific configs may be ignored.")
+	}
+
 	// 2. Download and Read the TKR
 	log.Event(logger.WrenchEmoji, "Resolving Tanzu Kubernetes Release (TKR)")
 	bomFileName, err := getTkrBom(scConfig.TkrLocation)


jpmcb:
Note: do we want to have a log message that warns the user we are _not_ building a config from any flags that have been provided? Putting anything in the `ProviderConfiguration` _will_ overwrite any other flags given by the user

stmcginnis:
> Note: do we want to have a log message that warns the user we are not building a config from any flags that have been provided? Putting anything in the ProviderConfiguration will overwrite any other flags given by the user

I think that would be a good idea. Check if flags were provided and if so, give some sort of warning that those flags will have no effect.

joshrosso (changes requested):
Going to review this today.

jpmcb:
Warning after it parses the config: 

image

cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
```diff
@@ -134,6 +134,11 @@ func (t *UnmanagedCluster) Deploy(scConfig *config.UnmanagedClusterConfig) error
 	log.AddLogFile(bootstrapLogsFp)
 	log.Event(logger.FolderEmoji, "Created cluster directory")
 
+	// Log a warning if the user has given a ProviderConfiguration
+	if scConfig.ProviderConfiguration != "" {
+		log.Style(outputIndent, color.FgYellow).ReplaceLinef("Reading ProviderConfiguration from config file. All other provider specific configs may be ignored")
```
stmcginnis, line 139: OK, I feel like a jerk posting this nit, but since you start with a full sentence that ends with a period, can you end the second sentence with a period?

Sorry. :]

jpmcb: Grammar! My greatest enemy! 😂 happy to fix!

jorgemoralespou:
I would rather fail, if possible, if the user is providing a `ProviderConfiguration` and conflicting arguments (e.g. portmappings, node counts) so that the user fixes that. It's the best way to educate. The warning might fall in the cracks.

jpmcb:
> I would rather fail, if possible, if the user is providing a ProviderConfiguration and conflicting arguments

Hmmm thanks for the input here - I'm really not sure what the best UX is.

> The warning might fall in the cracks.

That's very true and usually happens with most warnings.

---

My vote is for a warning but will leave it up to others. Maybe @garrying can chime in here?

garrying:
As a reference, and curiosity, I tried a similar scenario with `helm`; passing in a values file plus another conflicting argument. It seems like the conflicting arg was used instead of what's in the yaml file. I wonder if there are other examples.

I feel a warning is sufficient if it were explicit as to what args/config were ignored: `The following arguments were not used:[...]`. For additional visibility, I can imagine a warning output at the end of the operation.

stmcginnis:
The way it is now, all args are ignored. So you can either pass config on the command line, or you can provide a kind config. There is no combination of the two.

Another approach - that sounds like it would match the helm behavior - would be to replace our default base kind config with the one provided by the file, then continue to update that base config with anything provided on the command line. I could see that being useful if you had a common basic config you maintained, but then wanted to use that to "customize" certain deployments by overriding certain properties via command line arguments.

jorgemoralespou:
What @stmcginnis describes is the best experience, but it's more complicated to implement. If that one is done, then no need for warnings. If current behavior, as it's not what a normal user would expect (they would expect more the flags to override the values in the providerconfig) I would error out, as I described above.

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -53,11 +53,20 @@ func (kcm KindClusterManager) Create(c *config.UnmanagedClusterConfig) (*Kuberne
...
+	// assume it is a full kind configuration and don't attempt to produce
+	// a configuration from the unmanaged-cluster config
+	var parsedKindConfig []byte
+	var err error
+	if c.ProviderConfiguration != "" {
+		parsedKindConfig = []byte(c.ProviderConfiguration)
+	} else {
+		parsedKindConfig, err = kindConfigFromClusterConfig(c)
+		if err != nil {
+			return nil, err
```
joshrosso, line 66: We need to improve this error message when the provider configuration is returned.

It's good to return the error for `kind`, but the user needs context that it failed to evaluate the provide config.

image

cli/cmd/plugin/unmanaged-cluster/config/config.go
```diff
@@ -78,7 +78,7 @@ type UnmanagedClusterConfig struct {
 	Provider string `yaml:"Provider"`
 	// ProviderConfiguration offers optional provider-specific configuration.
 	// The exact keys and values accepted are determined by the provider.
-	ProviderConfiguration map[string]interface{} `yaml:"ProviderConfiguration"`
+	ProviderConfiguration string `yaml:"ProviderConfiguration"`
```
joshrosso, line 81: The long-term intent of this key is to offer provider-specific configuration for providers we introduce.

For example, let's say we introduce a VM provider. In this case, we might want to do things like:

```
provideringConfiguration:
    memory: 4
    cpu: 2
```

I wouldn't want to require this as a `string` representation.

If we can't represent an arbitrary config as a `map[string]interface{}` (i grok conversion code could be somewhat complex), then let's change this to `RawProviderConfiguration` and retrain provider configuration.

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -53,11 +53,20 @@ func (kcm KindClusterManager) Create(c *config.UnmanagedClusterConfig) (*Kuberne
...
-	// TODO(stmcginnis): Determine what we need to do for kind configuration
-	parsedKindConfig, err := kindConfigFromClusterConfig(c)
-	if err != nil {
-		return nil, err
+	// If a user has provided something in the ProviderConfiguration,
+	// assume it is a full kind configuration and don't attempt to produce
+	// a configuration from the unmanaged-cluster config
+	var parsedKindConfig []byte
+	var err error
+	if c.ProviderConfiguration != "" {
```
joshrosso, line 61: @jpmcb I wonder whether this `if` logic can be simplified.

For example, why not execute `[]byte(c.ProviderConfiguration)` no matter what? Even if `c.ProviderConfiguration == ""`. wouldn't the result be the same?

With that, this could probably go from `if else` -> `if`.

lmk if I'm missing something.

joshrosso (changes requested):
Changes requested.

Thanks for the this work; it's looking really good.

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
joshrosso: quick example of what i'm asking about: https://go.dev/play/p/UWvKRyh5VWU

cli/cmd/plugin/unmanaged-cluster/config/config.go
stmcginnis: I actually like this as a string. That keeps it simple and generic so each provider can decide what to do with it. So if kind expects it to be yaml, then it can parse the string as yaml. If a new provider expects json, it can parse it as json. We just provide a convenient way to pass that through. If we add an `alexa` provider, then you can pass through `ProviderConfiguration: "alexa, make me a cluster"`. ;)

Just weighing in with an opinion.

joshrosso: Yeah, that perspective makes sense. However, i'm not a fan of every provider having their config be a string when it should be a struct that can be validated against as structured data.

For example, take a look at provider specific configurations that has been proposed for CAPI and the ingress API, I think they were quite wise to offer this in a structured manner.

If you wanted this with your Alex provider, you'd say:

```
providerConfiguration:
    alexaCommandString: "alexa, make ma a cluster"
```

joshrosso: which makes me think, my preference would actually be:

```
providerConfiguration:
    rawKindConfig: |
        kind: Cluster
        apiVersion: kind.x-k8s.io/v1alpha4
        nodes:
        - role: control-plane
          extraPortMappings:
          - containerPort: 888
            hostPort: 888
            listenAddress: "127.0.0.1"
            protocol: TCP
        - role: worker
        - role: worker
```

We probably need to pull this out into an issue and have a design conversation.

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
joshrosso: My recommendation is:

```go
// If a user has provided something in the ProviderConfiguration,
// assume it is a full kind configuration and don't attempt to produce
// a configuration from the unmanaged-cluster config
var err error
parsedKindConfig := []byte(c.ProviderConfiguration)

// when kind config was empty, create a kind config from ClusterConfig settings
if len(parsedKindConfig) < 1 {
    parsedKindConfig, err = kindConfigFromClusterConfig(c)
    if err != nil {
        return nil, fmt.Errorf("failed to generate a viable kind config. error was: %s", err)
    }
}
```

lmk if this makes sense!

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
joshrosso: Looking closer at the code, I now see there is no pre-config validation before running the cluster create. So my recommendation is to improve the error as described in https://github.com/vmware-tanzu/community-edition/pull/3304/files#r819755863.

Then, simply mention that `kind` returned the error. So my suggestion would be:

```go
kindConfig := kindcluster.CreateWithRawConfig(parsedKindConfig)
err = kindProvider.Create(c.ClusterName, clusterConfig, kindConfig)
if err != nil {
    return nil, fmt.Errorf("kind returned error: %s", err)
}

```

joshrosso (approved):
Works great. Thanks @jpmcb! :rocket:

jorgemoralespou:
@jpmcb please, could the example be updated to reference the TCE's kind node image in the example you provide? otherwise a regular kind cluster will be created.

```
ClusterName: test
KubeconfigPath: ""
ExistingClusterKubeconfig: ""
NodeImage: ""
Provider: kind

ProviderConfiguration: 
  rawKindConfig: |
    kind: Cluster
    apiVersion: kind.x-k8s.io/v1alpha4
    nodes:
    - role: control-plane
      image: projects.registry.vmware.com/tce/kind/node:v1.22.5
      extraPortMappings:
      - containerPort: 888
        hostPort: 888
        listenAddress: "127.0.0.1"
        protocol: TCP
    - role: worker
      image: projects.registry.vmware.com/tce/kind/node:v1.22.5
    - role: worker
      image: projects.registry.vmware.com/tce/kind/node:v1.22.5

Cni: antrea
CniConfiguration: {}
PodCidr: 10.244.0.0/16
ServiceCidr: 10.96.0.0/16
TkrLocation: projects.registry.vmware.com/tce/tkr:v1.22.5
SkipPreflight: false
```

jpmcb:
@jorgemoralespou - good catch: I'll be doing a docs PR soon here and will capture this in those docs.

PR #3305, opened 2022-03-02
Fix continuous reconciliation in load-balancer-and-ingress-service

evankanderson:

Remove empty `status` fields on Gateway and GatewayClass so that `kapp-controller` doesn't continuously try to reconcile them. When these fields are set but empty, `kapp-controller` will attempt to clear them, and then apiserver will 

```release-note
Fixes continuous reconciliation of the load-balancer-and-ingress-service package.
```

Examined CRD definitions, `kapp-controller` output, and reviewed documentation from other systems.

I didn't load this onto a VSphere cluster for testing; I'm not sure how to add a test for this in this repo, but this can be recognized when "soaking" a cluster for e.g. 4 hours. If this is a problem, you will see approx 40 ConfigMaps (one every 5 minutes) in the `tkg-system` namespace with the format `load-balancer-and-ingress-service-ctrl-change-xyz12`


HanFa (approved):
LGTM, thanks @evankanderson for the fix. I will test these out in the downstream

PR #3334, opened 2022-03-03
Implements `unmanged-cluster docs` command

jpmcb:
- Provides docs generation in multiple formats of the `unmanaged-cluster` commands in support of having a
  local resource for users to reference
- Adds docs for exit codes

```release-note
Added unmanaged-cluster docs command for local documentation generation
```

N/a

Print docs to stdout. The default is markdown

```
$ tanzu unmanaged-cluster docs
## unmanaged-cluster completion

Generate the autocompletion script for the specified shell
...
```

Print docs to stdout in a different format
```
$ tanzu unmanaged-cluster docs --format rest
.. _unmanaged-cluster_completion:

unmanaged-cluster completion
----------------------------

Generate the autocompletion script for the specified shell
...
```

Don't print to `stdout`, dump to current working dir

```
$ tanzu unmanaged-cluster docs --format rest --no-stdout
$ tree
├── unmanaged-cluster_completion_bash.rst
├── unmanaged-cluster_completion_fish.rst
├── unmanaged-cluster_completion_powershell.rst
├── unmanaged-cluster_completion.rst
├── unmanaged-cluster_completion_zsh.rst
├── unmanaged-cluster_configure.rst
├── unmanaged-cluster_create.rst
├── unmanaged-cluster_delete.rst
├── unmanaged-cluster_docs.rst
├── unmanaged-cluster_list.rst
└── unmanaged-cluster.rst
```

Or the equivalent command is:
```
$ tanzu unmanaged-cluster docs --format rest --output-dir ./
$ tree
├── unmanaged-cluster_completion_bash.rst
├── unmanaged-cluster_completion_fish.rst
├── unmanaged-cluster_completion_powershell.rst
├── unmanaged-cluster_completion.rst
├── unmanaged-cluster_completion_zsh.rst
├── unmanaged-cluster_configure.rst
├── unmanaged-cluster_create.rst
├── unmanaged-cluster_delete.rst
├── unmanaged-cluster_docs.rst
├── unmanaged-cluster_list.rst
└── unmanaged-cluster.rst
```

Set the output directory to dump the files
```
$ tanzu unmanaged-cluster docs --format rest --output-dir /tmp
$ tree /tmp
├── unmanaged-cluster_completion_bash.rst
├── unmanaged-cluster_completion_fish.rst
├── unmanaged-cluster_completion_powershell.rst
├── unmanaged-cluster_completion.rst
├── unmanaged-cluster_completion_zsh.rst
├── unmanaged-cluster_configure.rst
├── unmanaged-cluster_create.rst
├── unmanaged-cluster_delete.rst
├── unmanaged-cluster_docs.rst
├── unmanaged-cluster_list.rst
└── unmanaged-cluster.rst
```

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/docs.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/docs.go
@@ -1,0 +1,170 @@
+// Copyright 2022 VMware Tanzu Community Edition contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
+)
+
+const docsDesc = `
+Provides documentation and usage for unmanaged-cluster in multiple formats.
+Gives all commands, their flags, and related commands.
+This can then be used as a local documentation resource when working with unmanaged-clusters.
+
+Users may generate man pages that can then be zipped and placed on the system.
+The 'man' program expects the zipped bundle to be in a location defined by 'manpath'
+and the user must run 'mandb' to inject the manpages onto their system.
+`
+
+var (
+	docsFormat    = ""
+	docsOutputDir = ""
+	noStdout      = false
+)
+
+// ListCmd returns a list of existing clusters.
+var DocsCmd = &cobra.Command{
+	Use:   "docs",
+	Short: "Generated documentation & usage of unmanaged-cluster",
+	Long:  docsDesc,
+	PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+		return nil
+	},
+	RunE: startDocsGen,
+	PostRunE: func(cmd *cobra.Command, args []string) (err error) {
+		return nil
+	},
+}
+
+func init() {
+	DocsCmd.Flags().StringVar(&docsFormat, "format", "markdown", "Output format (markdown|manpage|rest|yaml); default is markdown")
+	DocsCmd.Flags().BoolVar(&noStdout, "no-stdout", false, "Ignores printing to stdout and inspects 'output-dir' for location to dump files")
+	DocsCmd.Flags().StringVar(&docsOutputDir, "output-dir", "", "Location to output the doc page files; providing this prevents printing to stdout. Defaults to the current working dir")
+}
+
+// list outputs a list of all unmanaged clusters on the system.
+func startDocsGen(cmd *cobra.Command, args []string) error {
+	var err error
+
+	if len(args) != 0 {
+		return fmt.Errorf("docs generation does not support any arguments")
+	}
+
+	// Valid formats
+	if (docsFormat != "markdown" && docsFormat != "manpage") && (docsFormat != "rest" && docsFormat != "yaml") {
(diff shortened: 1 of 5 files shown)


cli/cmd/plugin/unmanaged-cluster/cmd/docs.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
```
stmcginnis, line 1: 2022?

cli/cmd/plugin/unmanaged-cluster/cmd/docs.go
```diff
@@ -0,0 +1,170 @@
...
+		if err := doc.GenMan(cmd.Root(), manHeader, buf); err != nil {
+			return err
+		}
+
+		for _, c := range cmd.Root().Commands() {
+			if err := doc.GenMan(c, manHeader, buf); err != nil {
+				return err
+			}
+		}
+	case "rest":
```
stmcginnis, line 142: "rest" probably isn't the right term to use for reStructuredText. I would keep it "rst"

jpmcb: I was mostly going off what `cobra` has documented: https://github.com/spf13/cobra/blob/master/doc/rest_docs.md

Looks like it's usually referred to as `ReST` which seemed wonky as a config option

jpmcb: Looks like on the official spec it gets refferred to as either `reStructuredText` or `ReST`: https://docutils.sourceforge.io/rst.html

stmcginnis: Huh, I've spent more time with RST having been in Python land for years, and I've never seen it referred to "rest". I would take that to mean some sort of API. :D 

If that's how they have it documented, then I guess that's good enough for me!

stmcginnis:
Kind of a meta-question, but does it make sense to implement doc generation per command? Or should we do something like they've done in `tanzu-framework`? https://github.com/vmware-tanzu/tanzu-framework/blob/main/Makefile#L587-L589

jpmcb:
@stmcginnis - good question! I had considered also implementing something like

```
$ tanzu unmanaged-cluster docs create
```

That would accept a set of valid args (the current commands we have) to print/dump into a file _just_ that command.

But sounds like you're more asking about _intelligent_ inspection of the current commands. Since `unmanaged-cluster` _doesn't_ support adding dynamically adding user plugins/commands, I don't think this is something we need to worry about right now.

stmcginnis:
> Since unmanaged-cluster doesn't support adding dynamically adding user plugins/commands, I don't think this is something we need to worry about right now.

Kind of wrong direction of my point - I don't think we should add documentation commands to every plugin. If we add it to `unamanged-cluster`, then we really should add it to all of the other (`managed-cluster`, `diagnostics`, etc) command. That's a lot of duplication of effort.

I think it would probably make more sense to change `tanzu generate-all-docs` to not be hidden. Then add additional arguments if we want to be able to generate docs in formats other than markdown.

jpmcb:
> I don't think we should add documentation commands to every plugin.
...
I think it would probably make more sense to change tanzu generate-all-docs to not be hidden. 

Hmmm that's a great point; I think this is actually work that could go upstream into tanzu framework. I'll see if they are open to doing something like this. For now, closing.

PR #3340, opened 2022-03-03
Delay pruning binaries until after the update buckets have been uploaded

davidvonthenen:

This is true for the other upload GCP workflows as well... we are deleting the binaries too soon. We need to build them in the right directory structure first:
https://github.com/vmware-tanzu/community-edition/blob/main/.github/workflows/release-ga-bucket.yaml#L67

But we had an oops where `prune-buckets` was added to the `release-buckets` target. We need to upload to the update GCP buckets, then call prune which we are doing here already:
https://github.com/vmware-tanzu/community-edition/blob/main/.github/workflows/release-ga-bucket.yaml#L102

Then upload the release GCP buckets.

```release-note
Delay pruning binaries until after the update buckets have been uploaded
```

NA

NA

NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,7 @@ prune-buckets:
 
 # The main target for GCP buckets. Please see above
 .PHONY: release-buckets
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Makes sense!

PR #3371, opened 2022-03-07
Fix dependabot.yaml

lubronzhan:

Fix dependabot since pipeline is failing

```release-note

```

Fixes: #https://github.com/vmware-tanzu/community-edition/issues/3372

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -43,13 +43,6 @@ updates:
       - dependency-name: "tanzu-framework"
 
   - package-ecosystem: "gomod"
-    directory: "hack/builder"
-    schedule:
-      interval: "daily"
-    ignore:
-      - dependency-name: "tanzu-framework"
-
-  - package-ecosystem: "gomod"
     directory: "hack/packages"
     schedule:
       interval: "daily"
@@ -143,6 +136,9 @@ updates:
   - package-ecosystem: "gomod"
     directory: "addons/packages/harbor/2.2.3/test"
     schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "tanzu-framework"
 
   - package-ecosystem: "gomod"
     directory: "addons/packages/calico/3.11.3/test"


lubronzhan:
I used this fix in CPI related PR, the pipeline passed. Then I removed this from that PR since it's unrelated to CPI

.github/dependabot.yml
```diff
@@ -42,13 +42,6 @@ updates:
     ignore:
       - dependency-name: "tanzu-framework"
 
-  - package-ecosystem: "gomod"
```
stmcginnis, line 45: This one shouldn't be removed as the `schedule` configuration is fine.

.github/dependabot.yml
```diff
@@ -140,10 +133,6 @@ updates:
...
-    directory: "addons/packages/harbor/2.2.3/test"
-    schedule:
```
stmcginnis, line 145: Rather than removing, this should actually have a schedule value set so it doesn't error on `none`.

stmcginnis (changes requested):
I don't think we want to remove these packages from the configuration.

.github/dependabot.yml
lubronzhan: Hi @stmcginnis did you check the failure of the dependabot pipeline? This line is duplicate of this https://github.com/vmware-tanzu/community-edition/blob/0a24ed9bace713bce0e8b6031da1a1009dd3b29f/.github/dependabot.yml#L10-L15

.github/dependabot.yml
lubronzhan: Hi @stmcginnis , this package is already  defined here https://github.com/vmware-tanzu/community-edition/blob/main/.github/dependabot.yml#L136-L141
Probably a wrong rebase result

stmcginnis: That one is for harbor 2.3.3, but this one is for 2.2.3. So I think we still want this, just with an actual schedule defined. Looks like maybe a copy/paste problem.

.github/dependabot.yml
stmcginnis: The pipeline error I saw was that "schedule" could not be none, so I think the main issue is the one below. But yeah, line 10 already has this one, so +1 to removing it then.

.github/dependabot.yml
lubronzhan: Thanks. I didn't notice the version. Let me add the same schedule here

stmcginnis (approved):
Thanks @lubronzhan!

PR #3383, opened 2022-03-07
Post daily build notifications to an existing pinned discussion

davidvonthenen:

This changes the behavior for the notification of daily builds to post a reply to a pinned existing discussion (https://github.com/vmware-tanzu/community-edition/discussions/3379) instead of creating a new discussion in the "Daily Builds" category. This change was done because new discussions flood the "discussion" view when first opening the page.

Once this PR gets merged, will delete all older threads to clean up the mess and delete the "Daily Build" category since it will no longer be used.

```release-note
Post daily build to an existing pinned discussion
```

NA

Ran functionality manually

NA

### DIFF
--- a/hack/dailybuild/dailybuild.go
+++ b/hack/dailybuild/dailybuild.go
@@ -234,7 +234,6 @@ func releaseNotes() (string, error) {
 }
 
 func postDiscussion(token string, builds *Builds, releaseNotes string) error {
-	titleTemplate := "Daily Development Build for %s"
 	bodyTemplate := "# Downloadable assets:\\n\\n" +
 		"Linux AMD64 - %s\\n" +
 		"Darwin AMD64 - %s\\n" +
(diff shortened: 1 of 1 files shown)


hack/dailybuild/dailybuild.go
```diff
@@ -244,15 +243,15 @@ func postDiscussion(token string, builds *Builds, releaseNotes string) error {
 	} else {
 		bodyTemplate += "%s" // this effectively becomes a noop because it will be empty
 	}
+
 	jsonTemplate := "{\n" +
-		"\"query\": \"mutation {   createDiscussion(input: {repositoryId: \\\"MDEwOlJlcG9zaXRvcnkzMDM4MDIzMzI\\\", categoryId: \\\"DIC_kwDOEhun3M4CA9pd\\\", body: \\\"%s\\\", title: \\\"%s\\\"}) {     discussion {       id     }   } }\"\n" +
+		"\"query\": \"mutation {   addDiscussionComment(input: {discussionId: \\\"D_kwDOEhun3M4AO9Ut\\\", body: \\\"%s\\\"}) { comment {       id     }   } }\"\n" +
```
stmcginnis, line 248: Looks good according to https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions#adddiscussioncomment

stmcginnis (approved):
Everything looks right to me. Let's give it a shot and address anything if it comes up.

Not for future consideration - I wonder if we should consider yearly or monthly threads to limit the number shown in each. Not sure if GitHub enforces a limit on discussion length, but might make it a little easier to find things.

But that said, this will make it much more organized than what we have today and I think it looks great.

dvonthenen:
> Everything looks right to me. Let's give it a shot and address anything if it comes up.
> 
> Not for future consideration - I wonder if we should consider yearly or monthly threads to limit the number shown in each. Not sure if GitHub enforces a limit on discussion length, but might make it a little easier to find things.
> 
> But that said, this will make it much more organized than what we have today and I think it looks great.

The monthly threads sounds like a great idea... when I get some time, I think I can easily make that change

PR #3468, opened 2022-03-09
packages/pinniped: jwtauthenticator can be configured with an audience cluster uid

benjaminapetersen:
packages/pinniped: jwtauthenticator can be configured with an audience that is a cluster uid rather than a cluster name.

Signed-off-by: Benjamin A. Petersen <ben@benjaminapetersen.me>

Historically we have been able to trust that a workload cluster `name` is unique.  However, it is best to use the cluster `uid` to ensure a unique value.

```release-note
The concierge cluster audience can be a cluster UID instead of a cluster name to ensure uniqueness. 
```

Fixes: #


addons/packages/pinniped/0.12.1/bundle/config/values.yaml
```diff
@@ -48,6 +48,9 @@ pinniped:
     tag: DEPRECATED
     repository: DEPRECATED
     pull_policy: DEPRECATED
+  #! NOTE: this must be on a workload cluster!!
```
benjaminapetersen, line 51: WIP, will remove notes when done.

addons/packages/pinniped/0.12.1/bundle/config/upstream/03-post-deploy.yaml
```diff
@@ -64,7 +64,9 @@ spec:
       restartPolicy: Never
       containers:
         - name: pinniped-post-deploy
-          image: projects.registry.vmware.com/tce/tanzu_core/addons/tkg-pinniped-post-deploy:v1.3.1
+          # BEN: update with a alt version of post-deploy job!
```
benjaminapetersen, line 67: WIP: will revert when done testing

addons/packages/pinniped/0.12.1/bundle/examples/mc-ldap.concierge-audience.yaml
```diff
@@ -0,0 +1,120 @@
+#@data/values
```
benjaminapetersen, line 1: WIP, will remove this file when done testing.

ankeesler (dismissed):
approve, but would love to do away with those images.yml changes if we don't need them

addons/packages/pinniped/0.12.1/bundle/.imgpkg/images.yml
```diff
@@ -5,21 +5,21 @@ images:
     kbld.carvel.dev/id: docker.io/getpinniped/pinniped-server:v0.12.1
```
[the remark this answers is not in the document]
benjaminapetersen: Nice, thx, agree this is leftovers from swapping branches, removing!

ankeesler (approved):

PR #3474, opened 2022-03-09
Perform signing check with source code from the release and not main

davidvonthenen:

When we cut a release, the signing check workflows is triggered off main and not off source code the release. So, for the signing check we want to checkout source code from the tag of the release and run those check scripts instead.

This is definitely an experiment, but one which we can revert the changes if this doesn't work. Depending on the input, there is another variation on this ref value and that would be `ref/tags/<version>`

```release-note
NONE
```

NA

NA

NA


joshrosso (dismissed):
approving. @dvonthenen is going to merge later when he's ready to validate.

joshrosso (approved):

PR #3476, opened 2022-03-09
Bump kapp-controller to 0.33.1

neil-hickey:
Signed-off-by: Neil Hickey <nhickey@vmware.com>

Bump kapp-controller 0.33.1 (latest)

```release-note
Bump kapp-controller 0.33.1 (latest)
```

Fixes: #

### DIFF
--- a/addons/packages/kapp-controller/0.33.1/bundle/.imgpkg/images.yml
+++ b/addons/packages/kapp-controller/0.33.1/bundle/.imgpkg/images.yml
@@ -1,0 +1,7 @@
+---
+apiVersion: imgpkg.carvel.dev/v1alpha1
+images:
+- annotations:
+    kbld.carvel.dev/id: ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:a97d9d43b8b5d996f6510ae88ee761ac0f1d883d04fa7402b32f8ce6f9f875bd
+  image: ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:a97d9d43b8b5d996f6510ae88ee761ac0f1d883d04fa7402b32f8ce6f9f875bd
+kind: ImagesLock
--- a/addons/packages/kapp-controller/0.33.1/bundle/config/overlays/update-deployment.yaml
+++ b/addons/packages/kapp-controller/0.33.1/bundle/config/overlays/update-deployment.yaml
@@ -1,0 +1,70 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("/values.star", "values", "generateBashCmdForDNS")
+#@ load("@ytt:yaml", "yaml")
+
+#@overlay/match by=overlay.subset({"kind":"Deployment","metadata":{"name": "kapp-controller"}})
+---
+metadata:
+  annotations:
+    #@overlay/match missing_ok=True
+    kapp.k14s.io/change-rule: "upsert after upserting apps.kappctrl.k14s.io/kapp-controller-config"
+spec:
+  template:
+    #@overlay/match-child-defaults missing_ok=True
+    spec:
+      containers:
+      #@overlay/match by=overlay.subset({"name":"kapp-controller"})
+      - args:
+        #@overlay/match by=overlay.subset("-packaging-global-namespace=kapp-controller-packaging-global")
+        - #@ "-packaging-global-namespace={}".format(values.kappController.globalNamespace)
+        #@overlay/append
+        - #@ "-concurrency={}".format(values.kappController.deployment.concurrency)
+        #@overlay/append
+        - #@ "-metrics-bind-address={}".format(values.kappController.deployment.metricsBindAddress)
+        ports:
+          #@overlay/match by="name"
+          - name: api
+            containerPort: #@ values.kappController.deployment.apiPort
+        env:
+          #@overlay/match by="name"
+          - name: KAPPCTRL_API_PORT
+            value: #@ str(values.kappController.deployment.apiPort)
+        #@ if values.kappController.deployment.coreDNSIP:
+        volumeMounts:
+          - mountPath: /etc
+            name: etc
+        #@ end
+      #@ if values.kappController.deployment.coreDNSIP:
+      #! Using init container bypasses the restriction of not having root access in main container
+      #! It modifies /etc/resolv.conf which is shared to main container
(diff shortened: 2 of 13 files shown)


addons/packages/kapp-controller/0.33.1/bundle/.imgpkg/images.yml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: imgpkg.carvel.dev/v1alpha1
+images:
+- annotations:
+    kbld.carvel.dev/id: ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:4141b93108b1f3431e786638a115a7e0876d756e7a13acfddd7aaa3bae41e202
```
blc1996, line 5: `ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:4141b93108b1f3431e786638a115a7e0876d756e7a13acfddd7aaa3bae41e202` is the image of 0.30.1

addons/packages/kapp-controller/0.33.1/bundle/config/overlays/update-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+      #! It modifies /etc/resolv.conf which is shared to main container
+      initContainers:
+      - args:
+        - -c
+        - #@ generateBashCmdForDNS(values.kappController.deployment.coreDNSIP)
+        command:
+        - /bin/sh
+        #! Beware, update this image with each version bump!!!!!!!
+        #! This image needs to be the same as the image used in basefile
+        image: ghcr.io/vmware-tanzu/carvel-kapp-controller@sha256:4141b93108b1f3431e786638a115a7e0876d756e7a13acfddd7aaa3bae41e202
```
blc1996, line 48: Please update this to use new image

addons/packages/kapp-controller/0.33.1/bundle/config/values.star
```diff
@@ -0,0 +1,18 @@
...
+else:
+    kappNamespace = values.namespace
+end
+
+def generateBashCmdForDNS(coreDNSIP):
+    # This command added the coreDNS IP as the first entry of resolv.conf
+    # In this way, Kapp Controller will have cluster IP access,
+    # and still able to resolve enternal urls when core DNS is unavailable
+
+    return "cp /etc/resolv.conf /etc/resolv.conf.bak; sed '1 i nameserver " + coreDNSIP + "' /etc/resolv.conf.bak > /etc/resolv.conf; rm /etc/resolv.conf.bak; cp -R /etc/* /kapp-etc; chmod g+w /kapp-etc/pki/tls/certs/ca-bundle.crt && chgrp 2000 /kapp-etc/pki/tls/certs/ca-bundle.crt"
```
blc1996, line 17: I saw some securityContext changes in upstream basefile. Please confirm that those doesn't introduce issues for accessing `/kapp-etc/pki/tls/certs/ca-bundle.crt` file.

`generateBashCmdForDNS ` was added to enable cluster DNS access for kapp-controller. Basically, it uses a init container to edit a volumeMount which has the content of `/etc`, and later attach it to kapp-controller's `/etc`

neil-hickey:
Steps followed:

1. edit vendir.yml to change tag.
2. `make vendir-sync-package PACKAGE=kapp-controller VERSION=0.33.1`
3. `make lock-package-images PACKAGE=kapp-controller VERSION=0.33.1`

I think step 3 above needed to be `edit config/overlays/update-deployment.yaml` to update `init-kapp-controller`'s image tag BEFORE you run `make lock-package-images`

addons/packages/kapp-controller/0.33.1/bundle/config/values.star
neil-hickey: hey @blc1996 What is the context that this starlark code gets run under? We removed the `kapp-controller` user group from our Dockerfile. And so I made a commit here that removes the part where you change the group, but I can't be sure if the command still works without knowing if this is run as `root` or otherwise. I don't think we realized you all were making changes to the ca-crt like this.

blc1996: This code runs as command in initContainer (running as root). It serves the purpose of adding core DNS IP into `/etc/resolv.conf` (only root user can edit) so that kapp-controller can resolve in-cluster service URLs.

We didn't make changes to cert files, the chmod there is just to make sure that kapp-controller container have the write access to the file.

neil-hickey: gotcha, in that case I think this should still work as is (and removing the `chgrp 2000 /kapp-etc/pki/tls/certs/ca-bundle.crt`)

blc1996: Basically, we want to add an entry to `/etc/resolv.conf`, but can't do it within kapp-controller container because it's not running as root.
We added a volumeMount and use a initContainer to copy everything in `/etc` to this volumeMount, and also edit `/etc/resolv.conf` to add the core DNS IP. We then attach this volumeMount to `/etc` of kapp-controller container. Kapp-controller will be read from this volumeMount. Also write to `/etc/pki/tls/certs/ca-bundle.crt`, so we need to make sure it has the right access privilage.

neil-hickey: Actually I wonder if we need to add `chgrp 2000 /kapp-etc/pki/tls/certs/ca-bundle.crt` back in here? @benmoss 

Let's hold off on merging until I can confirm

blc1996 (dismissed):
LGTM

neil-hickey:
Looks like kapp-controller has some spelling mistakes in our CRDs. I imagine someone can overwrite the blocker? This is a good prompt for us to fix them though 🥇

neil-hickey:
hey @blc1996 we are good to go on this one!

neil-hickey:
Hi @stmcginnis ! We have gone ahead and rebased, let us know if there's anything else needed to merge!

stmcginnis:
Awesome, thanks @neil-hickey!! Will take a look shortly.

stmcginnis (commented):
Everything looks good to me, but would love to get @seemiller to take a look.

joshrosso (approved):

Issue #882, opened 2021-06-29
Introduce versioning of documentation site to TCE


carlisia:
Some of what's involved for how Velero does it:
https://velero.io/docs/v1.6/release-instructions/#changelog-and-docs-pr
https://github.com/vmware-tanzu/velero/blob/main/Makefile#L351-L361
https://github.com/vmware-tanzu/velero/blob/main/hack/release-tools/gen-docs.sh#L48

kcoriordan:
moved this to 0,11,0 as we are not introducing doc versioning for 010.0

joshrosso:
@kcoriordan is investigating existing models for versioning.

If the project needs to figure this out, we'll switch ownership from Kay to engineering resources.

kcoriordan:
1. In `config.yaml ` added v`0.10.0 `under `doc versions` and `release_versions` and changed `doc_latest` to `v0.10.0 `

PR #3490, opened 2022-03-10
Add script and Makefile for docs versioning

jonasrosland:

This prepares the repository for docs versioning. There are still changes that needs to be done by @kcoriordan, such as:
- Migrating all docs from the `latest` directory
- Verifying all docs after the move
- Changes to `docs_versions` and `docs_latest` in `config.yaml`

Once those changes are PRd in we'll lock the website deployment in Netlify, merge Kay's changes, verify functionality, and then unlock the Netlify deployments.

```release-note
Implemented docs versioning scripts
```

Fixes: #
#882

### DIFF
--- a/docs/site/load-docs.sh
+++ b/docs/site/load-docs.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+# Original file from the Open Policy Agent project: https://github.com/open-policy-agent/opa/blob/master/docs/website/scripts/load-docs.sh
+# Edits based on the Harbor project: https://github.com/goharbor/website/blob/main/load-docs.sh
+
+set -xe
+
+ORIGINAL_COMMIT=$(git symbolic-ref -q --short HEAD || git name-rev --name-only HEAD)
+# If no name can be found "git name-rev" returns
+# "undefined", in which case we'll just use the
+# current commit ID.
+if [[ "${ORIGINAL_COMMIT}" == "undefined" ]]; then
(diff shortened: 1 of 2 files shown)


kcoriordan (approved):

docs/site/load-docs.sh
```diff
@@ -0,0 +1,127 @@
...
+
+echo "Saving current workspace state"
+STASH_TOKEN=$(od -A n -t d -N 3 /dev/urandom |tr -d ' ')
+git stash push --include-untracked -m "${STASH_TOKEN}"
+
+function restore_tree {
+    echo "Returning to commit ${ORIGINAL_COMMIT}"
+    git checkout ${ORIGINAL_COMMIT}
+
+    # Only pop from the stash if we had stashed something earlier
```
stmcginnis, line 64: Running `make prepare` fails for here:

```
error: could not restore untracked files from stash
The stash entry is kept in case you need it again.
make: *** [Makefile:2: prepare] Error 1
```

Looks like it's because of an earlier run and untracked files for the `docs/site/content/docs/[main|v0.10]` files. Maybe we need to have some sort of cleanup at the end for running locally? Probably not as important if we only expect this to run on an ephemeral GitHub Action runner.

stmcginnis (approved):
Looks good. Verified I could run `make prepare` and `make production-build` locally. Only thing I noticed is pointed out inline, but I don't think that's something blocking for moving ahead with this.

PR #3492, opened 2022-03-10
Remove/redirect /plugins

jacksoncvm:
https://tanzucommunityedition.io/plugins/ is just Lorem ipsum. 

```release-note
NONE
```

Fixes: #


stmcginnis (approved):
Thanks! I think this was a placeholder for a direction we didn't end up going. Not even sure if we really need the redirect.

@kcoriordan can you confirm there is no reason to keep this? It looks good to me, but will wait to merge until we get some confirmation.

PR #3556, opened 2022-03-16
Add preflight warnings

### LINKED ISSUE #3555: Support warnings for pre-flight checks in unmanaged-cluster
pre-flight checks should be able to return warnings and errors.

* errors: what we have today; fails the creation unless the user does `--skip-preflight`
* warnings: bubbles up to the user but continues creations. Should print in yellow when `tty` is true

As a first example warning, let's bubble up a warning for users on arm64, stating:

```txt
arm64 architecture detected. Support is currently experimental. Some packages may not install due to their arm64 image not being available.
```

stmcginnis:

We had previously only supported returning errors from preflight checks
when certain conditions were not met. We now have experimental support
for ARM64 platforms, so we would like to also have the ability to return
warnings from the results of our preflights.

This plumbs in the ability for our preflight calls to return warning
messages. It also adds a check for arm64 and uses the warning mechanism
to alert the user that this is an experimental platform that may not
entirely work.

```release-note
Unmanaged clusters will now only warn if running on an arm64 platform. Support for this platform is only experimental at this point.
```

Fixes: #3555

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/cluster.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/cluster.go
@@ -39,7 +39,8 @@ type Manager interface {
 	// PreflightCheck performs any pre-checks that can find issues up front that
 	// would cause problems for cluster creation. Returns nil if there are no
 	// errors found, otherwise a list of the errors that need to be resolved.
-	PreflightCheck() []error
+	// A list of warning messages can be returned that are not blocking errors.
+	PreflightCheck() ([]string, []error)
 	// ProviderNotify returns any provider specific notifications or messages.
 	// Each string will be displayed on its own line.
 	ProviderNotify() []string
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
@@ -291,28 +291,23 @@ func (kcm KindClusterManager) Prepare(c *config.UnmanagedClusterConfig) error {
 
 // PreflightCheck performs any pre-checks that can find issues up front that
 // would cause problems for cluster creation.
-func (kcm KindClusterManager) PreflightCheck() []error {
+func (kcm KindClusterManager) PreflightCheck() ([]string, []error) {
 	// Check presence of docker
 	cmd := exec.Command("docker", "ps")
 	if err := cmd.Run(); err != nil {
 		// In this case we can't check the rest of the settings, so just return
 		// the one error.
-		return []error{
-			fmt.Errorf("docker is not installed or not reachable. Verify it's installed, running, and your user has permissions to interact with it. Error when attempting to run docker ps: %w", err)}
+		return []string{}, []error{fmt.Errorf("docker is not installed or not reachable. Verify it's installed, running, and your user has permissions to interact with it. Error when attempting to run docker ps: %w", err)}
 	}
 
 	// Get Docker info
 	cmd = exec.Command("docker", "info", "--format", "{{ json . }}")
 	output, err := exec.Output(cmd)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -329,18 +327,20 @@ type dockerInfo struct {
...
+		return nil, []error{errors.New("unable to parse Docker information")}
 	}
 
-	issues := []error{}
-
 	if !strings.HasSuffix(info.Architecture, "x86_64") {
-		// Only amd64 supported right now, no arm
-		issues = append(issues, errors.New("only amd64 architecture currently supported"))
+		// Only amd64 supported right now, arm is experimental. Anything else is not supported.
+		if strings.HasSuffix(info.Architecture, "arm64") {
```
joshrosso, line 339: the correct suffix is `aarch64`

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -329,18 +327,20 @@ type dockerInfo struct {
...
+			warnings = append(warnings, "Arm64 architecture detected. Support is currently experimental. Some packages may not install due to their arm64 image not being available.")
+		} else {
+			issues = append(issues, errors.New("only amd64 architecture currently supported"))
```
joshrosso, line 342: this message needs to be updated

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
```diff
@@ -353,7 +353,7 @@ func validateDockerInfo(output []byte) []error {
 		issues = append(issues, fmt.Errorf("minimum %d GiB of memory is required", (minMemoryBytes/1024/1024/1024)))
 	}
 
-	return issues
+	return
```
joshrosso, line 356: seems strange, we don't need to return anything?

joshrosso: oh you're doing named return params,

can we switch back to not named ones?

joshrosso (changes requested):
few comments, i'd like to switch back to non-named return parameters for consistency in the code base. If we really need more clarity on what we're returning, let's use a struct.

thanks for this quick turn around!

cli/cmd/plugin/unmanaged-cluster/cluster/cluster.go
```diff
@@ -39,7 +39,8 @@ type Manager interface {
 	// PreflightCheck performs any pre-checks that can find issues up front that
 	// would cause problems for cluster creation. Returns nil if there are no
 	// errors found, otherwise a list of the errors that need to be resolved.
```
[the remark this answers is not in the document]
stmcginnis: Like on the next line? :) Or something else?

cli/cmd/plugin/unmanaged-cluster/cluster/kind.go
stmcginnis: Hmm, we may need a list of acceptable variations then as `arm64` is also a valid one.

stmcginnis: Follow up - I was crossing the go arch values with the architecture returned by `docker info`. `aarch64` is the correct one from docker.

cli/cmd/plugin/unmanaged-cluster/cluster/cluster.go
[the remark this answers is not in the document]
jpmcb: > Returns nil if there are no errors found, otherwise a list of the errors that need to be resolved.

makes it sound like there's only a list of errors that gets returned 

but just a nit :) feel free to ignore!

joshrosso (approved):

PR #3550, opened 2022-03-16
Introduce ARM64 support to unmanaged-cluster

joshrosso:
This provides initial ARM64 support for unmanaged-clusters. It does not
produce ARM64 binaries, so users will still need to execute against
amd64 binaries. Additionally, some packages do not have arm64 images,
thus they will fail to install.

## Details for the Release Notes (PLEASE PROVIDE)

```release-note
Initial ARM support for unmanaged clusters
```

## Which issue(s) this PR fixes

Fixes: https://github.com/vmware-tanzu/community-edition/issues/3529

## Describe testing done for PR

Create an unmanaged-cluster

PR #3560, opened 2022-03-16
Switch to OVS kernel datapath for antrea cni

### LINKED ISSUE #3058: Can not connect to a Cluster after it has restarted using Antrea CNI (docker provider)
## Bug Report
When creating a cluster with Docker provider (managed or unmanaged) and the cluster is restarted you can no longer connect to the cluster's API server when Antrea is the CNI. It works when using calico as the CNI.

Some pointers from the TCE team, says that [this antrea issue](https://github.com/antrea-io/antrea/issues/1113) might be related and that it's also related to the [code in here](https://github.com/antrea-io/antrea/blob/e23cf3b5714d15ae66a33196c1f870196065f651/hack/kind-fix-networking.sh#L17-L23)

## Expected Behavior
You can connect to the cluster after it has restarted.

## Steps to Reproduce the Bug
Create an unmanaged cluster with default config. Reboot docker. Try to connect again. 

## Environment Details

* Build version (`tanzu version`): 0.9.1 and 0.10.0
* Deployment (Managed/Standalone cluster): managed and unmanaged
* Infrastructure Provider (Docker/AWS/Azure/vSphere): Docker 
* Operating System (client): Not sure. It fails at least on Mac.

tnqn:

When using OVS userspace (netdev) datapath, there was a known checksum
issue causing packets being dropped. The issue was prevented by
patching the network after deploying antrea. However, restarting docker
or machine recreates the docker network, then users wouldn't be able to
connect to the cluster until the network is patched again.

The Linux VM used for Docker Desktop on macOS now includes the
openvswitch kernel module. This patch switches OVS datapath to kernel
(system) to get rid of the checksum issue and the requirement of
patching the network, clusters will continue working after restarting
docker or machine.

```release-note
Switch to OVS kernel datapath for antrea cni to fix the issue that clusters became unreachable after docker restart
```

Fixes: #3058

Create an unmanaged-cluster with antrea as the cni
Restart docker service
Make sure the cluster is still reachable and conformance test pass

I noticed unmanaged-clusters doesn't consume "addons/packages/antrea" but an antrea package in TKr, and "addons/packages/antrea"  is using upstream antrea image "antrea/antrea-ubuntu:v1.2.3" but its yaml has an init container "antrea-agent-tweaker" which uses a command only present in tkg specific image "tkg/antrea-standard-debian:v1.2.3_vmware.4". So the package itself cannot work directly. It should either remove the init container "antrea-agent-tweaker" from the yaml if using upstream antrea image is desired, or keep the init container but switches to the tkg specific image "tkg/antrea-standard-debian:v1.2.3_vmware.4".

The PR shouldn't be merged before TKr release is updated as removing "patchForAntrea" requires antrea's yaml is using kernel datapath.

### DIFF
--- a/addons/packages/antrea/0.11.3/bundle/config/overlay/antrea_overlay.yaml
+++ b/addons/packages/antrea/0.11.3/bundle/config/overlay/antrea_overlay.yaml
@@ -35,9 +35,7 @@ featureGates:
 #! - netdev
 #! 'system' is the default value and corresponds to the kernel datapath. Use 'netdev' to run
 #! OVS in userspace mode. Userspace mode requires the tun device driver to be available.
-#@ if values.infraProvider == "docker":
-ovsDatapathType: netdev
-#@ end
+#!ovsDatapathType: system
 
 #! Name of the interface antrea-agent will create and use for host <--> pod communication.
 #! Make sure it doesn't conflict with your existing interfaces.
@@ -179,34 +177,10 @@ metadata:
 spec:
   template:
     spec:
-      containers:
-        #@ if values.infraProvider == "docker":
-        #@overlay/match by=overlay.subset({"name":"antrea-ovs"})
-        - command:
-          #@overlay/match by=overlay.subset("start_ovs")
-          #@overlay/replace
-            - start_ovs_netdev
-
-        #@overlay/match by=overlay.subset({"name":"antrea-ovs"})
-        - volumeMounts:
-          #@overlay/append
-            - mountPath: /dev/net/tun
-              name: dev-tun
-        #@ end
-
       initContainers:
-        #@ if values.infraProvider == "docker":
-        #@overlay/match by=overlay.subset({"name":"install-cni"})
-        - command:
-          #@overlay/match by=overlay.subset("install_cni")
-          #@overlay/replace
-            - install_cni_kind
-        #@ end
-      volumes:
-      #@ if values.infraProvider == "docker":
-      #@overlay/append
-        - hostPath:
-            path: /dev/net/tun
-            type: CharDevice
-          name: dev-tun
-      #@ end
+      #@overlay/match by=overlay.subset({"name": "install-cni"})
+      - name: install-cni
+        volumeMounts:
+        #@overlay/match by=overlay.subset({"name": "host-depmod"})
+        #@overlay/remove
+        - name host-depmod
--- a/addons/packages/antrea/0.13.3/bundle/config/overlay/antrea_overlay.yaml
+++ b/addons/packages/antrea/0.13.3/bundle/config/overlay/antrea_overlay.yaml
@@ -43,9 +43,7 @@ featureGates:
 #! - netdev
 #! 'system' is the default value and corresponds to the kernel datapath. Use 'netdev' to run
 #! OVS in userspace mode. Userspace mode requires the tun device driver to be available.
-#@ if values.infraProvider == "docker":
-ovsDatapathType: netdev
-#@ end
+#!ovsDatapathType: system
 
 #! Name of the interface antrea-agent will create and use for host <--> pod communication.
 #! Make sure it doesn't conflict with your existing interfaces.
@@ -223,37 +221,3 @@ metadata:
   #@overlay/match missing_ok=True
   annotations:
     kapp.k14s.io/disable-default-label-scoping-rules: ""
-spec:
-  template:
-    spec:
-      containers:
-        #@ if values.infraProvider == "docker":
-        #@overlay/match by=overlay.subset({"name":"antrea-ovs"})
-        - command:
-          #@overlay/match by=overlay.subset("start_ovs")
-          #@overlay/replace
-            - start_ovs_netdev
-
-        #@overlay/match by=overlay.subset({"name":"antrea-ovs"})
-        - volumeMounts:
-          #@overlay/append
-            - mountPath: /dev/net/tun
-              name: dev-tun
-        #@ end
-
-      initContainers:
-        #@ if values.infraProvider == "docker":
-        #@overlay/match by=overlay.subset({"name":"install-cni"})
-        - command:
-          #@overlay/match by=overlay.subset("install_cni")
-          #@overlay/replace
-            - install_cni_kind
-        #@ end
-      volumes:
-      #@ if values.infraProvider == "docker":
-      #@overlay/append
-        - hostPath:
-            path: /dev/net/tun
-            type: CharDevice
-          name: dev-tun
-      #@ end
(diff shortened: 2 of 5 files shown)


tnqn:
If this PR is merged, will antrea-0.11.3 and antrea-0.13.3 packages be used by the new unmanaged-cluster code? If yes, we may need to update these packages as well, otherwise they wouldn't work without the execution of "patchForAntrea".

tnqn:
cc @jpmcb

jpmcb:
> I noticed unmanaged-clusters doesn't consume "addons/packages/antrea" but an antrea package in TKr, and "addons/packages/antrea" is using upstream antrea image "antrea/antrea-ubuntu:v1.2.3" but its yaml has an init container "antrea-agent-tweaker" which uses a command only present in tkg specific image "tkg/antrea-standard-debian:v1.2.3_vmware.4".

Right - we're moving to using the correct packages. That will be part of #3550 

> will antrea-0.11.3 and antrea-0.13.3 packages be used by the new unmanaged-cluster code? If yes, we may need to update these packages as well

I'd say we should also update those packages as well since there may be users who will want to deploy those versions of `antrea`, but ultimately, that's up to the package authors/owners to accept those

hangyan:
I have done some tests for both linux and macos, hope it helps.

ENV:
  tanzu version: v0.10.1
  antrea: v1.5.0 (standard yaml)

Docker Restart: recoverd
Conformance Test: pass
Networkpolicy: pass on linux (macos seems have some setup issues, still debugging it)

tnqn:
> Right - we're moving to using the correct packages. That will be part of #3550

@jpmcb I see #3550 has been merged, does it mean unmanaged-cluster now consumes antrea package in this repo? or we still need a new TKr release?

> I'd say we should also update those packages as well since there may be users who will want to deploy those versions of `antrea`, but ultimately, that's up to the package authors/owners to accept those

Got it, will update them together. And perhaps I should keep `patchAntrea` method for a while so that unmanaged-cluster plugin can work with both old and new antrea packages. As long as we replace the yaml, calling `patchAntrea` is not necessary, but harmless.

hangyan:
> Maybe. I will update the vm and test the networkpolicy case again.

jpmcb:
> Or we still need a new TKr release?

Yes; we still need this to be apart of `unmanaged-clusters` TKr which should get automatically consumed and generated for our v0.12.0 release using this package

hangyan:
> > > result: Docker Restart: recoverd Conformance Test: pass Networkpolicy Test: pass on linux, macos has one failure `[sig-network] NetworkPolicy [Feature:SCTPConnectivity][LinuxOnly][Disruptive] NetworkPolicy between server and client using SCTP should support a 'default-deny' policy [Feature:NetworkPolicy]`
> > 
> > 
> > Thanks @hangyan for verifying it. Is the failure related to kernel sctp module not available on macos?
> 
> Possible. I will update the vm and test the networkpolicy case again.



> > > result: Docker Restart: recoverd Conformance Test: pass Networkpolicy Test: pass on linux, macos has one failure `[sig-network] NetworkPolicy [Feature:SCTPConnectivity][LinuxOnly][Disruptive] NetworkPolicy between server and client using SCTP should support a 'default-deny' policy [Feature:NetworkPolicy]`
> > 
> > 
> > Thanks @hangyan for verifying it. Is the failure related to kernel sctp module not available on macos?
> 
> Possible. I will update the vm and test the networkpolicy case again.

After enable sctp kernel module on linuxkit, the test passed. Thanks @tnqn cc @jpmcb

hangyan:
@jpmcb Can this PR be merged or need something else to be done?

jpmcb:
Last I looked at this, we still need:

- [ ] Updates for this for other harbor packages
- [ ] a re-review (we've been busy cutting the v0.11.0 release)

Thanks for your patience!

joshrosso:
> Last I looked at this, we still need:
> 
>     * [ ]  Updates for this for other harbor packages
> 
>     * [ ]  a re-review (we've been busy cutting the v0.11.0 release)
> 
> 
> Thanks for your patience!

@jpmcb did you mean updates to other `antrea` packages?

jpmcb:
Yes - for the other antera versions, not harbor

joshrosso:
Validated that this does check out when apply the diff to `0.13.3`.

https://user-images.githubusercontent.com/6200057/161327330-66ef3b50-9c7f-46d3-b8db-cb17b502ddf4.mp4

Next is to try with arm64.

joshrosso:
@tnqn it appears `antrea` cannot deploy on arm64 Macs (using Docker Desktop). The CNI install fails due to missing the openvswitch module.

```
$ kubectl logs -n kube-system antrea-agent-pq6jr -c install-cni -p
modprobe: ERROR: could not insert 'openvswitch': Function not implemented
Failed to load the OVS kernel module from the container, try running 'modprobe openvswitch' on your Nodes
```

Here's a video detailing the troubleshooting:

https://user-images.githubusercontent.com/6200057/161329255-b3f9ed4b-6826-4afe-9b08-2d7e3ec1529f.mov

Is this something ya'll foresee as resolvable? Or?

hangyan:
@joshrosso Can you share more details of your setup? We cannot reproduce this issue using arm64 mac.

tnqn:
@joshrosso thanks for verifying it. could you share the docker desktop version on arm64 Mac? We didn't see it using arm64 mac and suspect it's because the docker desktop version is old, and want to evaluate whether there is a solution based on the version information.

BTW, @hangyan has appended commits to this PR to apply same change to antrea 0.11.3 and 0.13.3 packages.

joshrosso:
@vmware-tanzu/pkg-antrea-owners when ya'll have a chance, can you take a look at this PR.

The antrea team are suggesting these package updates. But I need ya'll to determine:

* Do we need a new semver of the package to introduce this patch (my guess is yes)
* Do these changes impact the package in a larger sense? Meaning managed-clusters or downstream products like Tanzu Kubernetes Grid (TKG).

In summary: wondering if we can get these patches into our official `atnrea` package.

hangyan:
@joshrosso @jpmcb The error on arm64 mac was caused by the lack of arm64 image of 0.13.3 antrea (well we have an arm64 image ,just not in the same tag). In that case,  the docker desktop on m1 will trying to run the amd64 image with qemu, which will cause unexpected error for our program.  Since both 0.11.3 and 0.13.3 both are pretty old versions, and it hasn't been maintained by antrea team for a long time. So is it possible to remove the support for these versions in tce ? Or maybe only for arm platform?

@tnqn

hangyan:
maybe related to #4213

hangyan:
@joshrosso @jpmcb  Any thoughts? I see there is also an another PR to bring antrea 1.5.2 to tce packages.

hangyan:
For the latest versions of antrea (including the just merged 1.5.2), it already contains the arm64 image. 

>

hangyan:
> Trying to get a handle on where things are here:
> 
>     * [ ]  Need arm64 images for latest antrea version
> 
>     * [ ]  Need to update PR and resolve merge conflicts
> 
> 
> Anything else?

For the latest antrea packages ( 1.2.3, including the just meged 1.5.2), arm64 image is already included.  I think we can switch the default antrea version to the above versions and remove the old versions ( in the future)?

tnqn:
I have resolved conflicts and applied same change to v1.5.2 package.

hangyan (approved):

hangyan:
@joshrosso Can we merge this PR now?

PR #3579, opened 2022-03-17
cartographer: add smoke test

cirocosta:

similar to other components that already have smoke tests, include one
for the cartographer package.

the test is limited to making sure that Cartographer has been properly
installed by submitting a couple of objects to kubernetes and then
waiting for the result of the supplychain to be materialized:

- supplychain w/ template for stamping out a configmap is submitted
- rbac-related objects are submitted so that the workload in the test
  namespace can have the controller creating what it needs to create
  (the configmap)
- workload is submitted; the controller then matches it against the
  supplychain, resulting in a configmap named `workload-<workload-name>`,
  which we wait for as a way of checking if things went well.

```release-note
Add smoke tests for Cartographer
```

1. created a cluster where the package provided by this repository (including 
  cert-manager, a dependency or Cartographer) are installed
2. ran the smoke test added in here
3. observed success

```console
$ make cartographer
./packages/cartographer/0.2.2/cartographer-test.sh
  cert-manager              cert-manager.community.tanzu.vmware.com       1.6.1            Reconcile succeeded
  cartographer              cartographer.community.tanzu.vmware.com       0.2.2            Reconcile succeeded
namespace/cartographer-14866 created
serviceaccount/test-basic created
role.rbac.authorization.k8s.io/test-basic created
rolebinding.rbac.authorization.k8s.io/test-basic created
clustertemplate.carto.run/test-basic unchanged
clustersupplychain.carto.run/test-basic unchanged
workload.carto.run/test-basic created

sleeping 1s to wait for configmap
Error from server (NotFound): configmaps "workload-test-basic" not found
sleeping 2s to wait for configmap
NAME                  DATA   AGE
workload-test-basic   0      0s

 Uninstalling package 'cartographer' from namespace 'default'
 Getting package install for 'cartographer'
 Deleting package install 'cartographer' from namespace 'default'
 'PackageInstall' resource deletion status: Deleting
 Deleting admin role 'cartographer-default-cluster-role'
 Deleting role binding 'cartographer-default-cluster-rolebinding'
 Deleting service account 'cartographer-default-sa'
Uninstalled package 'cartographer' from namespace 'default'

 Uninstalling package 'cert-manager' from namespace 'default'
 Getting package install for 'cert-manager'
 Deleting package install 'cert-manager' from namespace 'default'
 'PackageInstall' resource deletion status: Deleting
 Deleting admin role 'cert-manager-default-cluster-role'
 Deleting role binding 'cert-manager-default-cluster-rolebinding'
 Deleting service account 'cert-manager-default-sa'
Uninstalled package 'cert-manager' from namespace 'default'

namespace "cartographer-14866" deleted

cartographer passed
```

### DIFF
--- a/test/smoke/packages/cartographer/0.2.2/cartographer-test.sh
+++ b/test/smoke/packages/cartographer/0.2.2/cartographer-test.sh
@@ -35,7 +35,7 @@ for sleep_duration in {1..10}; do
     echo "sleeping ${sleep_duration}s to wait for configmap"
     sleep "$sleep_duration"
 
-    kubectl get configmap workload-test-basic && {
+    kubectl get configmap -n ${NAMESPACE} workload-test-basic && {
         packageCleanup cartographer cert-manager
         namespaceCleanup ${NAMESPACE}
         successMessage cartographer
--- a/test/smoke/packages/cartographer/0.2.2/cartographer-test.sh
+++ b/test/smoke/packages/cartographer/0.2.2/cartographer-test.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+TCE_REPO_PATH="$(git rev-parse --show-toplevel)"
+# shellcheck source=test/smoke/packages/utils/smoke-tests-utils.sh
+source "${TCE_REPO_PATH}/test/smoke/packages/utils/smoke-tests-utils.sh"
+
+MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Checking package is installed or not
+tanzu package installed list | grep "cert-manager.community.tanzu.vmware.com" || {
+    version=$(tanzu package available list cert-manager.community.tanzu.vmware.com | tail -n 1 | awk '{print $2}')
+    tanzu package install cert-manager --package-name cert-manager.community.tanzu.vmware.com --version "${version}"
+}
+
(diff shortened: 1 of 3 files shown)


cirocosta:
Hey @aman556, thanks for checking! indeed, the test did need a little tweak - the `k get configmap` should be happening against the same namespace as the workload. fixed in 7bdeaf5ea871870e05b24a121101b91f8dc4a960

please let me know how it goes 😁  thank you!

test/smoke/packages/cartographer/0.2.2/cartographer-test.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+
+NAMESPACE_SUFFIX=${RANDOM}
+NAMESPACE="cartographer-${NAMESPACE_SUFFIX}"
+kubectl create ns ${NAMESPACE}
+
+kubectl apply -n ${NAMESPACE} --filename "${MY_DIR}"/testdata.yaml
+
+for sleep_duration in {1..10}; do
```
aman556, line 34: can we check this something like 
`kubectl wait --for=condition=Ready pod volume-test -n ${NAMESPACE}`
i don’t know kubectl wait work for config-map or not just a ask if possible as this will be  fail and print appropriate logs too on failing.

cirocosta: that's a good suggestion but indeed, I don't think there's some command we could use (like `wait` as you suggested or something else similar to `rollout`) though

in this case, we're essentially just waiting for the object to exist (if at some point it exists, that means that cartographer is in good shape as it was able to create it according to the supply chain), so even if configmaps had conditions, `k wait` would still fail early as the object wouldn't initially exist

aman556: Makes Sense Thanks @cirocosta approving and merging it.

aman556 (approved):

PR #3626, opened 2022-03-21
Update package readme script for new doc structure

stmcginnis:
The "latest" folder has been removed from our docs structure to support
our current method for versioning docs. The script to update package
READMEs into the published docs was missed during this update, so it was
still trying to use the deeper folder structure and failing. This
updates the script to remove the "latest" folder.


davidvonthenen (approved):
That's how I remember it working. I believe this is all that's needed.

seemiller (approved):

PR #3637, opened 2022-03-21
Fix for the docs TOC view and URLs

jonasrosland:

The docs TOC isn't showing up properly with the new versioning changes, and the URLs seem to go to `/latest`. This PR fixes that so that is links to the correct version.

Fixes: #


kcoriordan (approved):

stmcginnis (approved):
Thanks!

PR #3643, opened 2022-03-21
Handle include links for versioned docs

stmcginnis:
With the switch to versioned docs, our include shortcodes were broken
and that content would not get added to the page. This was due to the
path changing from `/docs/file` to `docs/$version/file`.

This updates the shortcode to automatically adjust the path for the
given version. It also addressed a couple relative references for things
like images.

### DIFF
--- a/docs/site/.gitignore
+++ b/docs/site/.gitignore
@@ -1,2 +1,3 @@
 /public/
 /resources/_gen/
+.hugo_build.lock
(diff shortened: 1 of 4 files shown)


kcoriordan (approved):

kcoriordan:
thanks Sean LGTM

jonasrosland (approved):
LGTM! Thanks Sean!

PR #3658, opened 2022-03-22
Docs: Update docsearch facetfilters version variable

### LINKED ISSUE #3653: Documentation search is broken with new versioning mode
I believe due to the new versioning updates, our search functionality has broken.

As you can see in the URL bar after the search, the path resolves as `$VERSION/latest/$RESOURCE` when it should be `$VERSION/$RESOURCE`.

image

image

garrying:
Signed-off-by: Garry Ing <ging@vmware.com>

Updates the version variable in the `docsearch` configuration to use the relative version the user is on. Algolia indexes have been updated to reflect docs versioning.

Fixes: #3653

- `hugo server`
- Navigate to `/docs/v0.10/`


stmcginnis (approved):
Seems reasonable.

PR #3642, opened 2022-03-21
Removes mention of cgroups v1 being a requirement in docs

jpmcb:
## What this PR does / why we need it
Removes mention of cgroups v1 from docs (which has been fixed in upcoming `v0.11.0`)

## Which issue(s) this PR fixes
Fixes: #3516

## Describe testing done for PR
`hugo serve` and `make lint`

PR #3670, opened 2022-03-23
Cherry-pick: Removes mention of cgroups v1 being a requirement in docs (#3642)

jpmcb:
Cherry-picks removing mention of `cgroups` in docs for v0.11.0 releas

`hugo serve` & `make check`


stmcginnis (approved):
Looks good. No checks were run, but I'm assuming that was due to the current issues GitHub is experiencing. This ran on `main` and everything was good. so assuming checks would pass here as well.

PR #3675, opened 2022-03-23
homepage and packages updates

michow:

Fixes: #


docs/site/themes/template/layouts/partials/use-cases.html
```diff
@@ -12,7 +12,7 @@ <h3>Learn</h3>
                 </div>
             </div>
         </a>
```
jonasrosland, line 14: This section talks about a hands-on lab, not the blog. So if this should point to the blog then the copy is incorrect.

docs/site/themes/template/layouts/partials/use-cases.html
```diff
@@ -23,7 +23,7 @@ <h3>Evaluate</h3>
...
-        <a href="/docs/">
+        <a href="/resources">
```
jonasrosland, line 26: Why are we changing this from docs to resources?

jonasrosland (changes requested):
Minor copy and link changes that needs clarification

docs/site/themes/template/layouts/partials/use-cases.html
jkotzen: we currently link to docs from 2 of the three use case boxes here, and wanted to give the visitors more variety in their exploration path.  Linking this 'use' use case to the resources tab is a way to do that. More importantly, we are starting to build a good content library that will (hopefully soon!) show up in our resources page, where it will be organized by persona and use case - so in the mid to longer term that page should be the ideal place for people to go to land if they'd like help re: how to use TCE to solve their problems.

docs/site/themes/template/layouts/partials/use-cases.html
jkotzen: this is talking about TCE as an environment that people can use to learn in a hands-on way.  It is not referring to a HOL

jonasrosland: The copy still doesn't refer to the blog though, and the blogs don't focus on hands-on learning, so I'm unclear how the copy fits here.

docs/site/themes/template/layouts/partials/use-cases.html
jonasrosland: I think "Learn" fits better for /resources, and "Use" can still go to /docs.

jonasrosland (approved):
LGTM!

microwavables:
Also LGTM!

PR #3684, opened 2022-03-23
Fix mdlint error in support-matrix doc

stmcginnis:

The support matrix documentation was recently updates to remove
references to cgroupv1 limitations. For some reason the mdlint check did
not run on that change, and there was a minor issue with its formatting.

This removes an extra blank line from the end of the file to make the
linter happy.


kcoriordan (approved):

jpmcb (approved):
Thanks! Good catch!!

PR #3734, opened 2022-03-25
Bump multus-cni version to 3.8.0

wjun:
Bump multus-cni version to the latest one.

Passed e2e test on a repository bundle with multus-cni 3.8.0 package.


12345lcr (dismissed):
LGTM

addons/packages/multus-cni/3.8.0/test/README.md
```diff
@@ -0,0 +1,9 @@
+# MULTUS-CNI E2E TESTS
+
+## USAGE
```
seemiller, line 3: I was able to run the test and it passed. However, there was a bit of trial and error to get the test to pass. I first needed to install a package repository that contained the 3.8.0 package so that it could be installed by the test. Second, I had to create an empty `test/e2e/multus-cni-values.yaml` file since the package install in the test is looking for that.

Could you update the USAGE section with these and/or other additional steps that a user would need to take in order to have a successful test run.

addons/packages/multus-cni/3.8.0/README.md
```diff
@@ -0,0 +1,106 @@
...
+         "routes": [
+           { "dst": "0.0.0.0/0" }
+         ],
+         "gateway": "192.168.1.1"
+        }
+      }'
+    EOF
+    ```
+
+1. Deploy a sample pod using the network-attachment-defs defined above by the following  lines to the pod spec:
```
seemiller, line 78:
```suggestion
1. Deploy a sample pod using the network-attachment-defs defined above by the following lines to the pod spec:
```
Remove extra space.

seemiller (changes requested):
@wjun Overall looks good! Once you expand on the instructions for running the tests we'll get it merged!

addons/packages/multus-cni/3.8.0/README.md
wjun: Done

addons/packages/multus-cni/3.8.0/test/README.md
wjun: Good point, @seemiller . I added a prerequisites section into test readme. If you run make e2e-test from test directory, the test cases will find values file from e2e/multihomed-testfiles/multus-cni-values.yaml

seemiller (approved):

PR #3770, opened 2022-03-29
Add GCP bucket init back in

davidvonthenen:

This was an oops. Need to add back in for the release.


stmcginnis (approved):
Looks right to me, but I am not able to actually run and verify it. Would be good to have someone run through, but I think it's pretty safe.

PR #3772, opened 2022-03-29
Fix GCP bucket init on main

davidvonthenen:

Fix GCP bucket init on main for 0.11.0 release


stmcginnis (approved):
We should probably merge to main first, then backport to `release-v0.11`. But in the interest of time, looks fine and reasonable to get done in parallel.

PR #3806, opened 2022-03-30
GitHub Actions: Use Unique Runners Per `RunAttempt`

davidvonthenen:

After switching from the first version of the GH Actions using self-hosted runners in favor of the WebHook method for creating runners (v2), we regressed a feature that gave us unique runners per `RunAttempt` where we reused the same runner if a new `RunAttempt` was received before the self-hosted runner was deleted. This is a common occurrence especially when doing multiple subsequent pushes to your branch on the same PR. This PR now will give us a unique runner for each PR's RunAttempt.

Made a slight change to trigger the E2Es from the daily build.

NA

As best as I could without actually throwing this into production. I create a test binary that output in debug logs what I was expecting to happen without actually triggering the new webhook. Unfortunately, this PR needs to be merged at the same time the webhook service is replaced.

NA

### DIFF
--- a/.github/workflows/build-daily.yaml
+++ b/.github/workflows/build-daily.yaml
@@ -19,7 +19,10 @@ jobs:
         id: start-ec2-runner
         shell: bash
         run: |
-          INSTANCE_NAME="id-${GITHUB_RUN_ID}-${GITHUB_RUN_NUMBER}"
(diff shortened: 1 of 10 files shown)


stmcginnis (approved):
Let's give it a shot!

PR #3850, opened 2022-03-31
Flush Go Cache Each Month

davidvonthenen:

Currently it seems like we keep the Go Cache around forever. This leads to very long restore times when trying to load the Go Cache. A lot of this stuff might be very stale and we don't want to carry that stuff around. This is a compromise so that we flush out and store the go cache each month. Picking that timeframe because we would like to do monthly releases and it would seem to cycle out old dependencies each month.

NA

NA

NA

### DIFF
--- a/.github/workflows/check-lint.yaml
+++ b/.github/workflows/check-lint.yaml
@@ -26,11 +26,19 @@ jobs:
           go-version: "1.17"
         id: go
 
-      - name: Cache linting
-        uses: actions/cache@v2
+      - name: Get Date
+        id: get-date
+        shell: bash
+        run: |
+          echo "::set-output name=date::$(date -u "+%Y-%m")"
+
+      - name: Restore Lint Cache
+        uses: actions/cache@v3
         with:
           path: ${{ runner.temp }}/lint_cache
-          key: ${{ runner.os }}-lint-cache
+          key: ${{ runner.os }}-lint-cache-${{ steps.get-date.outputs.date }}
+          restore-keys: |
+            ${{ runner.os }}-lint-cache-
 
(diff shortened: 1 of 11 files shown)


.github/workflows/check-lint.yaml
```diff
@@ -26,11 +26,19 @@ jobs:
...
 
-      - name: Cache linting
-        uses: actions/cache@v2
+      - name: Get Date
+        id: get-date
+        shell: bash
+        run: |
+          echo "::set-output name=date::$(date -u "+%Y-%m")"
+
+      - name: Restore Go Cache
```
stmcginnis, line 35: nit: This one isn't a go cache.

stmcginnis (dismissed):
Looks good. If I understand right, the first time each action runs at the start of a new month, it will try to restore the cache but not find it. So a new cache will be created and subsequent runs for the rest of the month will be able to take advantage of the cached data. Then things get reset again the next month, so we don't end up accumulating more and more data in the cache.

dvonthenen:
> Looks good. If I understand right, the first time each action runs at the start of a new month, it will try to restore the cache but not find it. So a new cache will be created and subsequent runs for the rest of the month will be able to take advantage of the cached data. Then things get reset again the next month, so we don't end up accumulating more and more data in the cache.

That's absolutely right. Because the cache key is tied to "YYYY-MM" in the key.

.github/workflows/check-lint.yaml
dvonthenen: Ooops. Copy and paste. Will fix

stmcginnis (approved):

PR #3852, opened 2022-03-31
Remove schema files for older CPI package version 1.20.x 1.21.x

HanFa:

This PR removes the `schema.yaml` file for older versions of CPI package. 

Current issue: when creating a workload cluster with k8s below 1.22, the package will fail to be reconciled by kapp-controller due to the version misalignment with the data values secret (which is of version 1.22.x). Removing the schema file for those affected CPI package can workaround these issue.

Fixes: #

Downstream pipelines (ongoing)


blc1996 (approved):
LGTM. Need discussion on how we are going to handle these schema file going forward. cc @vijaykatam

seemiller (approved):

lubronzhan (approved):

PR #3922, opened 2022-04-06
Modify harbor 2.3.3 ipFamilies values and unittests

MinerYang:
harbor values `ipFamilies` didn't rendered properly as expected due to different ytt behaviors in TKG cluster .
For example, when user inputs `ipFamilies` with **IPv4 only** , it should rendered like
```
listen 8443 ssl;
```
But in TKG clusters we found it rendered  like below
```
listen 8443 ssl;
listen [::]:8443 ssl;
```
This shouldn't be the correct output for configuration with **IPv4 only**, same problem happened when configure **IPv6 only**.
This issue may caused by different ytt behaviors from TCE cluster to TKG clusters. Ytt template in TKG merged both upstream_lib with downstream data. 
**So we change the original default ipFamilies configuration from `["IPv4", "IPv6"]` to `[]`**, more details as below

Option1 :  **ipFamilies:[]** means configure with both IPv4 and IPv6, **same as Option4**.
Option2 : **ipFamilies:["IPv4"]** means configure with  IPv4 only
Option3 : **ipFamilies:["IPv6"]** means configure with  IPv6 only
Option4 : **ipFamilies:["IPv4", "IPv6"]** means configure with both IPv4 and IPv6, **same as Option1**.

- unittest
`make test`

- manually test
`install harbor package on TCE unmanagement cluster`
`install harbor package on TKG cluster`

This is a fix for tkg clusters. No impact on TCE clusters

### DIFF
--- a/addons/packages/harbor/2.3.3/bundle/config/values.yaml
+++ b/addons/packages/harbor/2.3.3/bundle/config/values.yaml
@@ -246,5 +246,6 @@ metrics:
     path: /metrics
     port: 8001
 
+#! Default [] is equivalent to have both ["IPv4","IPv6"]. Or you can choose one of ["IPv4"] or ["IPv6"]
 network:
-  ipFamilies: ["IPv4","IPv6"]
+  ipFamilies: []
--- a/addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-and-ipv6.yaml
+++ b/addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-and-ipv6.yaml
@@ -1,0 +1,4 @@
+#@data/values
+---
+network:
+  ipFamilies: ["IPv4", "IPv6"]
--- a/addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-only.yaml
+++ b/addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-only.yaml
@@ -1,0 +1,4 @@
+#@data/values
+---
+network:
+  ipFamilies: ["IPv4"]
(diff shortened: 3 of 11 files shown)


heww (approved):

addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-and-ipv6.yaml
```diff
@@ -0,0 +1,15 @@
+#@data/values
+---
+harborAdminPassword: Harbor12345
```
zyyw, line 3: line 3 to line 13 can be removed

addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv4-only.yaml
```diff
@@ -0,0 +1,15 @@
+#@data/values
+---
+harborAdminPassword: Harbor12345
```
zyyw, line 3: line 3 to line 13 can be removed

addons/packages/harbor/2.3.3/test/unittest/fixtures/values/ipv6-only.yaml
```diff
@@ -0,0 +1,15 @@
+#@data/values
+---
+harborAdminPassword: Harbor12345
```
zyyw, line 3: line 3 to line 13 can be removed

addons/packages/harbor/2.3.3/bundle/config/values.yaml
```diff
@@ -247,4 +247,4 @@ metrics:
     port: 8001
 
 network:
-  ipFamilies: ["IPv4","IPv6"]
+  ipFamilies: []
```
zyyw, line 250: add a comment before line 250. 
```
description: The array of network ipFamilies. Default [] is equivalent to have both ["IPv4","IPv6"]. Or you can choose one of ["IPv4"] or ["IPv6"]
```

zyyw (approved):
lgtm

heww (approved):

PR #3959, opened 2022-04-11
Add utility for parallel lint execution

stmcginnis:

This adds a small utility to give us more control over running multiple
lint checks in parallel, though it is flexible enough that it could be
used for any case where we have multiple make targets that we would like
to run in parallel.

Ran `make check` and `PRINT_FAILURES=1 make check` to verify proper operation.

We may want to think about moving some things like `imagelint` out of the default checks run as that's a very long running check.

### DIFF
--- a/hack/check/makerunner.go
+++ b/hack/check/makerunner.go
@@ -1,0 +1,136 @@
+// Copyright 2022 VMware Tanzu Community Edition contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"sync"
+	"time"
+)
+
+const (
+	Running  = string('\U0001F3C3')
+	Complete = string('\U0001F600')
+	Failed   = string('\U0001F4A5')
+)
+
+// Set a PRINT_FAILURES env variable to 1 to enable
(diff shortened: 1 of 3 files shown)


davidvonthenen (approved):
This is a great idea... I've been slowly trying to move our bash stuff to go and actually have go tests for it and any effort to go in that direction is thumbs up for me.

hack/check/makerunner.go
```diff
@@ -0,0 +1,136 @@
...
+			} else {
+				tasks[i].status = Complete
+				defer os.Remove(f.Name())
+			}
+
+			<-guard
+		}(tasks, target, i)
+	}
+
+	time.Sleep(500 * time.Millisecond)
```
jpmcb, line 97: Why the sleep here? 😴 can we go right into checking the status of the finished jobs?

jpmcb (approved):
Wow - this is amazing!!! 🏃🏼

hack/check/makerunner.go
stmcginnis: I saw a few cases where the output wouldn't get refreshed right. It looked like it was due to trying to refresh too quickly after kicking off the tasks. Since even a very simple target is going to take in the order of seconds, I figured it was safe to throw in a half second sleep before going into the update loop since we will need to wait anyway.

PR #3978, opened 2022-04-11
Add `make check-full` to make `check` more useful

stmcginnis:

The current `make check` runs all of our linters, which is great, but
they take so long to run that it almost guarantees no one will actually
run this locally more than once. We really do want folks running the
tests before proposing PRs and then needing to address issues after the
fact. To try to make it faster and easier, and make it more likely to be
used, this separates out the checks into:

- `make check` - runs a subset of all linters for the most common things
- `make check-full` - runs all the linter checks for full coverage

Our CI will continue to run the full suite of tests, but by default a
local user can just run the tests that are most likley to uncover any
concerns in their changes.

This also changes the async lint run to stream the output immediately to
the log file. This is useful to be able to tail a log while a linter is
running to get some visiblity into what is happening.

### DIFF
--- a/.github/workflows/check-all.yaml
+++ b/.github/workflows/check-all.yaml
@@ -52,7 +52,7 @@ jobs:
         shell: bash
         run: |
           make ensure-deps
-          PRINT_FAILURES=1 make check
+          PRINT_FAILURES=1 make check-full
       - name: Get dependencies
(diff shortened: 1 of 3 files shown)


davidvonthenen (approved):
woot woot!

PR #3981, opened 2022-04-11
Resolve minikube image via kubernetes version

### LINKED ISSUE #3982: unmanaged-cluster provider minikube does not support non-docker drivers
## Bug Report

Similar to `kind` the `minikube` unmanaged-cluster provider attempts to pre-pull the host image based on the TKr. However, the user's driver choice determines which image is needed (e.g. kvm2, vmware, docker, etc).

For now we should move to a model where we tell `minikube` the desired Kubernetes version and let it select the image based on the driver.

In the future, we should move all minikube images for each driver into our TKr so they can be referenced accordingly.

joshrosso:

This commit moves (back) to minikube resolution based on the kubernetes
version specified in the TKr. This is needed because there are unique
images for each driver type (e.g. kvm2 and docker).

Until we setup our TKrs to express many minikube images, we will stick
to this model and leverage default images resolved by minikube.

* Fixes: https://github.com/vmware-tanzu/community-edition/issues/3982

**:speaker: audio included**

https://user-images.githubusercontent.com/6200057/162842281-b774a074-72ee-4dd5-a912-5ec793346652.mp4

Review the video above, review the code changes, make suggestions.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/minikube.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/minikube.go
@@ -56,8 +56,6 @@ func (mkcm *MinikubeClusterManager) Create(c *config.UnmanagedClusterConfig) (*K
 	// Base start command arguments
 	args := []string{
 		"start",
-		"--driver",
-		mkcm.driver,
 		"--profile",
 		c.ClusterName,
 	}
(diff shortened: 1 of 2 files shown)


joshrosso:
@jpmcb please review when you have a moment. Thank you!

jpmcb (approved):
Nice! Looks good, works good 👍🏼

PR #3987, opened 2022-04-11
Ensure minikube cluster isn't nil (missing) before resolving fields

### LINKED ISSUE #3986: Potential panic in minikube provider's status if minikube cluster is deleted outside of `tanzu`
When we grab minikube status we assume the returned profile will have a value, yet it can be `nil`. In this case we need to return a statusunknown rather than try to introspect the struct.

joshrosso:

This is a minor bug fix that ensures that a missing minikube cluster does not try to resolve its status. In this case, we should report status Unknown.

* Fixes: https://github.com/vmware-tanzu/community-edition/issues/3986

Delete the minikube cluster using `minikube` then run `tanzu uc ls`

See above


joshrosso:
lint issues resolved by: https://github.com/vmware-tanzu/community-edition/pull/3989

jpmcb (approved):
Makes sense!! LGTM 🚀

PR #4024, opened 2022-04-14
Fix GCP upload to use the latest upload provider (auth change)

davidvonthenen:

The way you auth to the GCP buckets has changed and dependabot is going to keep forcing us to look at this so... you have this PR. It requires an explicit auth step and removes all auth tokens from subsequent steps.

More info here:
https://github.com/google-github-actions/upload-cloud-storage#authorization

NA

NA

NA


jpmcb (approved):
Makes sense, makes sense! 👍🏼 🪣

stmcginnis (approved):

PR #4039, opened 2022-04-14
Makes unmanaged cluster create objects idempotent

### LINKED ISSUE #3580: unmanaged-cluster with existing kubeconfig should be idempotent
## Bug Report
When `tanzufying` an existing cluster, the process should be idempotent because the unmanaged-cluster is not even created by TCE, it only installs stuff on top of an existing cluster, so it makes no sense to have fail in the following scenario:

- create a cluster outside TCE with name test
- create a TCE cluster with kubeconfig option (Tanzufy the cluster)
- delete the cluster outside TCE test
- create the TCE cluster with kubeconfig option (Tanzufy your new cluster)

This should work, but currently it fails because we keep track of bootstrap.log and config.yaml files for that cluster in the TCE filesystem are when the cluster is not managed by TCE, so deletion of the cluster will not remove this files.

```
$ kind create cluster --name kind-test
Creating cluster "kind-test" ...
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind-test"
You can now use your cluster with:

kubectl cluster-info --context kind-kind-test

Thanks for using kind! 😊
~
$ k get nodes
NAME                      STATUS     ROLES                  AGE   VERSION
kind-test-control-plane   NotReady   control-plane,master   8s    v1.21.1
~
$ k config get-contexts
CURRENT   NAME                            CLUSTER          AUTHINFO            NAMESPACE
*         kind-kind-test                  kind-kind-test   kind-kind-test
~
$ ~/bin/tanzu uc create -e ~/.kube/config --cni=none

📁 Created cluster directory

🔧 Resolving Tanzu Kubernetes Release (TKR)
   projects.registry.vmware.com/tce/tkr:v0.17.0
   TKR exists at /home/smcginnis/.config/tanzu/tkg/unmanaged/bom/projects.registry.vmware.com_tce_tkr_v0.17.0
   Rendered Config: /home/smcginnis/.config/tanzu/tkg/unmanaged/kind-kind-test/config.yaml
   Bootstrap Logs: /home/smcginnis/.config/tanzu/tkg/unmanaged/kind-kind-test/bootstrap.log

🔧 Processing Tanzu Kubernetes Release

🎨 Selected base image
   projects.registry.vmware.com/tce/kind:v1.22.4

📦 Selected core package repository
   projects.registry.vmware.com/tce/repo-10:0.10.0

📦 Selected additional package repositories
   projects.registry.vmware.com/tce/main:v0.11.0

📦 Selected kapp-controller image bundle
   projects.registry.vmware.com/tce/kapp-controller-multi-pkg:v0.30.1

🚀 Using existing cluster
   Warning: Components installed using this method will need to be manually removed.

   To troubleshoot, use:

   kubectl ${COMMAND} --kubeconfig /home/smcginnis/.kube/config

📧 Installing kapp-controller
   kapp-controller status: Running

📧 Installing package repositories
   Core package repo status: Reconcile succeeded

🌐 Installing CNI
   No CNI installed: CNI was set to none.

✅ Cluster created

🎮 kubectl context set to kind-kind-test

View available packages:
   tanzu package available list
View running pods:
   kubectl get po -A
Delete this cluster:
   tanzu unmanaged delete kind-kind-test
~
$ tree ~/.config/tanzu/tkg/
bom/           compatibility/ config.yaml    providers/     .tanzu.lock    unmanaged/
~
$ tree ~/.config/tanzu/tkg/unmanaged/
/home/smcginnis/.config/tanzu/tkg/unmanaged/
├── bom
│   └── projects.registry.vmware.com_tce_tkr_v0.17.0
└── kind-kind-test
    ├── bootstrap.log
    └── config.yaml

$ tanzu uc create -e ~/.kube/config --cni=none
directory /home/smcginnis/.config/tanzu/tkg/unmanaged/kind-uctest already exists, this cluster must be deleted before proceeding
Error: exit status 2

✖  exit status 2
```

See more context here: https://github.com/vmware-tanzu/community-edition/issues/3524#issuecomment-1070852748 and below in comments

## Environment Details

* Build version (`tanzu version`): 0.11.rc2
* Deployment (Managed/Unmanaged cluster): unmanaged-cluster
* Infrastructure Provider (Docker/AWS/Azure/vSphere): docker
* Operating System (client): mac

jpmcb:
- Makes kapp controller deployment, packagerepos, and packageinstalls
  idempotent. If the object exists on the cluster, won't attempt to
  create it, returns found objects.
- Noop cluster return an error to inform users that delete does not
  clean up cluster from provider

Fixes: #3580

First, create a new kind cluster to use as an "existing" cluster:
```
$ kind create cluster --name test
```

Run tanzu `create` with existing cluster:
```
$ tanzu unmanaged-cluster create --cni=none -e ~/.kube/config
```

Delete files known by unmanaged-cluster for this:
```
$ tanzu unmanaged-cluster delete kind-test

🧪 Deleting cluster: kind-test
   Warning - could not delete cluster through provider. Be sure to manually delete cluster. Error: cluster not deleted. Existing and noop clusters require manual deletion
   Local config files directory deleted: /home/jmcb/.config/tanzu/tkg/unmanaged/kind-test
Failed delete operation. Error: cluster not deleted. Existing and noop clusters require manual deletion
```

And re-attaching works 👍🏼 
```
$ tanzu unmanaged-cluster create --cni=none -e ~/.kube/config
```

Not sure if this is the best approach or if there's a better way to go about this. Hoping to get some feedback from @joshrosso & @stmcginnis on this approach

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cluster/noop.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cluster/noop.go
@@ -40,7 +40,7 @@ func (ncm NoopClusterManager) Get(clusterName string) (*KubernetesCluster, error
 
 // Delete for noop does nothing since these clusters have no provider and are not lifecycled
 func (ncm NoopClusterManager) Delete(c *config.UnmanagedClusterConfig) error {
-	return nil
+	return fmt.Errorf("cluster not deleted. Existing and noop clusters require manual deletion")
 }
 
 // Prepare doesn't perform any preparation steps before cluster creation.
--- a/cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
+++ b/cli/cmd/plugin/unmanaged-cluster/kapp/kapp.go
@@ -215,30 +215,47 @@ func applyObject(k Client, obj runtime.Object) (*unstructured.Unstructured, erro
 
 	nsInterface := uObj[metadataKey].(map[string]interface{})[namespaceKey]
 	var createObj *unstructured.Unstructured
+	var getObj *unstructured.Unstructured
+
+	gvr := schema.GroupVersionResource{
+		Group:    gvk.Group,
+		Version:  gvk.Version,
+		Resource: mapping.Resource.Resource,
+	}
+
 	if nsInterface != nil {
 		ns := nsInterface.(string)
-		createObj, err = k.dynClient.
-			Resource(schema.GroupVersionResource{
-				Group:    gvk.Group,
-				Version:  gvk.Version,
-				Resource: mapping.Resource.Resource,
-			}).Namespace(ns).
-			Create(context.TODO(), objectBody, metav1.CreateOptions{})
+		getObj, err = k.dynClient.Resource(gvr).Namespace(ns).Get(context.TODO(), objectBody.GetName(), metav1.GetOptions{})
 
-		if err != nil {
+		if err != nil && !strings.Contains(err.Error(), "not found") {
 			return nil, err
 		}
-	} else {
-		createObj, err = k.dynClient.
-			Resource(schema.GroupVersionResource{
-				Group:    gvk.Group,
-				Version:  gvk.Version,
-				Resource: mapping.Resource.Resource,
-			}).
-			Create(context.TODO(), objectBody, metav1.CreateOptions{})
 
-		if err != nil {
+		// Object doesn't exist, create it
+		if getObj == nil {
+			createObj, err = k.dynClient.Resource(gvr).Namespace(ns).Create(context.TODO(), objectBody, metav1.CreateOptions{})
+
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return getObj, nil
+		}
+	} else {
+		getObj, err = k.dynClient.Resource(gvr).Get(context.TODO(), objectBody.GetName(), metav1.GetOptions{})
+
+		if err != nil && !strings.Contains(err.Error(), "not found") {
 			return nil, err
+		}
+
+		if getObj == nil {
+			createObj, err = k.dynClient.Resource(gvr).Create(context.TODO(), objectBody, metav1.CreateOptions{})
+
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return getObj, nil
 		}
 	}
 
--- a/cli/cmd/plugin/unmanaged-cluster/packages/packages.go
+++ b/cli/cmd/plugin/unmanaged-cluster/packages/packages.go
@@ -9,6 +9,7 @@ package packages
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	kappapis "github.com/vmware-tanzu/carvel-kapp-controller/pkg/apis/kappctrl/v1alpha1"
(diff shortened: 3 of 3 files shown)


jpmcb:
cc @jorgemoralespou since you had made the original request for this 👀

joshrosso (changes requested):
I'm not so sure on this one.

I get the intent, however I think we need to give deeper thought to what it means to "attach" to an existing cluster and what the longer-term lifecycle management of workloads (such as kapp-controller) should look like.

For example, today, if I have a cluster where I deployed a `kapp-controller`, perhaps from a TKG package through cluster essentials, and UC is pointed at it (perhaps on accident), it will error out:

```
📧 Installing kapp-controller
failed to install kapp-controller, Error: namespaces "tkg-system" already exists
Error: exit status 9
```

I think this **is** the correct behavior.

I'm not saying no to an approach like this long-term, but I think we need better definitions around "attaching" to a cluster and what the expectations are.

I think we should revisit this topic in `v0.13.0`

jpmcb:
> I think we need to give deeper thought to what it means to "attach" to an existing cluster and what the longer-term lifecycle management of workloads (such as kapp-controller) should look like.

> I'm not saying no to an approach like this long-term, but I think we need better definitions around "attaching" to a cluster and what the expectations are.

Agreed - this first stab gives us a good idea of what that behavior might look like, but _it is_ a drifting UX since now with this change, if the objects are there, we skip installing instead of erroring out.

> I think we should revisit this topic in v0.13.0

Sounds good - I'll move #3580 into that project

jorgemoralespou:
>  today, if I have a cluster where I deployed a kapp-controller, perhaps from a TKG package through cluster essentials, and UC is pointed at it (perhaps on accident), it will error out:
   ```
   📧 Installing kapp-controller
   failed to install kapp-controller, Error: namespaces "tkg-system" already exists
   Error: exit status 9
   ```
>I think this is the correct behavior.

I agree. The caveat here is that I would expect some way of "identifying" who did what, so that, if `uc cluster create` was the   way, then it takes the actions to be idempotent, else, can error out saying, `hey, you already have this, installed by some other means, it's up to you to fix that`.

What I was asking was for the `tanzufy a cluster` action to be idempotent, meaning that, if I tanzufy a cluster I already tanzufied before, to not get an error. But yes, if it wasn't me the one that did it, I would agree that an error/confirmation can be the way.

jpmcb:
> I would expect some way of "identifying" who did what, so that, if uc cluster create was the way, then it takes the actions to be idempotent, else, can error out saying, hey, you already have this, installed by some other means, it's up to you to fix that.

Isn't that out of scope of what `unmanaged-cluster` is doing? We don't want to manage objects on the cluster in the way that managed clusters do. In my opinion, at this time, `unmanaged-cluster` should _not_ be attempting to lifecycle clusters or their components.

> What I was asking was for the tanzufy a cluster action to be idempotent, meaning that, if I tanzufy a cluster I already tanzufied before, to not get an error.

This PR was the only way I saw to accomplish that; when objects by name exist on a cluster, we skip creating them. That unfortunately, as Josh said, has some drifting ux and we need to better define what attaching and managing those objects means.

jorgemoralespou:
> Isn't that out of scope of what unmanaged-cluster is doing? We don't want to manage objects on the cluster in the way that managed clusters do. In my opinion, at this time, unmanaged-cluster should not be attempting to lifecycle clusters or their components.

I think you've misunderstood my comment. And probably this is the result of having in UC a side effect being a feature, which is "tanzufy a cluster". My point being that, if this function tanzufies a cluster, needs to be idempotent, and for that, it should know whether the "tanzu" related resources were added via this way, so that `idempotency` can be expected. This is just a matter of adding some annotations to the generated resources to understand provenance.

In any case, I agree that better definition of experience might be needed.

joshrosso:
> Sounds good - should we move https://github.com/vmware-tanzu/community-edition/issues/3580 into that project?

Done.

> In any case, I agree that better definition of experience might be needed.

:100:

PR #4043, opened 2022-04-14
Splits GettingStarted into GettingStarted and MC landing; adds WC landing

swalner-vmware:
[WIP] makes room for workload cluster workflow


miclettej (approved):
This looks good Shimon. We'll have to think through how to best make some of the CSS styling reusable across components so we are not defining duplicate styles. This can be done easily using standard scss files, but I'll have to look into creating variables for 'styled-components'

Issue #4102, opened 2022-04-19
Tanzu CLI update from 10.1 to 11 failed with - Error: could not read artifact "artifacts/pinniped-auth/v0.11.4/tanzu-pinniped-auth-linux_amd64": storage: object doesn't exist


trantor1:
## Bug Report

Tanzu cli update from commandline not working at this moment and failed with

`
Error: could not read artifact "artifacts/pinniped-auth/v0.11.4/tanzu-pinniped-auth-linux_amd64": storage: object doesn't exist
`

## Steps to Reproduce the Bug

* user@k8s-mgmt:~$ tanzu version
version: v0.10.1
buildDate: 2022-02-14
sha: 401d55b

* user@k8s-mgmt:~$ tanzu update
     the following updates will take place: 
     core v0.10.1 → v0.11.4
     package v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     pinniped-auth v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     secret v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     cluster v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     kubernetes-release v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     login v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
     management-cluster v0.10.1 → {v0.11.4 %!s(*cli.GCPBucketRepository=&{tce-tanzu-cli-framework artifacts core 0x13d53a0})}
would you like to continue? [y/n] y

## Environment Details

* Build version (`tanzu version`): version: v0.10.1, buildDate: 2022-02-14, sha: 401d55b
* Deployment (Managed/Unmanaged cluster): Managed
* Infrastructure Provider (Docker/AWS/Azure/vSphere): vSphere
* Operating System (client): Ubuntu 20.04.4 LTS, x86_64

Aradiv:
[...]

PR #4105, opened 2022-04-19
Issue Duplicator Utility (and move all utils to a single folder)

davidvonthenen:

For those that want this capability, this utility takes an issue from the TCE repo and creates a new duplicate issue in the Tanzu Framework repo. We do this when issues might be better suited for the TF repo instead of our repo.

When dup'ing the issue in the Tanzu Framework repo, this utility when creating the new issue:
- references the original issue in the TCE repo
- references the original author of the issue
- then copies and paste the original contents of the issue into the new TF issue

Example of this in action:
TCE Issue: https://github.com/vmware-tanzu/community-edition/issues/4102
duplicated...
TF Issue: https://github.com/vmware-tanzu/tanzu-framework/issues/2181

To run the utility, build this utility and place it in your path, either export an environment variable `GITHUB_TOKEN` or specify it on the command line, then just reference the issue number as it exists in the TCE repo:

```
GITHUB_TOKEN=<your token> dupissue -issue <TCE issue number>
```

Additionally, this also moves the tagger utility and this utility into a combined folder to house all utilities.

NA

NA

NA


hack/utils/dupissue/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	@echo "Error! GOLANGCI_LINT env var not set"
+else
+	$(GOLANGCI_LINT) run -v --timeout=5m
+endif
+
+get-deps: ## Get all go dependencies
+	go mod tidy
+
+test:
+	@echo "N/A: No unit tests for hack/packages"
```
stmcginnis, line 20: nit: should probably update from `hack/packages` to `hack/utils/dupissue` throughout.

And if updating, since this is a new file, it would be good to set the copyright year to 2022.

hack/utils/dupissue/dupissue.go
```diff
@@ -0,0 +1,138 @@
...
+func readTemplate() (string, error) {
+	template, err := os.ReadFile(TemplateFullPathFilename)
+	if err == nil {
+		fmt.Printf("Using local template\n")
+		return string(template), nil
+	}
+
+	homedir, err := os.UserHomeDir()
+	if err != nil {
+		fmt.Printf("Failed to get UserHomeDir: %v\n", err)
```
stmcginnis, line 78: Do we want to print out these messages within these calls? Looks like code calling these functions will print out the error again.

stmcginnis (dismissed):
Looks good. I pointed out some very minor things, but nothing I think is critical to merging this. Will wait to do the merge though in case we want to address them first.

hack/utils/dupissue/dupissue.go
```diff
@@ -0,0 +1,138 @@
...
+
+func getClientWithEnvToken() (*github.Client, error) {
+	var token string
+	if v := os.Getenv("GITHUB_TOKEN"); v != "" {
+		token = v
+	}
+
+	if token == "" {
+		return nil, fmt.Errorf("token is empty")
+	}
```
stmcginnis, line 31: Could do:

```go
token := os.Getenv("GITHUB_TOKEN")
if token == "" {
    return nil, fmt.Errorf("token is empty")
}
```

hack/utils/dupissue/dupissue.go
dvonthenen: yea, we dont have to since it's a fairly simple util

hack/utils/dupissue/Makefile
```diff
@@ -0,0 +1,30 @@
+# Copyright 2021 VMware Tanzu Community Edition contributors. All Rights Reserved.
```
jpmcb, line 1: I keep forgetting ... do files that are _created_ in 2022 get  `Copyright 2022`? Or is it the original copyright date in 2021?

hack/utils/dupissue/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	@echo "N/A: No unit tests for hack/packages"
+
+e2e-test:
+	@echo "N/A: No e2e tests for hack/packages"
+
+build: ## Build the executable
+	go build -o dupissue dupissue.go
+
+run: ## Run the dupissue program to duplicate issues to the tanzu framework repo
+	@echo "N/A: No e2e tests for hack/packages"
```
jpmcb, line 29: Should this chunk get updated to print that the user should build the binary first (to then be able to pass arguments to it)?

hack/utils/dupissue/dupissue.go
```diff
@@ -0,0 +1,138 @@
...
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+const (
+	// TemplateFullPathFilename filename
```
jpmcb, line 17: Is this commented out code needed?

jpmcb (dismissed):
Amazing!! I love this 👍🏼

A few small nits:

hack/utils/dupissue/dupissue.go
dvonthenen: yup!

hack/utils/dupissue/dupissue.go
dvonthenen: it's still being used. you can provide your own custom template and the search order is:
- next to he binary
- homedir
- or just use the default (contained in the source code)

hack/utils/dupissue/Makefile
dvonthenen: changed the comment around so there isnt a run command. the idea is the user builds the utility and drops it somewhere on their filesystem. we dont want to run from the makefile here

jpmcb: +1

hack/utils/dupissue/dupissue.go
jpmcb: 👍🏼

jpmcb (approved):

PR #4124, opened 2022-04-20
Add App Toolkit 0.2.0 package

ryankilroy:
Introduces App Toolkit 0.2.0 which swaps out a bunch of "ytt" steps for official TCE package usage!
- Use kpack-dependencies 0.0.9
- Use cartographer-catalog 0.3.0
- Bump to cartographer 0.3.0
- Add developer-namespace feature and requirement on secretgen-controller
- Switch App Toolkit Testing framework from go to bash
- Update Readme

Introduced a new bash testing script which does the following:
- Creates a new cluster
- Updates PackageRepo if url is provided
- Installs secretgen controller and sets up secrets
- Installs app-toolkit
- Tests a workload against app-toolkit

## Special notes for reviewer
This package depends on a few other packages being introduced into TCE, so you'll likely need to test it against a TCE 0.12.0 RC package repo. Namely, it expects:
- kpack-dependencies 0.0.9
- cartographer-catalog 0.3.0
- cartographer 0.3.0
- secretgen-controller 0.8.0


addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
```diff
@@ -0,0 +1,48 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: ootb-supply-chain-source-to-url-workload
+  namespace: #@ data.values.developer_namespace
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ootb-supply-chain-source-to-url-workload
```
scothis, line 45: referencing a specific cluster role for a specific supply chain is fragile. Using an aggregated cluster role to collect permissions into a stable name that can then be bound is preferable.

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
```diff
@@ -0,0 +1,48 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:yaml", "yaml")
+
+#@ if/end data.values.developer_namespace != "default":
+---
+apiVersion: v1
+kind: Namespace
+metadata:
```
scothis, line 8: other tools may expect to be able to create a namespace or add a label/annotation to define extra behavior. `kapp` will fail if the namespace already exists, and will remove any other labels/annotation a user expects to be able to place on the resource.

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
```diff
@@ -0,0 +1,48 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:yaml", "yaml")
+
+#@ if/end data.values.developer_namespace != "default":
+---
+apiVersion: v1
+kind: Namespace
```
scothis, line 7: Does kapp correctly handle the .spec.finalizers array that namespaces have?

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
```diff
@@ -0,0 +1,48 @@
...
+    secretgen.carvel.dev/image-pull-secret: ""
+type: kubernetes.io/dockerconfigjson
+data:
+  .dockerconfigjson: e30K
+
+---
+apiVersion: v1 
+kind: ServiceAccount
+metadata:
+  name: default
```
scothis, line 27: It's best not to bind additional permissions to the default service account. Workloads should not be run with the same permissions as a supply chain.

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
```diff
@@ -0,0 +1,48 @@
...
+  namespace: #@ data.values.developer_namespace
+  annotations:
+    kapp.k14s.io/create-strategy: fallback-on-update
+secrets:
+- name: registry-credentials
```
scothis, line 32: Is this needed? I understand using a pull secret that may be used to load sidecar images from a private registry

addons/packages/app-toolkit/0.2.0/README.md
```diff
@@ -0,0 +1,251 @@
...
+ 1. Create a registry secret `registry-credentials` using the below command
+
+    ```shell
+    tanzu secret registry add registry-credentials --server REGISTRY_URL --username REGISTRY_USER --password REGISTRY_PASS --export-to-all-namespaces`
+    ```
+
+      * `REGISTRY_URL` - URL for the registry you plan to upload your builds to.
+        * For Dockerhub, it would be <https://index.docker.io/v1/>
+        * For GCR, it would be gcr.io
+        * For Harbor, it would be [TO BE FILLED]
```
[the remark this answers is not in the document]
scothis: best to use a domain that is either under our control or can never be registered. Something like `myharbor.example`, or if the `.example` tld is too odd `myharbor.example.com`

addons/packages/app-toolkit/0.2.0/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+| Config | Values | Description |
+|--------|--------|-------------|
+| cartographer-catalog | | [See cartographer catalog documentation](https://tanzucommunityedition.io/docs/package-readme-cartographer-catalog-0.3.0/#configuration)|
+| cert_manager | | [See cert-manager documentation](https://tanzucommunityedition.io/docs/package-readme-cert-manager-1.6.1/#configuration)|
+| contour | | [See contour documentation](https://tanzucommunityedition.io/docs/package-readme-contour-1.20.1/#configuration-reference) |
+| excluded_packages  | Array of package names | Allows installers to skip deploying named packages |
+| knative_serving | | [See knative documentation](https://tanzucommunityedition.io/docs/package-readme-knative-serving-1.0.0/#configuration) |
+| kpack | | [See kpack documentation](https://tanzucommunityedition.io/docs/package-readme-kpack-0.5.1/#kpack-configuration) |
+| kpack-dependencies | | [See kpack dependencies documentation](https://tanzucommunityedition.io/docs/package-readme-kpack-dependencies-0.5.2/#kpack-dependencies-configuration) |
```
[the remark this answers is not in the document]
ryanmattcollins: Sorry, can you clarify? I believe the latest versions of `kpack` and `kpack-dependencies` respectively are 0.5.2 and 0.0.9. The docs for `kpack-dependencies` aren't published yet either, so this was more of an educated guess at where they will eventually live.

addons/packages/app-toolkit/0.2.0/bundle/config/kpack-dependencies.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - secretRef:
+      name: kpack-dependencies-values
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kpack-dependencies-values
+  namespace: tanzu-package-repo-global
+stringData:
+  values.yaml: #@ "kp_default_repository: " + data.values.kpack.kp_default_repository
```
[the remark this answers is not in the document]
ryanmattcollins: Yep, good catch. The README hadn't been updated

addons/packages/app-toolkit/0.2.0/README.md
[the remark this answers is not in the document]
tylerphelan: is it unexpected that the link for kpack deps docs uses kpack's version `0.5.2`? `https://tanzucommunityedition.io/docs/package-readme-kpack-dependencies-0.5.2`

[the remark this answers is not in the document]
tylerphelan: ultimately docs links can probably be vetted once they're live

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
ryanmattcollins: Notes from Zoom Call:
This could make it onerous if the user adds another supply chain because it means we'll have to specify another role binding. Aggregating it would mean less toil in setting up other supplychains because we could use the same role and binding and let the aggregated role fetch other roles by using labels.

Nice to have feature that we'll add to the backlog

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
ryanmattcollins: **Notes from Zoom Call:**
This could make it tricky if the user wants the namespace created by another tool. By specifying it here, we give `kapp` ownership meaning it will consistently reconcile it to this state making it impossible for other tools to modify it. Suggested fix is to add a flag that disables namespace creation in App Toolkit.

We predict App Toolkit won't be used any time soon for anything more than a playground, so we can cross this bridge when we get feedback that it is bothersome.

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
ryanmattcollins: **Notes from Zoom Call:**
Seems that `kapp` does, but even if it doesn't it should. Not our problem 😛

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
ryanmattcollins: **Notes from Zoom Call:**
Unfortunately, we're stuck using the default service account until cartographer-catalog fixes a bug where providing a service account to the package actually uses the new service account instead of default. Once that is fixed, we should be able to specify and use a different service account. TAP also does it this way, and that's bad.

Add to backlog and follow-up with cartographer-catalog to track their issue

addons/packages/app-toolkit/0.2.0/bundle/config/developer-namespace.yaml
ryanmattcollins: **Notes from Zoom Call:**
Having a separate `secret` and `imagePullSecret` for the supply chain to handle publishing of images and pulling of images is recommended. We should implement separate credentials rather than using the `registry-credentials` for both.

Add to the backlog for next release

joshrosso:
Digging into this today or tomorrow.

Who is going to review and provide approval from the Tanzu Application Toolkit side? Is that @tylerphelan or @scothis ?

kartiklunkad26:
@scothis given Kris Applegate is OOO (who is the usual reviewer), and you have at least reviewed a part of the package, would you be up for reviewing the overall PR as well?

addons/packages/app-toolkit/0.2.0/README.md
```diff
@@ -0,0 +1,245 @@
...
+As mentioned in the pre-requisite, ensure you provide the image registry configuration and ingress configuration while installing App-Toolkit package.
+
+```yaml
+contour:
+
+knative_serving:
+
+kpack:
+  kp_default_repository:
+  kp_default_repository_username:
```
tylerphelan, line 84: would we feel comfortable adding a templating-level validation to ensure these `kpack` fields are set? because the kpack package does not validate that the credentials are set and it could make installations safer.

kartiklunkad26: will take this as a feature request for future App Toolkit releases.

kartiklunkad26 (approved):

jpmcb:
Can we get final approval PR review from someone on the @vmware-tanzu/pkg-app-toolkit-owners? Thanks much.

jpmcb:
I pushed an alpha of our repo from this PR to `0.12.0-alpha-3899.1` and installed with:

```
tanzu unmanaged-cluster create test-cluster --additional-repo projects.registry.vmware.com/tce/main:0.12.0-alpha-3899.1
```

App toolkit is there and installed via instructions in README 👍🏼 
```
❯ tanzu package available list

  NAME                                                 DISPLAY-NAME                 SHORT-DESCRIPTION                                                                                                  LATEST-VERSION
  app-toolkit.community.tanzu.vmware.com               App-Toolkit package for TCE  Kubernetes-native toolkit to support application lifecycle                                                         0.2.0
```

mstergianis (approved):
Things look really good overall. I'm going to post this and continue with my installation to ensure things look good on that side. I don't consider any of my requests blocking, so approved :smile:

joshrosso (approved):

joshrosso:
package included in:

```
projects.registry.vmware.com/tce/main:0.12.0-rc4
```

addons/packages/app-toolkit/0.2.0/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+For a simple tutorial on how to use Application Toolkit, please refer to the "Getting Started with Creating a Tanzu Workload" guide.
+
+## Supported Providers
+
+Application Toolkit is currently tested with Unmanaged Clusters on any of the below providers.
+
+| Unmanaged Clusters| AWS | Azure | vSphere | Docker |
+|-------------------|-----|-------|---------|--------|
+| ✅                | ✅   | ✅     | ✅       | ✅      |
```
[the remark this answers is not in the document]
kartiklunkad26: @ryanmattcollins I'm good with reverting this section to how it looks for v0.1.0 given we haven't explicitly supported managed clusters yet.

addons/packages/app-toolkit/0.2.0/README.md
kartiklunkad26: +1 to @mstergianis. The validation should be done in `kpack` rather than doing so in App Toolkit. `kpack` as a package does not function without a valid set of registry credentials.

addons/packages/app-toolkit/0.2.0/README.md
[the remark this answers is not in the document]
ryanmattcollins: SGTM

PR #4130, opened 2022-04-20
Add update strategy and node selector overlay to Antrea 1.2.3 package

blc1996:
Signed-off-by: Lucheng Bao <luchengb@vmware.com>

We need to enforce the update strategy of all the deployment and daemonset running on Supervisor cluster. Also, add nodeSelector configuration to deployments.

Fixes: # https://github.com/vmware-tanzu/tanzu-framework/issues/1850

Manually tested the templates

I understand that Antrea intentionally set the update strategy of deployment as reCreate to avoid running 2 instances of Antrea controller pods at the same time. However, it is required by superviser cluster in vSphere that we must enforce these strategies. This is the current implementation in Supervisor cluster's guest cluster controller. Manual testing also didn't show immediate error.


shivaani0505 (dismissed):
LGTM

knabben (approved):

addons/packages/antrea/1.2.3/test/antrea_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	appsv1 "k8s.io/api/apps/v1"
```
seemiller, line 11: @blc1996 This file is failing the Golang linter check here (I believe these imports should be alphabetized). You can run the check yourself with `make lint` to see the error.

seemiller (changes requested):

addons/packages/antrea/1.2.3/test/antrea_test.go
blc1996: Fixed

seemiller (approved):

PR #4131, opened 2022-04-20
Add update strategy and node selector overlay to Calico 3.22.1 package

blc1996:
Signed-off-by: Lucheng Bao <luchengb@vmware.com>

We need to enforce the update strategy of all the deployment and daemonset running on Supervisor cluster. Also, add nodeSelector configuration to deployments.

Fixes: # https://github.com/vmware-tanzu/tanzu-framework/issues/1850

Manually tested the templates

I understand that Calico intentionally set the update strategy of deployment as reCreate to avoid running 2 instances of Calico controller pods at the same time. However, it is required by superviser cluster in vSphere that we must enforce these strategies. This is the current implementation in Supervisor cluster's guest cluster controller. Manual testing also didn't show immediate error.


12345lcr (commented):
We have some template testing on `addons/packages/calico/3.22.1/test`. Perhaps it would be better to add some tests there for the newly added fields

12345lcr (approved):
LGTM

PR #4160, opened 2022-04-23
Make unmanaged e2e test runs explicit action

### LINKED ISSUE #4112: `unmanaged-cluster` e2e tests should _not_ be included in unit tests
## Bug Report

Currently, when running the `go test` suite for `unmanaged-cluster`, the e2e tests also get picked up.

This is ok for now, but problematic in the long run since e2e tests take longer and are far more resource intensive. We shouldn't slow down unit testing turn around time with e2e tests.

## Expected Behavior

`go test ./...` should run _only_ the unit tests.

## Steps to Reproduce the Bug

Run `go test ./...` in `unmanaged-cluster` directory and notice the e2e tests run.

stmcginnis:

The current setup causes e2e tests to be run like any normal unit test,
making it difficult to quickly run unit testing.

This separates out the e2e tests so they need to be explicitly invoked
by requiring `-tags=e2e` in the invocation of `go test`.

Fixes: #4112 

Ran unit and e2e tests locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -507,6 +507,6 @@ vsphere-management-and-workload-cluster-e2e-test:
 	BUILD_VERSION=$(BUILD_VERSION) test/vsphere/run-tce-vsphere-management-and-workload-cluster.sh
 
 unmanaged-cluster-e2e-test:
-	cd cli/cmd/plugin/unmanaged-cluster/test/e2e && BUILD_VERSION=$(BUILD_VERSION) ROOT_DIR=$(ROOT_DIR) go test -test.v -timeout 180m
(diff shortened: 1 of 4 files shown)


aman556 (dismissed):
LGTM! 
Thanks Sean.

joshrosso (dismissed):
lgtm.

In [the issue](https://github.com/vmware-tanzu/community-edition/issues/4112#issuecomment-1107537432) you mentioned:

> I think we should adopt this and get it documented as what we expect when adding e2e tests.

Can we add documentation somewhere around this behavior change?

I believe the right location is:

https://tanzucommunityedition.io/docs/main/contribute/contributing/#running-tanzu-community-edition-specific-plugin-tests

jpmcb (approved):
This LGTM!! Also looking at the followup PR

Issue #3588, opened 2022-03-17
Proposal: Introduce `wizard` bootstrapping plugin


jpmcb:
In addition to the [Guided Graphical UI #3235](https://github.com/vmware-tanzu/community-edition/issues/3235), we should consider implementing a `wizard` CLI plugin. This plugin would walk users through bootstrapping a management / unmanaged cluster, _right in the CLI._

### _But why?_

Some users don't have access to a web browser (like in a headless environment or when on a remote server). Or their network configuration won't allow for easily tunneling a request to their bootstrapping machine (like deploying through a jumpbox).

This issue needs a formal proposal and is a placeholder

---

Examples:
- [AWS CLI configuration wizard](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-wizard.html)
- [Powerlevel10k wizard](https://github.com/romkatv/powerlevel10k#configuration-wizard)

joshrosso:
:+1: When it comes around to proposal let's reconsider the trade-offs of embedding this functionality alongside plugins that benefit from a wizard, vs doing a dedicated wizard plugin.

Issue #3235, opened 2022-02-23
TCE Guided UI Experience


miclettej:
## Abstract
We believe one of the largest areas of friction for TCE end users is successfully creating clusters. This process includes:

1. Determine which cluster deployment model is most appropriate
    a. ie: unmanaged vs managed
2. Deploy a cluster
3. Delete a cluster

Along with introducing a new UI, this proposal focuses on an initial use case where a user can accomplish the above with a simplified user experience by leveraging sensible defaults. This increases the likelihood of success for users with limited expertise in Kubernetes and cloud-providers.

## Proposal
To achieve the above, we will:

Offer a web-based UI that provides guided and simplified user experiences for underlying CLI functionality. The initial focus will be to provide simplified paths for cluster creation. Including:

- Unmanaged-cluster running locally
- Management and workload clusters running on AWS
- Management and workload clusters running on vSphere
- Management and workload clusters running on Azure
- Management and workload clusters running locally via Cluster API Docker (CAPD)

Users should be able to achieve each of the above without subject matter expertise in Kubernetes or the underlying infrastructure provider. As this new UI model is proved out, we’ll expand into other areas the community sees as viable. These could include things like:

- Advanced cluster configurations
- Cluster upgrades (via Tanzu Kubernetes Releases TKrs)
- Troubleshooting 

The above list can also be considered our non-goals for this initial effort.
The remainder of this proposal describes how we are going to achieve the above.

### Create new UI Tanzu CLI plugin
A new CLI plugin, named ui, will be introduced to TCE. This plugin will live in the TCE repository as we incubate the project. Based on usage and adoption, we’ll reconsider this home over time. This plugin will be used to launch the web-based UI along and for interacting with other CLI plugins, based on user input. A high-level diagram of this relationship is:

![image](https://user-images.githubusercontent.com/6200057/157936451-f9ec336c-922b-4c6a-bda8-80c4502d84ac.png)

The ui plugin is intended to serve as a UI platform for surfacing selected underlying Tanzu CLI functionality in the form of a GUI. The UI is to be built on web technologies and embedded in a new Tanzu CLI plugin which will be entirely dedicated to running the UI and communicating with other Tanzu CLI plugins. 

[...]

PR #4212, opened 2022-04-26
🧙 POC: Implements `wizard` command for unmanaged-cluster

jpmcb:
### Summary

Include a proof of concept `wizard` command in the unmanaged-cluster plugin which provides an interactive terminal user interface for bootstrapping clusters and gain community feedback.

Fixes: #3588

### Motivation

In addition to the [Guided Graphical UI in #3235](https://github.com/vmware-tanzu/community-edition/issues/3235), users need a terminal based `wizard`. At a high level, this should walk users through bootstrapping cluster, _right in the CLI._

_Sure, cool. But why?_

Some users don't have access to a web browser (like in a headless environment or when on a remote server). Or their network configuration won't allow for easily tunneling a request to their bootstrapping machine (like deploying through a jumpbox). In short, we should make deploying a cluster as simple as possible via the CLI.

This PR implements that proof of concept in `unmanaged-cluster` and includes 3 options: cluster name, number of worker nodes, & number of control plane nodes

### 🔉 Demo Video

https://user-images.githubusercontent.com/23109390/165395342-1f6ef755-ecdb-4502-87f7-a1b850424118.mp4

### Goals

- Bootstrap unmanaged-clusters using options configured interactively in the CLI

### Non-goals

- Create 1 to 1 parity of options with the browser based GUI
- Bootstrap managed-clusters
- Create a separate `wizard` plugin

##### User Story 1: Configuration

- User can configure unmanaged-cluster name for cluster via typing their prefered cluster name in the TUI.
- User can configure unmanaged-cluster number of worker nodes via typing their preferred node count in the TUI
- User can configure unmanaged-cluster number of control plane nodes via typing their preferred control plane node count in the TUI

##### User Story 2: Validation (still to do)

- User receives immediate feedback on their inputs given constraints of unmanaged-cluster:
  - If user provides no name in TUI user is interactively prompted to provide a name.

#### Graduation details

This POC is ready to go and would be an ok starting point to get community feedback. Otherwise, we can iterate here and add more configuration options and features as we go.

Looking for feedback on this as well.

### Implementation details

This rough POC uses bubble and  bubbletea for the TUI control flow and interactive components. Note the additions to the `go.mod` and `go.sum`

`tanzu unmanaged-cluster wizard`

### Appendix / useful terms

- TUI: Terminal User Interface; generally refers to an _interactive_ graphical user interface
- bubble, bubbletea & lipgloss: a popular TUI Go suite of libraries that enable flow control, TUI components, and styling
  - https://github.com/charmbracelet/bubbletea & https://github.com/charmbracelet/bubble

Useful examples in other CLIs & inspiration:
- [AWS CLI configuration wizard](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-wizard.html)
- [Powerlevel10k wizard](https://github.com/romkatv/powerlevel10k#configuration-wizard)

( ⬇️  go ahead and give this a thumbs up / thumbs down to reflect your thoughts here!)


dvonthenen:
Dude yes!!!! I love this!!!

joshrosso:
This is really awesome!

Based on the demo, I wonder if:

1. wizard should not actually trigger bootstrapping, but instead be a way to generate a proper config file. I’d be open to it prompting to jump right into bootstrapping, but this decoupling seems good for us.
2. wizard should not be another command (thus having 2 ways to create a cluster) but instead be a flag on the existing configure command.

RussellHamker:
Looks great. Really love the idea of having a terminal wizard! Keep up the great work.

garrying:
A thought while I stepped through the POC: I wonder if there should be messaging to help users know how to escape from the wizard. `ctrl-c`ing doesn't escape (was my initial intuition), while `esc` does.

jpmcb:
> ctrl-cing doesn't escape (was my initial intuition), while esc does.

Sure thing, that actually wouldn't be difficult to also add as functionality alongside `<esc>` 👍🏼

PR #4225, opened 2022-04-27
Add timeout to cache restore in Actions

### LINKED ISSUE #3961: Restore Golang Cache Step in GitHub workflow is timing out
Example - https://github.com/vmware-tanzu/community-edition/actions/runs/2074379555/attempts/1 - https://github.com/vmware-tanzu/community-edition/runs/5780484278?check_suite_focus=true

https://github.com/vmware-tanzu/community-edition/actions/runs/2140183772/attempts/1 - https://github.com/vmware-tanzu/community-edition/runs/5953769059?check_suite_focus=true

Restore Golang Cache job took up 6 hours to execute, timing out the GitHub workflow as [6 hours is the default max timeout time for jobs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)

stmcginnis:

We have been seeing the restore of cache files basically hang during the
restore process, causing it to eventually timeout after 6 hours of
inactivity. This fails the entire job which then needs to be rerun.

This change sets a 10 minute timeout for the cache restore. Under normal
conditions, the restore is fairly quick. So if it's taken 5 minutes, we
know something isn't quite right.

It also sets the flag to tell the action to continue on this step
failure. It's better to have decreased performance from cache misses
than to fail if it's not present.

Fixes: #3961

### DIFF
--- a/.github/workflows/build-daily.yaml
+++ b/.github/workflows/build-daily.yaml
@@ -55,6 +55,8 @@ jobs:
           echo "::set-output name=date::$(date -u "+%Y-%m")"
       - name: Restore Go Cache
         uses: actions/cache@v3
+        timeout-minutes: 10
+        continue-on-error: true
         with:
           path: ~/go/pkg/mod
           key: ${{ runner.os }}-go-${{ steps.get-date.outputs.date }}-${{ hashFiles('**/go.sum') }}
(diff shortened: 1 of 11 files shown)


davidvonthenen (approved):
Looks good to me. Let's give this a try.

jpmcb (approved):
Looks good to me - read through a few of the `actions/cache` issues on this and seems like this is the best work around for now! 👍🏼

PR #4227, opened 2022-04-27
addons/packages/pinniped: add unit testing

ankeesler:

Adds unit testing for Pinniped package

None

Ran the unit tests locally with `make -C addons/packages/pinniped test`.

None.


stmcginnis:
Ah, the expectation is there is a `Makefile` in the package directory. Take a look at https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.20.1/test/Makefile - you can probably wholesale copy that over.

benjaminapetersen (dismissed):
lgtm

addons/packages/pinniped/test/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[0-9A-Za-z_-]+:.*?##/ { printf "  \033[36m%-45s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+# We pass -count 1 to the unit tests so that the go test cache is ignored. Since
+# the tests reference files outside of this package (i.e., they reference test
+# fixtures in other packages), then we could end up with false positive test runs
+# because our test fixtures is not part of the go test cache key.
+test: ## Run the unit tests
+	go test -count 1 $(GO_TEST_FLAGS) .
+
+update: ## Update all expected test fixtures with their actual value
```
stmcginnis, line 21: Can we just add certain directories to a .gitignore path and dynamically create them every time by making this a prereq target to the `test` target?

```
test: update ## Run the unit tests
```

ankeesler: mmm. fair point. that would avoid the onslaught of new files. grumble grumble grumble package directory structure grumble grumble grumble...

@benjaminapetersen - what do you think of this suggestion?

@stmcginnis - i'm out tomorrow but can incorporate this monday if ben thinks it is a good idea as well

ankeesler: @stmcginnis - one thing that just came to mind. we have different test cases for each package version. e.g., executing the 0.4.3 package on a set of data values might yield a different output than executing the 0.12.1 package. furthermore, we have times where we wouldn't want to run a test case on a particular package version.

i know we could structure our single test fixture directory to work with these constraints, but would you be amenable to us just checking in the test cases per each package to avoid test setup complexity? that would look the same if each of the package versions were git branches, anyway.

stmcginnis: I think I grok what you're saying - if so, yep, that makes sense. Isn't that how you have it structured today?

For developer testing, it might be useful to have targets per package version. So like a `test-v0120`, `test-v0121`, etc. The the expected `test` target could just be an uber wrapper that causes all of the individual ones to run.

ankeesler: Yep - that is indeed how we have it structured today. Sounds good - I added a `test-v0121` target since that is the only package that we manage right now. I am a little worried about whether we will remember to keep adding these, but at the very least it does provide a nice example of how to run the `make test` target.

benjaminapetersen: So I'm down with `test-v0121` providing an example of how to run `make test`, but why would we want to run tests on only one package?

benjaminapetersen (approved):

addons/packages/pinniped/test/Makefile
benjaminapetersen: (assuming just for dev purposes)

sabbey37 (approved):

PR #4241, opened 2022-04-28
addons/packages/pinniped/0.12.1: fix template error when idp_type:none results in no render

benjaminapetersen:
Signed-off-by: Benjamin A. Petersen <ben@benjaminapetersen.me>

There is currently a template rendering error which is a safety mechanism of `kapp`:

```
kapp: Error: Trying to apply empty set of resources will result in deletion of resources on cluster. Refusing to continue unless --dangerous-allow-empty-list-of-resources is specified.
```
however, when `identity_management_type: none` we explicitly want an empty render set. 

Fixes: #

### DIFF
--- a/addons/packages/pinniped/0.12.1/package.yaml
+++ b/addons/packages/pinniped/0.12.1/package.yaml
@@ -23,4 +23,5 @@ spec:
               - kbld-config.yaml
               - .imgpkg/images.yml
       deploy:
-        - kapp: {}
+        - kapp:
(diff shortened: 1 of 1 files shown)


sabbey37 (commented):
Looks good!  Are there any concerns/reasons we shouldn't enable the flag?

benjaminapetersen:
not for us, deletion of EVERYTHING is exactly what we want!

sabbey37 (approved):

PR #4267, opened 2022-04-29
Automaticaly Merge Select Dependabot PRs

davidvonthenen:

The AWS SDK is updated daily which in turn always creates a new PR each to bump the dependency version. For select (ie only one) utility project, we should just merge the PR automatically. This code isn't used in the TCE release and only is run when the webhook service is compiled and uploaded to the webhook servers to be updated.

NA

NA

NA


stmcginnis (approved):
I like it!

PR #4282, opened 2022-05-02
Add go and cloning to auto merge of dependabot

stmcginnis:

The action for performing the automatic merge of some updates was
failing due to the environment not being fully set up. This adds
installing go and cloning the repo so the `ensure-deps` step can run.

### DIFF
--- a/.github/workflows/build-automatically-merge-pr.yaml
+++ b/.github/workflows/build-automatically-merge-pr.yaml
@@ -15,6 +15,13 @@ jobs:
     name: Automatically Merge PR
     runs-on: ubuntu-latest
     steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v3
+        with:
+          go-version: "1.17"
+        id: go
+      - name: Check out code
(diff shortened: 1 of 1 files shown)


davidvonthenen (approved):

jpmcb:
Hmm, I didn't know we wanted to automatically merge dependabot PRs.

Will we get into problems where dependabot is attempting to re-base a PR after a previous one has merged creating conflicts? Or will the `gh` CLI tool prevent us from creating those clashes.

stmcginnis:
Yeah, that's a valid concern @jpmcb. I was worried at first too, but since we're restricting this to only `"hack/runner/webhook/go.**"`, I think the risk impact is pretty low. David and I had chatted about it and it would be easy enough to revert something if we find an issue.

Definitely would not want to do this repo-wide.

jpmcb:
Ahhhh gotcha gotcha - that makes sense. I missed the detail where this was only running on `- "hack/runner/webhook/go.**"`

Thanks for the clarification! 🙏🏼

Issue #4325, opened 2022-05-03
Add "image" to the schema for antrea versions 0.11.3, 0.13.3, 1.2.3


shivaani0505:
## Feature Request

Add "image" to the schema for antrea versions 0.11.3, 0.13.3, 1.2.3

PR #4326, opened 2022-05-03
Add "image" to the schema for antrea package versions: 0.11.3, 0.13.3, 1.2.3

shivaani0505:
Add "image" to the schema for antrea package versions: 0.11.3, 0.13.3, 1.2.3

Fixes: #4325 

Run

### DIFF
--- a/addons/packages/antrea/0.11.3/bundle/config/schema.yaml
+++ b/addons/packages/antrea/0.11.3/bundle/config/schema.yaml
@@ -31,3 +31,12 @@ antrea:
       AntreaTraceflow: false
       #@schema/desc "Flag to enable/disable network policy stats"
       NetworkPolicyStats: false
+  image:
+    #@schema/desc "The repository of antrea image"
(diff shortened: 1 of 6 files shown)


addons/packages/antrea/0.11.3/bundle/config/schema.yaml
```diff
@@ -31,3 +31,12 @@ antrea:
       AntreaTraceflow: false
       #@schema/desc "Flag to enable/disable network policy stats"
       NetworkPolicyStats: false
+  image:
+    #@schema/desc "The repository of antrea image"
```
blc1996, line 35: I think these added fields should be nullable.

shivaani0505: I am following the same logic as vsphere-cpi

knabben (approved):

blc1996 (approved):
lgtm

PR #4349, opened 2022-05-04
TCE extension for DD

jorgemoralespou:
Initial version of Tanzu Community Edition extension for Docker Desktop

Signed-off-by: Jorge Morales Pou <jorgemoralespou@gmail.com>


jpmcb:
I dove into the dependency failures and have a fix for most of the dependencies except for `react-scripts`.

But the `npm audit` alerts around `react-scripts` aren't real, unfortunately. And there's no way to prevent `dependency-check` from ignoring NPM `devDependencies` (which don't get shipped but are part of the build toolchain)

On `npm audit --production`, everything is good, but it'll keep failing on dev deps through this github action.

I think we're ok to merge this and 

Ref:
- https://github.com/facebook/create-react-app/issues/11174
- https://github.com/actions/dependency-review-action/issues/43

jpmcb (approved):

PR #4359, opened 2022-05-04
UC: raise log level for some log messages

### LINKED ISSUE #4351: Update unmanaged-cluster logging to include log lines only relevant to docker extension
To support embedding unmanaged-cluster in the docker desktop extension, we must update the logging to only include minimal details that are relevant to running it in contexts such as the docker desktop extension.

image

In the near-term, I suggest increasing our info logging level of the lines mentions above. Unmanaged-cluster should then default to that higher verbosity.

Long-term, we'll look to re-write the interactions from the extension to unmanaged-cluster and ensure the logs we consume are more eloquently handled.

stmcginnis:

This raises the level of some local path and other unmanaged cluster
output that we don't want emitted by default for the Docker Desktop
integration. It also raises the default verbosity of commands so even
though these are logging at a higher level, they will still be printed
in a normal run.

This allows the Docker Desktop integration to explicitly run these calls
with a lower verbosity and not include the messages.

Fixes: #4351 

```
go run . create foo
go run . rm foo
go run . create -v 2 foo
go run . rm -v 2 foo
```

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
@@ -58,7 +58,7 @@ func configure(cmd *cobra.Command, args []string) error {
 		clusterName = args[0]
 	}
 
-	log := logger.NewLogger(TtySetting(cmd.Flags()), 0)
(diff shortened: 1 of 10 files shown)


joshrosso (approved):
lgtm. thanks for introducing the Default `const`.

PR #4360, opened 2022-05-05
UI plugin: WC - adds CC data for all providers

swalner-vmware:


cli/cmd/plugin/ui/web/tanzu-ui/src/views/workload-cluster/WorkloadClusterUtility.tsx
```diff
@@ -0,0 +1,52 @@
...
+//       we accumulate them in a single object per management cluster.
+//       This makes using them much easier (later), and prevents "collisions" if the user switches management clusters.
+function getClusterVariableData(managementCluster: ManagementCluster, state: any): any {
+    if (!managementCluster || !managementCluster.name) {
+        console.error('getClusterVariableData() called with undefined/unnamed cluster')
+        return {}
+    }
+    const ccVarData = { ...state.data.CLUSTER_CLASS_VARIABLE_VALUES }
+    const ccVarClusterData = ccVarData[managementCluster.name]
+    return ccVarClusterData ? ccVarClusterData : {}
```
miclettej, line 27: If you wanted to get rid of the expression, I thiiiiink you can just `return ccVarData[managementCluster.name] || {}` which should return the empty object if ccVarData[...] is undefined

miclettej (approved):
One small suggestion you can try but looks great

PR #4365, opened 2022-05-05
Update load-docs.sh

a-mccarthy:
Updates load-docs script to look for files in the release branch in the docs/edge folder.

This PR also needs to be merged at the same time that we update the docs in the release branches

Fixes: #


a-mccarthy:
i noticed that the PR i made yesterday had the whole folder changes again. which are already in main.

kcoriordan (approved):

PR #4374, opened 2022-05-05
Add doc for Component list

### LINKED ISSUE #3940: Expose all the versions that make up a TCE release.
When we release TCE, we should have an easy way to expose users to all the software that underlies a release.

We also have multiple TKrs that make up a release, so we need to introspect our `tkg-bom`, which points to our `tkr-compatibility`, then for each tkr in that compatibility file, we should gather this data. Output would look something like:

* v1.22.5
    * calico-version:
    * antrea-version:
    * vsphere-csi-version: 
* v1.21.7
    * calico-version:
    * antrea-version:
    * vsphere-csi-version:

kcoriordan:

Not sure if this is something you want to merge or not, I know the idea here is to do something automated. But wanted to get something in for this. 
This is a component list taken from the TKR file with some clean up. 
If this is something we want to include, maybe it would be better as a section in the previous Support Matrix topic, rather than out as a standalone topic in the toc

Steps take:

Install v0.12.0-rc.2
From the `\.config\tanzu\tkg\bom\tkr-bom-v1.22.8+vmware.1-tkg.1-tf-v0.11.4`, copy component list (row 4 to row 226) to excel and clean up, copy in to md file and format

Fixes: #3940

Preview here: https://deploy-preview-4374--quirky-franklin-8969be.netlify.app/docs/edge/components/

### DIFF
--- a/docs/site/content/docs/edge/components.md
+++ b/docs/site/content/docs/edge/components.md
@@ -1,0 +1,31 @@
+# Components
+
+The Tanzu Community Edition contains the following software component versions:
+
+- ako-operator:  - version: v1.5.0+vmware.6  
+- antrea:  - version: v1.2.3+vmware.4-standard  
+- calico:  - version: v3.19.1+vmware.1  
+- carvel-secretgen-controller:  - version: v0.7.1+vmware.1  
+- cloud_provider_vsphere:  - version: v1.22.6+vmware.1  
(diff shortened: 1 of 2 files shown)


stmcginnis:
@seemiller can you take a look at this one?

seemiller (approved):

PR #4375, opened 2022-05-05
Fix dependabot automerge

davidvonthenen:

This attempts to fix the dependabot automerge for daily dependency updates.

NA

Only way to really test is to try it out....

NA


stmcginnis (approved):
Let's give it a shot!

PR #4380, opened 2022-05-05
Container deleted out of band checked

### LINKED ISSUE #4352: Docker extension needs to handle cases where the underlying TCE cluster (container) is killed and/or removed
In the current extension, if you `kill` and/or `rm` the cluster (kind container) the UI/extension is unable to restart it.

We need to ensure the extension is resilient for this case.

jorgemoralespou:
Creation of cluster now makes sure that deletes stale files or containers in not running state, so it should always create.
It also adds some interval checks to verify underlying status of TCE container, and errors out appropriatelly.

Fixes: #4352 

- Created cluster, left and get back to extension (while creating and after creating)
- Stop underlying container. It's noticed within check interval (5 seconds). If restarted won't notice in the UI unless you get out of the extension and come back. Delete works fine. And can restart.
- Kill underlying container. It's noticed within check interval (5 seconds). Delete works fine and can restart.
- If you get out of the extension, and come back, it prompts to create, and previous exited container is removed succesfully.

Signed-off-by: Jorge Morales Pou <jorgemoralespou@gmail.com>


stmcginnis (approved):
Looks fine to me. There's a bunch of stuff included here that doesn't look like it's related to the main reason for this PR, but looks harmless enough.

Will wait and let @jpmcb take a look before merging.

stmcginnis (approved):

jpmcb (approved):
Nice! Works great!! 

Built and installed using `make extension && make install`. Then created cluster. After creating the cluster, went to the command line and `$ docker kill <container-id>`

Than I see:

Screen Shot 2022-05-06 at 8 39 25 AM

and can delete the cluster, than being prompted to create a new one 👍🏼 

Also able to re-start container and see that the extension picks up on that and reports it's running again. Great work!!

PR #4393, opened 2022-05-06
UC: Raise cluster delete log level

### LINKED ISSUE #4392: Docker Desktop cluster creation output says to use `tanzu uc rm` to delete
=## Bug Report

The end of the output for cluster creation in Docker Desktop includes messages stating to delete the cluster by running `tanzu unmanaged-cluster delete [NAME]`. This is now how clusters are deleted with the Docker Desktop integration.

## Expected Behavior

Cluster deletion steps should not be included in output.

stmcginnis:

On cluster creation we print out steps explaining how to delete the
cluster. This raises the log level on those messages so they are not
printed when verbosity is set lower so things like the Docker Desktop
integration where the default instructions do not apply.

Fixes: #4392

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
+++ b/cli/cmd/plugin/unmanaged-cluster/tanzu/tanzu.go
@@ -371,8 +371,8 @@ func (t *UnmanagedCluster) Deploy(scConfig *config.UnmanagedClusterConfig) (int,
 	log.Style(outputIndent, color.FgGreen).Infof("tanzu package available list\n")
 	log.Infof("View running pods:\n")
 	log.Style(outputIndent, color.FgGreen).Infof("kubectl get po -A\n")
-	log.Infof("Delete this cluster:\n")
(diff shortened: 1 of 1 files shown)


jpmcb (commented):
When deploying an unmanaged cluster via the command line, this logging is removed:

```
❯ tanzu unmanaged-cluster create test

...

View available packages:
   tanzu package available list
View running pods:
   kubectl get po -A

~/workspace/vmware/community-edition/cli/cmd/plugin/unmanaged-cluster
❯   
```

which feels weird from the command line; it'd be nice if these logs still could surface for command line users.

But I guess since the extension is `exec`-ing out to the unmanaged-cluster binary, there isn't a great way to get around this

stmcginnis:
> it'd be nice if these logs still could surface for command line users

Hmm, something odd there. This had been working right - the default command line should still include the level 2 output, then things like the Docker Desktop integration can call with `-v 0` to lower the log level and cause these not to be printed.

Will have to investigate what happened.

jpmcb (approved):
This LGTM after reviewing #4414 👍🏼

PR #3769, opened 2022-03-29
Fixing doc versions for release-0.10

kcoriordan:
Signed-off-by: kcoriordan <koriordan@vmware.com>

## What this PR does / why we need it
**On release-0.10 branch:** Replaced the following shortcode variables with hard coded values because with new doc versioning, by default- anything outside of docs/content will always read from the main branch - which results in wrong version.

{{< release_latest >}}
{{< release_latest_no_v >}}
{{< pkg_repo_latest >}}

## Which issue(s) this PR fixes

Fixes: #3750

PR #4397, opened 2022-05-06
Fixing doc versions

kcoriordan:

Replace the following shortcode variables with hard coded values for 0.12.0  because with current setup, changing them to 0.12.0 in `(docs/site/config.yaml)` would also change it for 0.11.0 branch
```
  release_latest: 
  release_latest_no_v: 
  pkg_repo_latest:
```
  
  
 Create an issue for 0.13.0 to address this so that we can use shortcodes on a per release branch basis rather than the global basis it currently is: https://github.com/vmware-tanzu/community-edition/issues/4399
 
 
 ref: https://github.com/vmware-tanzu/community-edition/pull/3769

Fixes: #

### DIFF
--- a/docs/site/content/docs/edge/assets/cli-install-linux.md
+++ b/docs/site/content/docs/edge/assets/cli-install-linux.md
@@ -25,7 +25,7 @@
 
 ### Option 2: Curl GitHub release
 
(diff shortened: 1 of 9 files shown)


stmcginnis (approved):
Good catch!

PR #4414, opened 2022-05-06
Fix unmanaged cluster log verbosity enforcement

### LINKED ISSUE #4413: Unmanaged cluster log level not being enforced
## Bug Report

As noticed in https://github.com/vmware-tanzu/community-edition/pull/4393#pullrequestreview-964773328 the default
logging level of 2 is not being honored. Logging appears to be at level 0, so higher level log output is being skipped
by default.

## Expected Behavior

Default log level should be 2 and log output should only be skipped if user explicitly passes `-v 0`.

## Steps to Reproduce the Bug

`tanzu uc create foo`

stmcginnis:

There were issues with the way log verbosity was being set which caused
the incorrect level to be used, resulting in some expected lines not
printing at the default log level.

This updates the handling to correctly parse and set the level from the
flag. Also some minor refactoring to make it a little easier to
understand.

Fixes: #4413 

```
go run . create foo
go run . rm foo
go run . create -v 0 foo
go run . rm -v 0 foo
```

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/utils.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/utils.go
@@ -13,6 +13,8 @@ import (
 	"golang.org/x/term"
 )
 
+const defaultLogVerbosity = 2
+
 // TtySetting gets the setting to use for formatted TTY output based on whether
 // the user explicitly set it with a command line argument, or if not, whether
 // there is an environment variable set. If neither of these things, it will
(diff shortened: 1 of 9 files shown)


cli/cmd/plugin/unmanaged-cluster/cmd/utils.go
```diff
@@ -57,6 +59,17 @@ func SetupRootCommand(rootCmd *cobra.Command) {
 	rootCmd.SetHelpTemplate(helpTemplate)
 }
 
+// LoggingVerbosity will get the configured logging level for this command.
+func LoggingVerbosity(cmd *cobra.Command) int {
+	level, err := cmd.Flags().GetInt32("verbose")
```
stmcginnis, line 64: @jpmcb would love to get your cobra expertise to verify this is how this parent flag should be handled. Thanks!

jpmcb (approved):
Nice!!! This looks good to me

PR #4421, opened 2022-05-08
Metrics-server should not create or addopt kube-system namespace.

### LINKED ISSUE #1708: metrics-package should not have createNamespace:true for kube-system
**Bug description**
`resources are adopted by default in kapp. so if you specify a resource in your config and it already exists on the cluster, and is not owned by any app, then it will be adopted`
Since metrics-server defines kube-system as the namespace to use and it has the createNamespace: true, it will then `own` the kube-system namespace. This prevents users from adding labels/annotations to this namespace and also poses a risk as if metrics-server package was removed, then the cluster would no longer work.

**Affected product area (please put an X in all that apply)**

- [x] Addons

**Expected behavior**
`createNamespace` should be defaulted to false, as the `kube-system` namespace already exist and is not to be owned by this package.

Relevant discussion: https://kubernetes.slack.com/archives/C02GY94A8KT/p1645018465370589

**Version** (include the SHA if the version is not obvious)
All

adduarte:
If the metric server creates or addopts the kube-system namespace,
it will be marked for deletion whenever the metric-server package is
deleted. kube-system namespace deletion is not allowed.

Fixes https://github.com/vmware-tanzu/tanzu-framework/issues/1708

### DIFF
--- a/addons/packages/metrics-server/0.5.1/bundle/config/values.yaml
+++ b/addons/packages/metrics-server/0.5.1/bundle/config/values.yaml
@@ -12,7 +12,7 @@ daemonset:
   updateStrategy: null
 metricsServer:
   namespace: null
-  createNamespace: true
+  createNamespace: false
   config:
     securePort: 4443
     updateStrategy: RollingUpdate
--- a/addons/packages/metrics-server/0.4.0/bundle/config/values.yaml
+++ b/addons/packages/metrics-server/0.4.0/bundle/config/values.yaml
@@ -4,7 +4,7 @@
 namespace: kube-system
 metricsServer:
   namespace: null
-  createNamespace: true
+  createNamespace: false
   config:
     updateStrategy: RollingUpdate
     args: [] #! Add any command args here
--- a/addons/packages/metrics-server/0.5.1/bundle/config/schema.yaml
+++ b/addons/packages/metrics-server/0.5.1/bundle/config/schema.yaml
@@ -28,7 +28,7 @@ metricsServer:
   #@schema/nullable
   namespace: kube-system
   #@schema/desc "Whether to create namespace specified for metrics-server"
-  createNamespace: true
+  createNamespace: false
   config:
     #@schema/desc "The HTTPS secure port used by metrics-server"
     securePort: 4443


addons/packages/metrics-server/0.5.1/bundle/config/values.yaml
```diff
@@ -12,7 +12,7 @@ daemonset:
   updateStrategy: null
 metricsServer:
   namespace: null
-  createNamespace: true
+  createNamespace: false
```
shyaamsn, line 15: Couple of scenarios we need to handle:
1. User giving createNamespace: true and namespace: kube-system - In this case lets not create namespace
2. Upgrade: User upgrades the package but has older data values i.e. User in their data values has createNamespace: true. So even though package defaults are changed the user has createNamespace: true and hence during delete it will attempt to delete the namespace. Lets also add `https://carvel.dev/kapp/docs/v0.46.0/apply/#kappk14siodelete-strategy` delete-strategy:orphan in the namespace to prevent deletion of namespace

adduarte: namespace creation is handled by kapp-controller. I don't tink we can do much here to prevent the creation or adoption of the namespace when the user intentionally sets createNamespace: true. 
However adding delete-strategy:orphan to the namespace will mitigate this as well.

adduarte: It looks like this behavior has been fixed in https://github.com/vmware-tanzu/carvel-kapp/pull/468


see discussion here: https://kubernetes.slack.com/archives/C02GY94A8KT/p1645018465370589
There is a request to add a feature to kapp to prevent a package from taking ownership: https://github.com/vmware-tanzu/carvel-kapp-controller/issues/523

and the [fix](https://github.com/vmware-tanzu/carvel-kapp/pull/468) is referenced there. 
looks like fix should be available as of  kapp v0.47.0

adduarte: for deployments that use older kapp-controller the best approach will be to add https://carvel.dev/kapp/docs/v0.46.0/apply/#kappk14siodelete-strategy delete-strategy:orphan, and modify the values.yaml as suggested in this pr

shyaamsn: 1. Lets add https://carvel.dev/kapp/docs/v0.46.0/apply/#kappk14siodelete-strategy delete-strategy:orphan for older kapp-controller deployments
2. Can we add a check here  https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/metrics-server/0.5.1/bundle/config/overlays/overlay-namespace.yaml#L12 
`#@ if data.values.metricsServer.createNamespace and data.values.namespace != "kube-system":` 
To prevent even creating namespace if kube-system is given ?

adduarte: sounds like a good plan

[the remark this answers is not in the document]
adduarte: ack will do

adduarte:
In order to move forward we will scale down this patch to only fix the incorrect settings in metrics-server package bundle. 
A different patch in [tanz-framwork/addons/a](https://github.com/vmware-tanzu/tanzu-framework/tree/main/addons) will be used to correct any deployments which have the kube-system namespace marked with the kapp-controller annotations.

PR #4430, opened 2022-05-09
UI plugin: WC - adds dataPath

swalner-vmware:
This PR adds dataPath to the formReducer and handleValueChange() method. It allows calling components to specify where in the store their data lives.
Note that this removes a fair amount of code from WorkloadClusterUtility


raymondz1 (approved):
LGTM

miclettej (approved):

PR #4433, opened 2022-05-09
remove octant doc

kcoriordan:

Confirmed with @joshrosso awhile back that Octant doc should be removed, removing now.

Fixes: #


jpmcb (approved):
👍🏼 - Looks good to me. I'm guessing this also needs to be backported to `0.12.0`, `0.11.0` and `0.10.0`?

kcoriordan:
@jpmcb 0.12.0, 0.11.0 - yes I think so
really not sure about 0.10.0 not sure when support was removed

PR #4516, opened 2022-05-12
UI plugin: WC - error display for CC variables input

swalner-vmware:


stmcginnis (approved):
Looks fine to me from an API perspective. We won't be able to actually use this for awhile, so no way to test things out on the backend, but doesn't look like there will be any issues when that time comes.

cli/cmd/plugin/ui/web/tanzu-ui/src/views/workload-cluster/select-management-cluster.scss
```diff
@@ -7,3 +7,26 @@
 .header-mc-grid {
   background-color: #222222;
 }
+
+.cc-description-text {
+  font-size: var(--cds-global-typography-font-size-3);
+  color: var(--cds-global-color-white);
+}
+
+.cc-parent-card {
```
miclettej, line 16: I have a css class in index.scss called 'section-raised' that might be worth trying and see if you like how it looks in the context of this UI. I should probably document the usage of that class a little bit, and also open to renaming the class. If you don't like how it looks in this area of the UI, please disregard. 

I would also maybe consider moving the accordion styles to index.scss so that these styles can be consumed in other areas of the app. They could be renamed to drop the 'cc-' prefix if so.

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/models/ClusterClass.ts
```diff
@@ -14,11 +15,42 @@ export interface ClusterClassVariable {
...
+    taxonomy: ClusterClassVariableType,             // field classified according to known types
+    uiClassification: CCUiClassification, // field categorized as where it belongs in the UI
+    description?: string,
+    default?: any,
+    required?: boolean,
+    possibleValues?: string[],
+    children?: CCVariable[],
+}
+
+// for some reason, eslint is reporting these enum values as unused
```
miclettej, line 38: I think this may be caused by us setting 'no-unused-vars' in the eslintConfig in package.json. Can you try changing it to '@typescript-eslint/no-unused-vars'

miclettej (approved):
Just two minor comments then good to go for me.

cli/cmd/plugin/ui/web/tanzu-ui/src/views/workload-cluster/select-management-cluster.scss
swalner-vmware: Works for me

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/models/ClusterClass.ts
swalner-vmware: The `no-unused-vars` comment works as is; my "for some reason" comment is just lamenting that eslint bothers us with an unused enum

miclettej (approved):

PR #4528, opened 2022-05-13
Contour README.md updates

skriss:
Updates the Contour 1.20.1 README.md to
conform to the template.

Signed-off-by: Steve Kriss <krisss@vmware.com>

Updates the Contour 1.20.1 README.md to
conform to the template.

N/A, docs only

N/A

### DIFF
--- a/addons/packages/contour/1.20.1/README.md
+++ b/addons/packages/contour/1.20.1/README.md
@@ -1,35 +1,34 @@
-# Contour Package
+# Contour
 
 This package provides an ingress controller using [Contour](https://projectcontour.io/).
 
-## Supported Providers
-
-The following tables shows the providers this package can work with.
-
-| AWS  |  Azure  | vSphere  | Docker |
-|:---:|:---:|:---:|:---:|
-| ✅  |  ✅  | ✅  | ✅  |
-
-## Components
-
-* Contour controller
-* Envoy reverse proxy and load balancer
+This documentation provides information about the specific TCE package.
+Please visit the [TCE package management page](https://tanzucommunityedition.io/docs/package-management/) for general information about installation, removal, troubleshooting, and other topics.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


addons/packages/contour/1.20.1/README.md
```diff
@@ -48,11 +47,116 @@ The following tables shows the providers this package can work with.
...
+
+An example of the data values file looks like:
+
+```yaml
+namespace: my-custom-contour-namespace
+contour:
+  replicas: 3
+envoy:
+  logLevel: debug
+```
```
skriss, line 150: This section seems pretty redundant given other content, personally I would drop it but it's in the template so I have it here for now.

addons/packages/contour/1.20.1/README.md
```diff
@@ -48,11 +47,116 @@ The following tables shows the providers this package can work with.
...
+
+## Package Limitations
+
+- Not all Contour command-line flags can be customized via data values file. For example, Contour's `--ingress-class-name` flag cannot be specified.
+- The package only supports installing Envoy as a DaemonSet.
```
skriss, line 155: Perhaps move this below "Usage Example"? I have it here since that's how it's ordered in the template.

sunjayBhatia (approved):
LGTM, but yeah agree about the redundant bits

addons/packages/contour/1.20.1/README.md
LappleApple: Hey @skriss, totally cool to say "see other sections in this documentation" or the like to avoid redundancies while keeping the section for consistency/in case users look for it.

skriss: Alright, I think as-is is OK then since it links to the prev section for details and includes a brief example.

skriss:
@LappleApple any other feedback from your side, or is this good to merge?

LappleApple:
@skriss All sections are there, looks good. Thanks for doing this so quickly.

PR #4567, opened 2022-05-19
Fix auto merge part 5

davidvonthenen:

I think I finally (maybe) figured this out. Turns out the dependabot has it's own different set of secrets. Added one in the last attempt, but don't need this explicit login though which is causing the failure.

NA

NA

NA


stmcginnis (approved):
Try try again. ;)

PR #4604, opened 2022-05-24
Updated to knative 1.4.0 and more configuration available

### LINKED ISSUE #3129: [Knative package] Support additional values for knative package
## Feature Request
Add additional configurations to kantive serving package.
currently the package is very minimalistic in its config options. the main flag that would helpful to add is to update the config-network configmap in the knative-serving namespace to allow setting a different naming template (domain-template field) for service FQDNs.
The default value is "{{.Name}}.{{.Namespace}}.{{.Domain}}". this value is not great if we want to do TLS certs and is much easier to use something like "{{.Name}}-{{.Namespace}}.{{.Domain}}".

Currently i have an overlay i append to the package with the following content which could simply be added to the package itself as well and exposed via a data value:

```
#@ load("@ytt:overlay", "overlay")
#@overlay/match by=overlay.subset({"kind":"ConfigMap", "metadata": {"name": "config-network"}})
---
metadata:
  annotations:
    #@overlay/match missing_ok=True
    kapp.k14s.io/update-strategy: skip
data:
  #@overlay/match missing_ok=True
  #@yaml/text-templated-strings
  domain-template: "{{.Name}}-{{.Namespace}}.{{.Domain}}"
```

## Additional context
This would make day0 usage much easier especially for local setups. by adding the "kapp.k14s.io/update-strategy: skip" annotation we can make sure that we dont block users from changing this but it makes for in my opinion a much better default value then the current situation.

jorgemoralespou:
Updates knative-serving package to 1.4.0 and exposes 2 additional sets of configuration:
- domain template pattern
- cluster wide scaling configuration

README.md, values file and schema has been properly updated with new configuration.

Fixes: #3129 

Deployed the generated files in a fresh TCE cluster, and tested scaling and domain template options.
- With min_scale set to 1 knative service never scales down (unless annotated)
- With url_template like `{{.Name}}-{{.Namespace}}.{{.Domain}}`, we get this sample service: http://testknativeapp-testknativeapp.127.0.0.1.sslip.io (I used [failk8s-packages/testknative-app package](https://github.com/failk8s-packages/testknativeapp-package) to test knative)

Signed-off-by: Jorge Morales Pou [jorgemoralespou@gmail.com](mailto:jorgemoralespou@gmail.com)


jorgemoralespou:
@seemiller  I haven't added schema yet, because the current version of the schema don't give any benefit but otherwise complicated things. I did implement them and soon reverted the work, mostly because there's no way to implement null strings. I would rather wait on the schema updates that should come soon (hopefully).
The new readme format is something I'm not familiar with, so either you take a stab or point me into the right direction.

jorgemoralespou:
@seemiller just saw #4616 . Once that one is merged, I will update the docs to match the changes brought via that PR.

seemiller (approved):
I deployed this package to AWS and was able to successfully follow the usage example. Let's get this package merged and we can update the docs in a follow up PR.

PR #4611, opened 2022-05-25
Add FluxCD Helm Controller v0.21.0

hemakshis:
Releasing v0.21.0 of fluxcd-helm-controller
Adding latest version v0.21.0 of fluxcd-helm-controller compatible with source-controller latest version

Create a kind cluster and attach it to TMC.
Submit the metadata.yaml and 0.21.0/package.yaml files
Install the package using the tanzu cli
tanzu package install fluxcd-helm-controller -p helm-controller.fluxcd.community.tanzu.vmware.com -v 0.21.0
/ Installing package 'helm-controller.fluxcd.community.tanzu.vmware.com'
| Getting package metadata for 'helm-controller.fluxcd.community.tanzu.vmware.com'
| Creating service account 'fluxcd-helm-controller-default-sa'
| Creating cluster admin role 'fluxcd-helm-controller-default-cluster-role'
| Creating cluster role binding 'fluxcd-helm-controller-default-cluster-rolebinding'
| Creating package resource
/ Waiting for 'PackageInstall' reconciliation for 'fluxcd-helm-controller'
\ 'PackageInstall' resource install status: Reconciling
Added installed package 'fluxcd-helm-controller'

Added installed package 'fluxcd-helm-controller'


addons/packages/fluxcd-helm-controller/0.17.2/package.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   refName: helm-controller.fluxcd.community.tanzu.vmware.com
   version: 0.17.2
-  releasedAt: 2022-05-01T16:40:55Z
+  releasedAt: "2022-05-01T16:40:55"
```
seemiller, line 8:
```suggestion
  releasedAt: 2022-05-01T16:40:55Z
```

The releasedAt needs to be in this format, otherwise the package repository will not reconcile. It throws the following error:
```
parsing time "2022-05-01T16:40:55" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "Z07:00"
```
If I recall correctly, we had this problem when the 0.17.2 version was originally added. The change was only made here in the PR. To correctly fix this problem, the package yaml needs to be updated in the `package-for-helm-controller` source repository. With the fix made there, when this package is updated in the future and `vendir sync` is run, we will get the correctly updated file. This will be the same fix for the readme file as well.

addons/packages/fluxcd-helm-controller/0.17.2/README.md
```diff
@@ -37,7 +37,7 @@ To install FluxCD helm-controller from the Tanzu Application Platform package re
    For example:
 
     ```shell
-    $ tanzu package install fluxcd-source-controller -p fluxcd-source-controller.community.tanzu.vmware.com -v 0.21.2
+    tanzu package install fluxcd-source-controller -p fluxcd-source-controller.community.tanzu.vmware.com -v 0.21.2
```
seemiller, line 40: The `0.17.2` readme should not reference the `0.21.2` version.

seemiller (changes requested):

addons/packages/fluxcd-helm-controller/0.17.2/package.yaml
hemakshis: I have resolved this. The issue was the assets under `v0.17.2` were still having the old format and caused this issue.

addons/packages/fluxcd-helm-controller/0.17.2/README.md
hemakshis: It is referencing to the `fluxcd-source-controller:v0.21.2` which is compatible with the `fluxcd-helm-controller:v0.17.2` so the README is correct in that sense

seemiller (approved):

hemakshis:
@seemiller I am unable to merge this PR. All checks are passed but still it shows "Some checks haven't completed yet".

PR #4632, opened 2022-05-26
UI Plugin: Upgrade to React v18

miclettej:
Upgrades UI plugin to React 18 and adjusts Index.tsx to use react-dom render in new format

Fixes: #

App builds and runs as expected


raymondz1 (approved):
lgtm

PR #4700, opened 2022-06-02
UI plugin: CC - proxyServer and dynamic group validation

swalner-vmware:
Adds proxy server validation (format) and dynamic group validation (for optional groups)


miclettej:
Nice I like the proxy server validation

miclettej (approved):

Issue #4701, opened 2022-06-02
UI plugin: feature flags for Workload and Unmanaged Clusters


swalner-vmware:
To promote releasable code, move the related UI elements for workload clusters (and unmanaged clusters) behind feature flags. This will allow the management cluster work to be released independently of when the backend supports workload clusters (and unmanaged clusters).

PR #4702, opened 2022-06-02
UI plugin: feature flags for workload clusters and unmanaged clusters

swalner-vmware:
Fixes: #4701


miclettej (approved):
As mentioned offline, I think this solution works just fine for now. An enhancement would be to set a default state in the app store with enabled/disabled features. It's not a must-have and approving the PR to give you the option to leave as-is.

raymondz1 (approved):
lgtm

PR #4709, opened 2022-06-03
Fix doc issue link and remove edit option

### LINKED ISSUE #4672: [Docs] Adjust the issue template form when open issues directly via docs
## Description

When selecting **REPORT ISSUE** (see screenshot below) on TCE docs, users will automatically be forwarded to a templated issue-form on Github. The prefabricated issue-title can be confusing though. 

Example: `[Docs] [](/docs/v0.12/package-readme-cert-manager-1.8.0/)` 

Basically, the idea is good to have something in the title which indicates the location where you are on docs when choosing to open an issue on Github. However, this will lead to way too long issue-titles when users understand that it's intended to put the actual title into the brackets `[]`. Rather than appending docs-url to the title, we should automatically put it into the e.g. `Location` section within the issue.

## Location

From my description above, we probably should put e.g. `/docs/v0.12/package-readme-cert-manager-1.8.0/` in here.

## Additional Context

Screenshot:

image

stmcginnis:

The docs pages have a "Report Issue" link in the right nav to quickly
get someone over to the docs template for filing any issues they see in
the documentation. This had been attempting to put a link in the title
based on the title of the page, but for most of our pages this ends up
being an empty value.

This updates the query to just put the relative link to the page
directly in the issue title, and puts a placeholder to try to nudge the
user to add an actual description to it.

While looking at this it was also noticed that we had an "Edit" link
there that would attempt to use the GitHub UI editor to be able to make
changes and submit a new PR. We had disabled this ability, so the link
would actualy bring the user to a GitHub 404 page.

Since the ability to file an issue is sufficient, to address the Edit
404 the link is just removed.

Fixes: #4672 

Made edits and ran `hugo` locally to verify rendered HTML looked correct and link works.

### DIFF
--- a/docs/site/themes/template/layouts/partials/docs-right-bar.html
+++ b/docs/site/themes/template/layouts/partials/docs-right-bar.html
@@ -1,15 +1,14 @@
 {{ if .Site.Params.docs_right_sidebar }}
 <div class="right-nav" id="right-nav">
     <div class="right-nav-content">
-        <ul class="buttons">
+	<ul class="buttons">
             {{ if (or .IsNode .IsPage) }}
-                {{ $issueBody := printf "[Docs] [%s](%s)" .Title .RelPermalink | htmlEscape }}
-                {{ $issueQuery := (querify "" $issueBody) }}
-                <li><a href="{{ $.Site.Params.github_base_url }}/issues/new?assignees=&labels=kind%2Fdocs%2C+triage%2Fneeds-triage&template=documentation.md&title{{ $issueQuery | safeURL }}" target="_blank">Report Issues</a></li>
-                {{ $editQuery := (querify "description" "Signed-off-by: NAME <EMAIL_ADDRESS>\n\n") }}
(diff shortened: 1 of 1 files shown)


jorgemoralespou:
It would be good, as I've seen on some other GH issue templates if we could modify the GH issue template to point that the permalink might have been copied in the Issue name if coming from the documentation site directly, so that the user can use the value there to fill the expected value, e.g.:

```
## Description

<!-- Please add a detailed description of the issue or suggestion -->

## Location

<!-- If this is related to an existing page, please include a link to the current page so we can make sure we know where to look. If you are click on reporting an issue from the tanzu-community-edition documentation site, the location might be already present in the Issue name, copy that value from there in here and remove it from the issue name. -->

## Additional Context

<!-- Any additional context that would be useful -->
```

stmcginnis:
> the location might be already present in the Issue name, copy that value from there in here and remove it from the issue name

I don't think we want them to remove it from the title. Maybe say something like they can skip this location section if it's already present in the title?

davidvonthenen (approved):

PR #4714, opened 2022-06-06
Update installation documentation for unmanaged-cluster

marckhouzam:

1. Typos
2. ARM64 is not supported (nor experimental) for Windows
3. Match requirements table for Windows with the managed cluster req table

Ran with `hugo server` and looked at http://localhost:1313/docs/edge/getting-started-unmanaged/

### DIFF
--- a/docs/site/content/docs/edge/assets/prereq-unmanaged-windows.md
+++ b/docs/site/content/docs/edge/assets/prereq-unmanaged-windows.md
@@ -4,7 +4,6 @@
 
 Tanzu Community Edition supports two cluster providers for unmanaged clusters: Kind and minikube
 
-- Kind is the default cluster provider and is included as default with the unmangaged cluster binary, you just need to install Docker.
+- Kind is the default cluster provider and is included as default with the unmanaged cluster binary, you just need to install Docker.
 - minikube is an alternative cluster provider, if you plan to use minikube as your cluster provider, you must first install minikube and a minikube supported container or virtual machine manager such as Docker.
 
-Note: ARM64 support is experimental. Some packages might not install due to their ARM64 image not being available.
--- a/docs/site/content/docs/edge/assets/prereq-unmanaged-linux.md
+++ b/docs/site/content/docs/edge/assets/prereq-unmanaged-linux.md
@@ -4,5 +4,5 @@
 
 Tanzu Community Edition supports two cluster providers for unmanaged clusters: Kind and minikube
 
-- Kind is the default cluster provider and is included as default with the unmangaged cluster binary, you just need to install Docker.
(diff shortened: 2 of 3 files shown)


docs/site/content/docs/edge/assets/prereq-unmanaged-windows.md
```diff
@@ -1,10 +1,9 @@
-| Architecture   | CPU | RAM  | Required software |
-|:---------------|:----|:-----|:------------------|
-| x86_64 / amd64 | 1   | 2 GB | [Docker Desktop](https://www.docker.com/get-started), (Optional) [minikube](https://minikube.sigs.k8s.io/docs/start/) |
+| CPU | RAM  | Required software |
+|:----|:-----|:------------------|
+| 1   | 2 GB | [Docker Desktop](https://www.docker.com/get-started), (Optional) [minikube](https://minikube.sigs.k8s.io/docs/start/) |
```
stmcginnis, line 3: It might be good to keep that Architecture column to keep it explicit that only `x86_64/amd64` architectures are supported.

marckhouzam: I removed it because the Windows instructions for Managed Clusters don't have it. 

But I can add it back.

dvonthenen: Yea, there might be (going to bring this up in a second) that we support arm64 and we should make it explicit

marckhouzam: Do you want me to do the opposite and also add the architecture column to the Managed Clusters page?

stmcginnis: Let's just keep it here for now. With managed clusters it gets a little trickier because we need to distinguish deploying-from and deploying-to. With unmanaged clusters it's a little more straight forward, so I think having the column here makes it a little more clear.

My 2 cents at least.

jpmcb: > It might be good to keep that Architecture column to keep it explicit that only x86_64/amd64 architectures are supported.

I agree - we should keep arch explicit here since it is supported on arm for M1 macs

jpmcb (commented):
Thanks for always catching my mis-spellings Marc! 👏🏼 😆 😭

marckhouzam:
Ok, I added back the architecture column for the reqs table.
The changes are some minor typos and removing the sentence implying we have experimental support for ARM64 on **Windows**

marckhouzam:
I converted this PR to a draft.  As I continue ramping up I'm finding other minor typos that I might as well include here.

PR #4760, opened 2022-06-11
GIT-4729: enable io.Writer based logger

### LINKED ISSUE #4729: Unmanaged-cluster: write logs to `io.Writer` instead of hard coded stdout
## Feature Request

Currently, stdout is hard coded as the default loging location for the unmanaged-cluster writer.

https://github.com/vmware-tanzu/community-edition/blob/aba53f70df579a69e5e56e93793fae55be8f5a59/cli/cmd/plugin/unmanaged-cluster/log/log.go#L114-L125

We should evaluate and attempt to implement using some `io.Writer` so that consumers of the `unmanaged-cluster` packages as code aren't forced to log to stdout. This would also enable passing `ioutil.Discard` to throw away logs if desired.

## Additional context

This would be useful for the Docker Desktop extension consumer of `unmanaged-cluster`

harshanarayana:
Currently the `CMDLogger` hardcodes the `output` to `os.Stdout`. Which means the consumers can't customize the logging behavior by either capturing it into a buffer or by discarding the logs by pointing that to `ioutil.Discard`. 

This PR enables that behaviour in two ways. 

1. Provides a new Logger instance creator that accepts an `io.Writer` to configure the `output` accordingly
2. A `Writer(w io.Writer)` method of the `Logger` interface that can be used to programatically switch the `output` of the logger at different stages of the workflow

Fixes: #4729 

UTs have been included for the new feature in question

Introducing the `defaultWriter` attribute under the `CMDLogger` to be able to handle the `AddLogFile` to make sure the `io.MultiWriter` will always honor the custom `io.Writer` configured and not mutate if the `AddLogFile` is invoked multiple times from different point in workflow. (This is something I am not entirely sure about from a thread safety perspective. I think we need to lock the call to `AddLogFile` with a mutex may be ?)

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/log/log_test.go
+++ b/cli/cmd/plugin/unmanaged-cluster/log/log_test.go
@@ -1,0 +1,41 @@
+package log
+
+import (
+	"bufio"
+	"bytes"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestNewLoggerWithWriter(t *testing.T) {
+	var b bytes.Buffer
+	writer := bufio.NewWriter(&b)
+	logger := NewLoggerWithWriter(false, 5, writer)
+	logger.Info("Generating Message in Info")
+	logger.Info("Generating Message in Info again")
+	writer.Flush()
+	lineCount := len(strings.Split(strings.Trim(b.String(), "\n"), "\n"))
+	if lineCount != 2 {
+		t.Errorf("Expected 2 lines to be captured into buffer but found %d", lineCount)
+	}
+}
+
+func TestNewLogger(t *testing.T) {
+	var b bytes.Buffer
+	writer := bufio.NewWriter(&b)
+	logger := NewLogger(false, 5)
+	logger.Writer(writer)
+	logger.Info("Generating Message in Info")
+	logger.Info("Generating Message in Info again")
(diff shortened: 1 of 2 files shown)


cli/cmd/plugin/unmanaged-cluster/log/log_test.go
```diff
@@ -0,0 +1,41 @@
+package log
```
stmcginnis, line 1: For some reason the linter output isn't showing it, but the issue is this file needs the standard copyright/license header. You should be able to just copy the header from another file.

harshanarayana: Yes sir. I'm in the process of doing that.. Missed it the first time around

harshanarayana: Changes have been updated. @stmcginnis  PTAL

stmcginnis (approved):
Great side benefit of being able to unit test some of the logger using a buffer!

This looks good to me. I have NOT compiled and run to test it all out, but just based on code inspection everything looks good to me. Will wait to merge until either I get a chance to fully test, or someone else gets a chance to also review.

jpmcb:
Thanks for reviewing and merging this Sean! Great value add!! Thanks so much for this!! 👍🏼

PR #4787, opened 2022-06-14
UI plugin: vSphere MC credentials osImage check

swalner-vmware:


cli/cmd/plugin/ui/web/tanzu-ui/node-server/src/routes/api/endpoints/vsphere.js
```diff
@@ -158,16 +158,24 @@ router.get('/resourcepools', (req, res) => {
  * Mock route for getting VC os images
  */
 router.get('/osimages', (req, res) => {
-    winston.info('Mock UI FETCH DATACENTERS (#' + mockOsImageRequestCounter + ')');
-    let osImageResponse = [];
-    if (mockOsImageRequestCounter > 0) {
-        osImageResponse = readFile('provider-vsphere-osimages.json');
+    const dc = '' + req?.query?.dc;
```
raymondz1, line 161: do we need this empty string?

swalner-vmware: I believe so, because otherwise `dc` can be undefined

swalner-vmware: I removed it; I like it better without and whatever problem I was having before isn't there :)

miclettej (approved):
Looks great, thanks for the updates

PR #4790, opened 2022-06-14
Remove ioutil usage

stmcginnis:

A few instances of using the deprecated ioutil module have snuck in.
This updates those to use their proper io or os replacements.

Also added a simple check to the lint job so we catch and prevent these
from happening again.

https://go.dev/doc/go1.16#ioutil

Ran updated `make lint` target without changing source files and verified it failed from detecting `ioutil` usage.
Updated source files and ran `make lint` again to verify it was happy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,13 @@ lint: tools verify-modules
 			$(MAKE) lint || exit 1; \
 			cd $$working_dir; \
 		fi; \
-	done
+	done; \
+	CHECK=$$(grep -r --include="*.go" ioutil ./); \
+	if [ -n "$${CHECK}" ]; then \
+		echo "ioutil is deprecated, use io or os replacements"; \
+		echo "$${CHECK}"; \
+		exit 1; \
+	fi
 
 mdlint:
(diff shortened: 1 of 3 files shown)


davidvonthenen (approved):

Makefile
```diff
@@ -181,7 +181,13 @@ lint: tools verify-modules
 			$(MAKE) lint || exit 1; \
 			cd $$working_dir; \
 		fi; \
-	done
+	done; \
+	CHECK=$$(grep -r --include="*.go" ioutil ./); \
```
jpmcb, line 185: Would it be possible to add [the `ioutilDeprecated` check](https://go-critic.com/overview.html#ioutildeprecated) to our golangci-lint configuration vs writing our own check?

stmcginnis: Thanks! Yeah, I've tried poking around it a bit and really have no idea why it's not failing today. According to the config, it should be running that check and flagging it. And I vaguely remember it actually doing so either here or in `tanzu-framework`, but it appears neither one is catching it anymore.

PR #4793, opened 2022-06-15
Adding fluxcd source controller 0.24.4+update.1

chaudharyalka:
In this version, following fixing are getting added:-
Fixing OpenApi schema for v0.24.4
Removing alpine image by distroless base image

manually tested


seemiller (changes requested):
@chaudharyalka I was unable to install this package to a cluster. The version that you chose, `0.24.4.1` is not a valid Semantic Version and kapp-controller fails when trying to load a registry containing this. I would recommend using a version like `0.24.4-2`

```
Invalid value: "0.24.4.1": must be valid semver:
      Invalid character(s) found in patch number "4.1" (reason: Invalid). Reconcile failed: Deploying: Error (see .status.usefulErrorMessage for details)
```

I would also recommend making these updates in the existing `addons/packages/fluxcd-source-controller/0.24.4` directory.

seemiller (approved):

Issue #4780, opened 2022-06-14
Verify that the imagelint action can properly detect Alpine images


seemiller:
## Bug Report

There is the possibility that the imagelint action is not properly detecting images that utilize an Alpine base image. See the fluxcd-source-controller 0.24.4 package as an example.

## Expected Behavior

When run, the imagelint process should properly detect if an image Alpine based.

## Environment Details

* Build version (`tanzu version`): 
* Deployment (Managed/Unmanaged cluster): 
* Infrastructure Provider (Docker/AWS/Azure/vSphere): 
* Operating System (client):

harshanarayana:
@seemiller  @stmcginnis   I am not sure if this is a problem. Isn't the `fluxcd-source-controller-bundle` coming from https://github.com/vmware-tanzu/package-for-source-controller/blob/main/packages/fluxcd-source-controller/0.24.4/Dockerfile

This is not running with the same base image that upstream `fluxcd/source-controller` is running. It is using `alpine` there vs the `bullseye` here in the tanzu Org. 

```bash
FROM harbor-repo.vmware.com/dockerhub-proxy-cache/library/debian:bullseye

ARG TARGETPLATFORM
```

Pardon my ignorance or lack of understanding if this is wrong.

harshanarayana:
```bash
❯ make imagelint                                                                                                                                                                                                                                                     80%  ▓▒░
User given Path /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition
Total number of images to process: 1
Result 1 of 1 images
Status:   Fail
Image:    fluxcd/source-controller:v0.25.6
Message:  alpine image
-------------------------------  Fail  Summary  ---------------------------------------------
Total  Fail  : 0
-------------------------------  Pass  Summary  ---------------------------------------------
Total  Pass  : 0
-------------------------------  Pull Failed  Summary  ---------------------------------------------
Total  Pull Failed  : 0
-------------------------------  Not Identified  Summary  ---------------------------------------------
Total  Not Identified  : 0
-------------------------------  Summary  -----------------------------------------------
Total  Images                : 1
Total  Occurrences           : 0
Total  Pass                  : 0
Total  1  file(s) contain(s) this image
File Path: /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition/addons/packages/fluxcd-source-controller/0.24.4/package.yaml
Image Position: 44 : 6

Total  Fail                  : 0
[...]

PR #4794, opened 2022-06-15
GIT-4780: enhance image linter validation behavior

harshanarayana:
This PR changes the behavior of how the `imagelint` Make target validates the images for `alpine` base Images. The PR  makes the existing checks a bit more robust  by  changing a few things of how the validation is done.

Not all containers will have the `ENTRYPOINT` configured by default or will have an `ENTRYPOINT` that can just work when invoked with `docker run`.

One could suggest we can simple `docker run --entrypoint cmd` to customize this. However, there is no guarantee that it will work either. It very much depends on what the base image is. This can all lead to false negative/positive results. 

Instead, `docker` has a wonderful ability to just `create` a container and them copy the artifacts from it without every having to actually `run` it. This bypasses all the worry about `ENTRYPOINT` commands. We can simple create container and copy the required file to perform the validation. This also save a good amount of compute resources that would be taken up by the running containers if any.

Fixes: #4780 

```bash
❯ make imagelint                                                                                                                                                                                                                                                     80%  ▓▒░
User given Path /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition
Total number of images to process: 1
Result 1 of 1 images
Status:   Fail
Image:    fluxcd/source-controller:v0.25.6
Message:  alpine image
-------------------------------  Fail  Summary  ---------------------------------------------
Total  Fail  : 0
-------------------------------  Pass  Summary  ---------------------------------------------
Total  Pass  : 0
-------------------------------  Pull Failed  Summary  ---------------------------------------------
Total  Pull Failed  : 0
-------------------------------  Not Identified  Summary  ---------------------------------------------
Total  Not Identified  : 0
-------------------------------  Summary  -----------------------------------------------
Total  Images                : 1
Total  Occurrences           : 0
Total  Pass                  : 0
Total  1  file(s) contain(s) this image
File Path: /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition/addons/packages/fluxcd-source-controller/0.24.4/package.yaml
Image Position: 44 : 6

Total  Fail                  : 0
Total  Pull Failed           : 0
Total  Not Identified        : 0
[Note:All totals for Pass|Fail|Not Identified|Pull Failed are based on number of occurrences]
--------------------------------------------------------------------------------------------

exit status 1
make: *** [imagelint] Error 1
```

```bash
User given Path /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition
Total number of images to process: 1
Result 1 of 1 images
Status:   Pass
Image:    projects.registry.vmware.com/tce/fluxcd-source-controller-bundle@sha256:4dc37bdc0a6188137905639cccd1938edb1ae882698b3d58a24615b44eb1fa3f
Message:  According to the image history, this is not an Alpine Image
Total  1  file(s) contain(s) this image
File Path: /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition/addons/packages/fluxcd-source-controller/0.24.4/package.yaml
Image Position: 44 : 6

-------------------------------  Fail  Summary  ---------------------------------------------
Total  Fail  : 0
-------------------------------  Pass  Summary  ---------------------------------------------
Image:     projects.registry.vmware.com/tce/fluxcd-source-controller-bundle@sha256:4dc37bdc0a6188137905639cccd1938edb1ae882698b3d58a24615b44eb1fa3f
File Path: /Users/harshnar/Work/Tesseract/Maglev/mks/k8s/community-edition/addons/packages/fluxcd-source-controller/0.24.4/package.yaml
Image Position: 44 : 6

Total  Pass  : 1
-------------------------------  Pull Failed  Summary  ---------------------------------------------
Total  Pull Failed  : 0
-------------------------------  Not Identified  Summary  ---------------------------------------------
Total  Not Identified  : 0
-------------------------------  Summary  -----------------------------------------------
Total  Images                : 1
Total  Occurrences           : 1
Total  Pass                  : 1
Total  Fail                  : 0
Total  Pull Failed           : 0
Total  Not Identified        : 0
[Note:All totals for Pass|Fail|Not Identified|Pull Failed are based on number of occurrences]
--------------------------------------------------------------------------------------------
```

NA

### DIFF
--- a/hack/check/imagelinter/main.go
+++ b/hack/check/imagelinter/main.go
@@ -129,16 +129,6 @@ func LintAll(imc *imglint.ImageLintConfig, wrapper *imgwrapper.Wrapper, key stri
 		}
 	}
 
-	// Check the docker history
-	skip, err := wrapper.Validate(imc.SuccessValidators)
-	if skip && err == nil {
-		return &CheckResult{
-			Image:   key,
-			Status:  imglint.Pass,
-			Message: "According to the image history, this is not an Alpine Image",
-		}
-	}
-
 	_, err = wrapper.CreateContainer()
 	if err != nil {
 		// May be a local problem, but we can't tell
@@ -180,6 +170,16 @@ func LintAll(imc *imglint.ImageLintConfig, wrapper *imgwrapper.Wrapper, key stri
 		}
 	}
 
+	// Check the docker history
+	skip, err := wrapper.Validate(imc.SuccessValidators)
+	if skip && err == nil {
+		return &CheckResult{
+			Image:   key,
+			Status:  imglint.Pass,
+			Message: "According to the image history, this is not an Alpine Image",
+		}
+	}
+
 	return &CheckResult{
 		Image:   key,
 		Status:  imglint.Pass,
@@ -191,7 +191,8 @@ func LintAll(imc *imglint.ImageLintConfig, wrapper *imgwrapper.Wrapper, key stri
 // otherwise returns nil if either it is not alpine or if it could not be determined.
 func IsAlpine(wrapper *imgwrapper.Wrapper, key string) error {
 	// Find the os-release file in the container
-	localName := fmt.Sprintf("./os-release-%s", key)
+
+	localName := fmt.Sprintf("./os-release-%s", wrapper.Container)
 	_, err := wrapper.ContainerCP("/etc/os-release", localName)
 	if err != nil {
 		_, _ = wrapper.ContainerCP("/usr/lib/os-release", localName)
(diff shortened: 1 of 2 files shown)


harshanarayana:
> But this check is still showing it is not:

This image is actually different from the upstream one is it not ? The image layers of the upstream `fluxcd/source-controller` vs the one used by the repo `projects.registry.vmware.com/tce/fluxcd-source-controller-bundle` are entirely different if you check the layers. 
Screenshot 2022-06-15 at 6 51 33 PM


The tool did not flag it as Alpine because there is nothing indicating that in the actual image being inspected.

harshanarayana:
> I don't think the checks being done in Validate are good enough to skip further checks.

This I agree. I think it is better to do the reverse of the checks to start with. Check if the `**/os-release` is not indicating it is Alpine and if it says so, stop right there before even checking anything else.  Only in case if the `os-release` is showing as non Alpine do we need to go further.

PR #4797, opened 2022-06-15
[E2E] Used tanzu lib to create and delete UC in UC e2e-test

### LINKED ISSUE #4110: `unmanaged-cluster` e2e tests fail if `tanzu` CLI not installed / found
## Bug Report

Running the `unmanaged-cluster` e2e tests will fail if the user does  not have `tanzu` installed or available.

Since `unmanaged-cluster` is fully runnable and seperate from `tanzu`, users may not have the full `tanzu` suite on their local workstation. This is a frequent workflow I take. Further, running `go build` for just unmanaged-cluster is much much faster than re-compiling and re-initilizing the entire `tanzu` suite

## Expected Behavior

Run and complete e2e tests without requiring the `tanzu` parent CLI

## Steps to Reproduce the Bug

1. From the root of the tce project, clean everything up with `make clean`
2. `cd cli/plugin/unmanaged-cluster`
3. `go test ./...`

Looks like this is happen from:
https://github.com/vmware-tanzu/community-edition/blob/d4f8a24f55b82ebf56cf8c22ac0264b1135f7501/cli/cmd/plugin/unmanaged-cluster/test/e2e/utils/cli.go#L63

aman556:
Signed-off-by: Aman Sharma <amansh@vmware.com>

This PR is using the tanzu library in Unmanaged cluster plugin to create and delete UC in UC e2e-test. In this way the creation and deletion of cluster become independent of tanzu cli.

Fixes: #4110 

Ran `make unmanaged-cluster-e2e-test` from root.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/test/e2e/utils/cli.go
+++ b/cli/cmd/plugin/unmanaged-cluster/test/e2e/utils/cli.go
@@ -10,6 +10,8 @@ import (
 	"log"
 	"os"
 	"os/exec"
+
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/tanzu"
 )
 
 func InstallTCE() error {
@@ -75,4 +77,13 @@ func cliRunner(name string, input io.Reader, args ...string) (string, error) {
 	}
 
 	return stdOut.String(), err
+}
+
+func ContainsUC(ucLists []tanzu.Cluster, e string) bool {
+	for _, uc := range ucLists {
+		if uc.Name == e {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


cli/cmd/plugin/unmanaged-cluster/test/e2e/utils/cli.go
```diff
@@ -6,6 +6,7 @@ package e2e
 import (
 	"bytes"
 	"fmt"
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/tanzu"
```
stmcginnis, line 9: This should be moved down into its own local import group.

stmcginnis (commented):
Looks great, just minor linting issues to fix.

cli/cmd/plugin/unmanaged-cluster/test/e2e/utils/cli.go
aman556: Hey @stmcginnis Thanks for the review I am fixing the linting issue. Just a min.

stmcginnis (approved):
Awesome!

PR #4822, opened 2022-06-17
correct setting of coreDNSIP in kapp-controller package

cppforlife:

in v0.38.1 kapp-controller package:
- overlays were not correctly updated to account for kapp-controller now having 2 containers
- image was not updated to the correct digest

tested manually via `ytt -f bundle/config | kapp deploy -a kc -f- -c`

also includes a change (chmod) that makes cert reloading compatible with how kapp-controller does cert reload.

(im assuming PackageCR needs to be updated as well. not sure if we already released this version of PackageCR into the wild.)

Fixes: #


cppforlife:
cc @joe-kimmel-vmw @joaopapereira

joe-kimmel-vmw (approved):

PR #4854, opened 2022-06-21
Unmanaged-cluster: Protect cli against extra arguments

### LINKED ISSUE #4842: Unmanaged-cluster: Extra arguments are not being checked
## Bug Report

Some of the unmanaged-cluster sub-commands do not check for too many arguments at the command-line.  This can cause unexpected behaviour for the cli.  Take the `tanzu uc stop` command as an example.  A user may wonder if the command allows to stop multiple clusters at once (it does not allow it) and try:
```
$ tanzu uc stop cluster1 cluster2

🧪 Attempting to stop cluster:
Failed to stop cluster. Error: cluster name is required
```
Notice that the presence of the second argument causes an unexpected behaviour where the cli attempts to stop an unnamed cluster.

## Expected Behavior

The user should be informed that the number of arguments is wrong.

## Steps to Reproduce the Bug

Always reproducible.
```
$ tanzu uc stop cluster1 cluster2

🧪 Attempting to stop cluster:
Failed to stop cluster. Error: cluster name is required
```

## Environment Details

Happening on the current main branch.

marckhouzam:

Protects against unexpected behaviour of the unmanaged-cluster cli when too many arguments are specified.

Fixes #4842

I called unmanaged-cluster sub-command with zero, one and two arguments and verified arguments were properly checked.
For the `uc create` command, I also checked each combo with and without the `-f` and `-e` flags.
Below is a script that can be sourced that tests all combinations:
```
for cmd in config delete start stop;do
  echo \$ uc $cmd
  uc $cmd 2>&1 |head -3
  echo "============"
  echo \$ uc $cmd c1
  uc $cmd c1 2>&1 |head -3
  echo "============"
  echo \$ uc $cmd c1 c2
  uc $cmd c1 c2 2>&1 |head -3
  echo "============"
done

echo \$ uc list
uc list
echo "============"
echo \$ uc list extra_arg
uc list extra_arg 2>&1 |head -3
echo "============"

echo \$ uc create
uc create 2>&1 |head -3
echo "============"
echo \$ uc create c1 -p p
uc create c1 -p p
echo "============"
echo \$ uc create c1 c2
uc create c1 c2 2>&1 |head -3
echo "============"
echo \$ uc create -f test
uc create -f test
echo "============"
echo \$ uc create -e test
uc create -e test
echo "============"
echo \$ uc create c1 -f test
uc create c1 -f test
echo "============"
echo \$ uc create c1 -e test
uc create c1 -e test
echo "============"
echo \$ uc create c1 c2 -f test
uc create c1 c2 -f test 2>&1 |head -3
echo "============"
echo \$ uc create c1 c2 -e test
uc create c1 c2 -e test 2>&1 |head -3
echo "============"
```

Before `tanzu uc list extra` would work but now it won't.  This is technically not backwards-compatible.

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/configure.go
@@ -26,6 +26,7 @@ var ConfigureCmd = &cobra.Command{
 	Short:   "Generate a config file to be used in cluster creation",
 	Long:    configureDesc,
 	RunE:    configure,
+	Args:    checkSingleClusterArg,
 }
 
 //nolint:dupl
@@ -49,14 +50,8 @@ func init() {
 }
 
 func configure(cmd *cobra.Command, args []string) error {
-	var clusterName string
-
-	// validate a cluster name was passed
-	if len(args) < 1 {
-		return fmt.Errorf("cluster name not specified")
-	} else if len(args) == 1 {
-		clusterName = args[0]
-	}
+	// args have already been checked by ConfigureCmd.Args()
+	clusterName := args[0]
 
 	log := logger.NewLogger(TtySetting(cmd.Flags()), LoggingVerbosity(cmd))
 
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/create.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"errors"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -72,7 +73,15 @@ var CreateCmd = &cobra.Command{
 	Short: "Create an unmanaged cluster",
 	Long:  createDesc,
 	Run:   create,
-	Args:  cobra.MaximumNArgs(1),
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 1 {
+			return errors.New("only a single cluster name should be specified")
+		}
+		if len(args) == 0 && co.clusterConfigFile == "" && co.existingClusterKubeconfig == "" {
+			return errors.New("must specify a cluster name or configuration file")
+		}
+		return nil
+	},
 }
 
 var co = createUnmanagedOpts{}
--- a/cli/cmd/plugin/unmanaged-cluster/cmd/delete.go
+++ b/cli/cmd/plugin/unmanaged-cluster/cmd/delete.go
@@ -4,7 +4,6 @@
 package cmd
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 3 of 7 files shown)


stmcginnis (dismissed):
> I like the custom message which provides useful info in this case.

I agree, it is nice being able to provide useful feedback.

Maybe a good feature request for cobra:

```
func MaximumNArgs(n int, tooManyErrorMsg string) PositionalArgs {}
```

jpmcb (dismissed):
Nice! This looks great 👏🏼

harshanarayana:
> Maybe a good feature request for cobra:

Is this not already kind of customisable ?

```golang
var CreateCmd = &cobra.Command{
	Use:   "create <cluster name>",
	Short: "Create an unmanaged cluster",
	Long:  createDesc,
	Run:   create,
	Args: func(cmd *cobra.Command, args []string) error {
		if len(args) == 0 {
			return errors.New("Error: A cluster name is required")
		}
		if len(args) != 1 {
			return errors.New("Error: must specify one cluster name to create a cluster")
		}
		return nil
	},
}
```

marckhouzam:
> Is this not already kind of customisable ?
[...]
> Can we not this way for all of the unmanaged-cluster CLI commands and standardize the error messages and validation behaviours ?

Good point @harshanarayana, it would be more elegant to use Cobra's `Args` field to do argument validation. Thanks!  I can update the PR that way.

Can I make a similar change for the `create` command @stmcginnis ?  That way we align all the commands.  If you agree, do you prefer a different PR or can we put everything here?

stmcginnis:
> That way we align all the commands

Great idea! I'd love to see this done in a consistent way across commands. Up to you whether you want to add it to this PR or do another - I'm good either way.

marckhouzam:
I've updated the PR taking all recommendations and suggestions into account.
I've updated the description of the PR to match, which includes a little script that tests all the different combinations of modified commands and arguments.

Finally, there is a potential backwards-compatibility issue as `tanzu uc list extra` used to pass but now it will reject the useless `extra` argument.  I'm not familiar with TCE's backwards-compatibility policy, so if someone can let me know if I should revert that part of the change, I would appreciate.

jpmcb (dismissed):
> Finally, there is a potential backwards-compatibility issue as tanzu uc list extra used to pass but now it will reject the useless extra argument

That's fine. It was invalid behavior before so therefore a bug which wouldn't be supported in older versions.

This looks good to me!

jpmcb:
@marckhouzam - can you squash to one commit?

harshanarayana (commented):
This one looks great now. Thanks @marckhouzam

marckhouzam:
Squashed.
Thanks for the quick review.

stmcginnis (approved):
Looks great to me.

PR #4868, opened 2022-06-22
UI: Explicitly set default CAPI providers

stmcginnis:

This only modifies the AWS management cluster workflow for now to
validate this is the right approach.

This gets the default providers from the BOM file and passes that
through to the management cluster creation workflow.


swalner-vmware (approved):
Looks fine to me, although I don't understand the context well enough to understand where the CoreProvider, BootstrapProvider and ControlPlaneProvider were coming from before this change

PR #4871, opened 2022-06-22
Add dynamic shell completion for unmanaged-cluster

marckhouzam:

This adds dynamic shell completion of cluster names for the different commands of unmanaged-cluster.
It also disables file completion for the cases where file names are not accepted.

Dynamic completion for flag values is not handled yet.

Related to https://github.com/vmware-tanzu/tanzu-framework/issues/701

To test you should have unmanaged clusters available:
```
# Create a Running cluster
./unmanaged-cluster create c1

# Create a stopped cluster
./unmanaged-cluster create c2 && ./unmanaged-cluster stop c2

# Create a cluster that will have an Unknown status which will be named kind-c1
./unmanaged-cluster create c3 -e $HOME/.config/tanzu/tkg/unmanaged/c1/kube.conf
```

Setup completion for zsh:
```
autoload -Uz compinit;compinit # just in case completion was never setup

source <(./unmanaged-cluster completion zsh)
compdef _unmanaged-cluster unmanaged-cluster
```
or setup completion for bash:
```
# You need to have installed the bash-completion v2 package if you haven't already
source <(./unmanaged-cluster completion bash)
```
Now run some completion tests
```
# Completion of all cluster names
./unmanaged-cluster delete <tab>

# Completion of stopped or unknown clusters
./unmanaged-cluster start <tab>

# Completion of running or unknown clusters
./unmanaged-cluster stop <tab>

# No file completion or any other completion for the following
./unmanaged-cluster start c2 <tab>
./unmanaged-cluster stop c1 <tab>
./unmanaged-cluster delete c1 <tab>
./unmanaged-cluster list <tab>
./unmanaged-cluster config <tab>
./unmanaged-cluster create <tab>
```

**New package**
I've created a new package `completion` and completely isolated the completion code in it.  This is the approach that was requested for `kubectl`.  For `helm` however, I chose to keep the completion code in the same file as the rest of the command (e.g., in configure.go for the configure command).  The advantage of keeping the completion with the rest of the code is that I find it reduces the risk of forgetting to update it when something changes.  What do we want for TCE?

**Completion descriptions**
Cobra allows to add descriptions to each completion.  It aims to help the user make a decision more easily.
Choosing a good description often requires some experience with the product so I would particularly like your input. 

I chose to add the status and provider, but it may not be useful, or if it is, we could revisit the output format.  I currently _don't_ like the format I chose:
zsh: `cluster1  -- status: Running - provider: kind`
bash: `cluster1       (status: Running - provider: kind)`

### DIFF
--- a/cli/cmd/plugin/unmanaged-cluster/completion/completion.go
+++ b/cli/cmd/plugin/unmanaged-cluster/completion/completion.go
@@ -1,0 +1,71 @@
+// Copyright 2022 VMware Tanzu Community Edition contributors. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package completion
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/cluster"
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/cmd"
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/config"
+	logger "github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/log"
+	"github.com/vmware-tanzu/community-edition/cli/cmd/plugin/unmanaged-cluster/tanzu"
+)
+
+const clusterAnyStatus = ""
+
+func Setup(rootCmd *cobra.Command) {
+	// The below commands accept the names of existing clusters
+	cmd.DeleteCmd.ValidArgsFunction = completeExistingClusterName(clusterAnyStatus)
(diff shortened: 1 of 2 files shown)


stmcginnis:
> The advantage of keeping the completion with the rest of the code is that I find it reduces the risk of forgetting to update it when something changes. What do we want for TCE?

I like the idea of keeping it with the code so there's less chance it will be overlooked. No strong feelings either way though, so I'm good with whatever the team decides.

Will check this out tomorrow. Thanks!

jpmcb:
> I like the idea of keeping it with the code so there's less chance it will be overlooked. No strong feelings either way though, so I'm good with whatever the team decides.

Also no strong feelings, but I think I'd also prefer it to stay next to the command initialization code. It'd be easy to just call `Setup()` without knowing what it's doing or what conditions it assumes are present.

PR #4878, opened 2022-06-23
UI plugin: small refactor of MC vSphere credentials step

swalner-vmware:


cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/FormInputSections/ClusterNameInput.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ChangeEvent } from 'react';
+import { CdsControlMessage, CdsFormGroup } from '@cds/react/forms';
+import { CdsInput } from '@cds/react/input';
+
+export function ClusterNameSection(
+    field: string,
```
mpanchajanya, line 6: Suggestion: - For Args instead of using the params if we use object params. It will be compile error if we misplace the arguments.  
Ex: Here we have errors and register as type any.  we could easily pass register , errors in args and compiler wont show error.  Instead we could have function ClusterNameSection({field, errors, register, onEnterClusterName})

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/vsphere/vsphere-mc-common/VsphereCredentialsStep.tsx
```diff
@@ -223,7 +223,7 @@ export function VsphereCredentialsStep(props: Partial<StepProps>) {
         } else {
             setLoadingDatacenters(false);
         }
-    }, [connected, clearDatacenters]);
+    }, [connected]);
```
raymondz1, line 226: it may cause a dependency issue if we removed clearDatacenters from the dependency list.

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/FormInputSections/ClusterNameInput.tsx
```diff
@@ -0,0 +1,24 @@
...
+    errors: any,
+    register: any,
+    onEnterClusterName: (evt: ChangeEvent<HTMLSelectElement>) => void
+) {
+    return (
+        <div cds-layout={`vertical gap:lg gap@md:lg col@sm:6 col:6`}>
```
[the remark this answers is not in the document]
swalner-vmware: Good point. I had originally been passing in a variable, and when I removed it, I should have changed it back

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/FormInputSections/ClusterNameInput.tsx
swalner-vmware: Also a good point. I think I should either do the object approach you use as an example, or correctly type the `errors` and `register` params

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/vsphere/vsphere-mc-common/VsphereCredentialsStep.tsx
swalner-vmware: Somehow having `clearDatacenters` in the list causes an infinite loop. My best guess is that's because clearDatacenters is actually called in the block, and calling it triggers the block again?

mpanchajanya (approved):

PR #4887, opened 2022-06-23
When publishing choco, checkout the code based on the release branch

davidvonthenen:

When publishing choco, we need to checkout the code based on the release branch and then publish with those choco files. 

NA

Unable to test workflow since it involves action/checkout. Simulated what the output branch should be.

NA


stmcginnis (approved):
Looks good to me!

PR #4892, opened 2022-06-24
new Resources page redesign

michow:
- updated resources page
- new list of resources data

Fixes: #


jkotzen:
I have reivewed these updates to the Resources page, and they look good - please merge!

docs/site/themes/template/layouts/partials/getting-started.html
```diff
@@ -8,7 +7,7 @@ <h2>Join us!</h2>
 		{{ else }}
 			<h2>Ready to dive in?</h2>
 			<p>Our documentation is a great place to start!</p>
-			<a href="/docs/" class="button navy-blue">Documentation</a>
+			<a href="/docs/{{ $.Site.Params.docs_latest }}" class="button navy-blue">Documentation</a>
```
stmcginnis, line 10: I don't think we want this change. This will cause docs from past releases to be redirected to the latest release.

docs/site/themes/template/layouts/partials/use-cases.html
```diff
@@ -23,7 +23,7 @@ <h3>Evaluate</h3>
                 </div>
             </div>
         </a>
-        <a href="/docs">
+        <a href="/docs/{{ $.Site.Params.docs_latest }}">
```
stmcginnis, line 26: Same here.

stmcginnis (changes requested):
This PR contains 62 commits. I don't think that was intentional?

You can try squashing or rebasing your changes, but to make sure you only have the changes you are trying to make, it might end up being easier to abandon this PR and create a new one, making sure you have an up to date fork of main first.

PR #4924, opened 2022-06-29
add fluent-bit 1.8.15 && 1.9.5

m1zzx2:

Fixes: #


addons/packages/fluent-bit/1.7.5/bundle/.imgpkg/bundle.yml
```diff
@@ -1,7 +1,7 @@
 apiVersion: imgpkg.carvel.dev/v1alpha1
 kind: Bundle
 metadata:
-  name: gatekeeper
+  name: fluent-bit
```
12345lcr, line 4: If anything changed in 1.7.5, the package should be rebuilt also. Or maybe it is better to correct this in a separate pr since this is not related to 1.8.15 and 1.9.5

PR #4939, opened 2022-06-30
Contour 1.20.2: Fix contour Deployment overlay error

sunjayBhatia:
Forgot to commit this fix with original package PR

Would originally replace the whole `containers` array rather than just the `args` list

N/A

Applied YAML to cluster to ensure it works!

N/A


skriss (commented):
fix LGTM, possible to add a unit test that cover this?

skriss (approved):

PR #4960, opened 2022-07-03
Skip sudo on install if running as root

### LINKED ISSUE #4957: install.sh fails if `sudo` is not in the `PATH`
In the install.sh file at line:84 (`sudo install "${MY_DIR}/tanzu" "${TANZU_BIN_PATH}" || handle_sudo_failure`). This line does not take ALLOW_INSTALL_AS_ROOT=true into consideration (upon inspecting I did not find implementation for handle_sudo_failure). As a result when logged in as root user the install.sh script fails to install tanzu cli.

TCE version: 0.12.1

Error:
=========
When logged in as root and executing ./install.sh under dir ~/tanzu/tce-linux-amd64-v0.12.1/ despite setting `ALLOW_INSTALL_AS_ROOT=true` the script yields error and fails to install tanzu on host machine. 

Error Message:
==============
```
Installing tanzu cli to /usr/local/bin/tanzu
./install.sh: line 84: sudo: command not found
./install.sh: line 84: handle_sudo_failure: command not found

./install.sh: line 111: tanzu: command not found
./install.sh: line 112: tanzu: command not found
./install.sh: line 114: tanzu: command not found
./install.sh: line 116: tanzu: command not found
./install.sh: line 118: tanzu: command not found
```

expected behavior:
==============
When export ALLOW_INSTALL_AS_ROOT=true the script to successfully install tanzu cli is expected. The script should not throw error due to not being able to use sudo command. A condition or an appropriate usage implementation of handle_suro_failure should tackle the issue.

Thanks

stmcginnis:

Our install script attempts to figure out the best place to install the
`tanzu` binary to. It will look in the user's home directory for a
`~/bin` path, and if that is not found it will try to install to the
common path of `/usr/local/bin`.

In the latter case, in normal cases this needs to be run using sudo to
have permissions to install to the system-wide path. But when running as
root, the use of sudo is a noop since there are no elevated permissions
needed.

On most systems, when running as root the `sudo` command is still
present somewhere in `$PATH`. But some systems do not have this
available. So despite the fact that the root user doesn't actually need
`sudo`, the command fails due to the command not being found.

To avoid this situation, this changes the install script to detect if we
are running as root, and if so, run the install to the system-wide path
without trying to invoke using sudo.

Fixes: #4957

### DIFF
--- a/hack/release/install.sh
+++ b/hack/release/install.sh
@@ -9,6 +9,11 @@ set -o pipefail
 set -o xtrace
 set +x
 
+SUDO="sudo"
+if [[ "$EUID" -eq 0 ]]; then
+    SUDO=""
+fi
+
 debug="false"
 if [[ $# -eq 1 ]] && [[ "$1" == "-d" ]]; then
     debug="true"
@@ -88,7 +93,7 @@ if [[ -n "${TANZU_BIN_PATH}" ]]; then
   if [[ -n "${TANZU_BIN_PATH}" ]]; then
     # best effort, so just ignore errors
     echo "Unable to delete Tanzu CLI. Retrying using sudo."
-    sudo rm -f "${TANZU_BIN_PATH}" > /dev/null
+    ${SUDO} rm -f "${TANZU_BIN_PATH}" > /dev/null
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


davidvonthenen (approved):

davidvonthenen (changes requested):
I think there is actually a modification that needs to be done for the uninstall script too if we want to be consistent

jpmcb (approved):
This looks good to me!

davidvonthenen (approved):

PR #5023, opened 2022-07-12
Add FluxCD kustomize controller v0.24.4+update.1

chaudharyalka:
Add FluxCD kustomize controller v0.24.4+update.1

Fixes: #[OLYMP-30817](https://jira.eng.vmware.com/browse/OLYMP-30817)

Manually added package

### DIFF
--- a/addons/packages/fluxcd-kustomize-controller/0.24.4/package.yaml
+++ b/addons/packages/fluxcd-kustomize-controller/0.24.4/package.yaml
@@ -11,7 +11,7 @@ spec:
     spec:
       fetch:
         - imgpkgBundle:
-            image: projects.registry.vmware.com/tce/fluxcd-kustomize-controller-bundle@sha256:424e08bc92768bbb179a8ba09c0062619dabcba1614736bbd69c025422bbd2f4
+            image: projects.registry.vmware.com/tce/fluxcd-kustomize-controller-bundle@sha256:288c5bae080a994dbb382b904d30f8323b54fe65be3c07837c2bdc8a90c92d2c
       template:
         - ytt:
             paths:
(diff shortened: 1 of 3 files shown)


seemiller (changes requested):
@chaudharyalka It is not necessary to create a new `0.24.4+update.1` directory. You can update the vendir yaml files and sync the new package contents directly into the existing `0.24.4` directory. That will keep the directory structure much cleaner.

chaudharyalka:
@seemiller As we discussed offline I have updated the PR. Please check.

addons/packages/fluxcd-kustomize-controller/0.24.4/package.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     spec:
       fetch:
         - imgpkgBundle:
-            image: projects.registry.vmware.com/tce/fluxcd-kustomize-controller-bundle@sha256:424e08bc92768bbb179a8ba09c0062619dabcba1614736bbd69c025422bbd2f4
+            image: projects.registry.vmware.com/tce/fluxcd-kustomize-controller-bundle:0.24.4-update.1
```
seemiller, line 14: @chaudharyalka You don't want to use a tag here. On an OCI Registry, the tag can potentially be changed to another image/bundle, so there is no guarantee of what you're going to actually be referring to. By using the sha256, you're guaranteed to get the correct image.

Can you switch this back to using the sha?

seemiller (changes requested):

addons/packages/fluxcd-kustomize-controller/0.24.4/package.yaml
chaudharyalka: @seemiller Done.

PR #5028, opened 2022-07-12
UI plugin: Introduces ConnectionNotification component (and small fit-n-finish tweaks)

swalner-vmware:


cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.scss
```diff
@@ -0,0 +1,3 @@
+.hide-me {
```
miclettej, line 1: This would be a good shared class to add to index.scss. We have a class in that file named 'hidden' which sets display to none. Maybe we should rename that class to no-display (or something similar) and then declare a class for no-visibility...could tweak naming but that's the general idea

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.tsx
```diff
@@ -0,0 +1,41 @@
+// React imports
+import React from 'react';
+// Library imports
+import { CdsAlert, CdsAlertGroup } from '@cds/react/alert';
+// App imports
+import './ConnectionNotification.scss';
+
+export function ConnectionNotification(
+    connected: boolean,
```
mpanchajanya, line 9: I feel like we could refactor this a bit. to be more dynamic and extensible.
I'm thinking instead of passing multiple params for connected, connecting,
we could define one config of type
interface IConnectionNotification {
status: 'connected' | 'connecting' | 'failed'
message: 'sample message',
}

and pass in the updated config object.  This could be big change. we could address it later point of time.

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/vsphere/vsphere-mc-common/VsphereCredentialsStep.tsx
```diff
@@ -183,26 +186,32 @@ export function VsphereCredentialsStep(props: Partial<StepProps>) {
...
-                setConnection(true);
-            },
-            (reason: any) => {
-                console.warn(`When trying to connect to the server, encountered error ${JSON.stringify(reason)}`);
-                let msg = 'Unable to connect to server! (See console for details.)';
-                if (reason?.status === SERVER_RESPONSE_BAD_CREDENTIALS) {
-                    msg = 'Incorrect username/password combination! Please try again. (See console for technical details.)';
+        // TODO: remove setTimeout(), which is just here to simulate a backend call delay
+        setTimeout(() => {
+            VsphereService.setVSphereEndpoint(vSphereCredentials).then(
```
mpanchajanya, line 199: Adding to my previous comment on having config. we could have a custom hook that returns the connection state. so the state is managed in custom hook and not in the implementation components.

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.scss
miclettej: Yes that would be better, thanks

raymondz1:
please add unit tests for the new component or create a github issue to track it.

mpanchajanya (approved):

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.tsx
```diff
@@ -0,0 +1,41 @@
...
+            )}
+            {!connected && errorMessage && (
+                <CdsAlertGroup status="danger">
+                    <CdsAlert>{errorMessage}</CdsAlert>
+                </CdsAlertGroup>
+            )}
+            {!connected && !connecting && !errorMessage && (
+                <div className="hide-me">
+                    <CdsAlertGroup status="neutral">
+                        <CdsAlert>&nbsp;</CdsAlert>
```
raymondz1, line 35: what is the space &nbsp; for?

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/vsphere/vsphere-mc-common/VsphereCredentialsStep.tsx
raymondz1: And please use async await to call the api

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.tsx
swalner-vmware: I agree with you in principle, and it definitely ran through my mind that the parameter list is awkward. Happy to discuss options, or leave it for a later refactor.

cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/ConnectionNotification/ConnectionNotification.tsx
swalner-vmware: The idea is to have an "empty" or hidden alert taking up the space so that the screen doesn't "jump" when the message displays. If there's a better way to do it, I'm happy to change it.

miclettej (approved):

PR #5052, opened 2022-07-14
[external-dns] Add default securityContext in deploy

christianang:

This PR sets a default limited privileged security context for external-dns package similar to what is shown here: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/security-context.md. We did not set the `runAsUser` field to 65534 as shown in the doc because we use the bitnami image which already sets a `USER` directive in the Dockerfile to `1001`, so it will already be a non-root user by default.

If a custom security context is provided to the external-dns package via data values it will override the entire security context for the deployment.

Fixes: n/a

Ran `make e2e-test`
Also validated the settings of the limited security context on a running cluster. This included checking capabilities, user running external-dns process, and the read-only setting on the root filesystem.


adobley (approved):
LGTM

PR #5076, opened 2022-07-16
Retrieve the image

ChiChiHeHeZhou:

Retrieve OS Image for Azure and AWS

image

image

Fixes: #


cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/management-cluster-settings-step/ManagementClusterSettings.tsx
```diff
@@ -77,6 +89,22 @@ function ManagementClusterSettings(props: Partial<MCSettings>) {
...
+    const handleMachineImageChange = (event: ChangeEvent<HTMLSelectElement>) => {
+        images.some((image) => {
+            if (image.name === event.target.value) {
+                setImageParameters(image);
+            }
+        });
+    };
+
+    useEffect(() => {
+        getImageMethod?.(awsState[STORE_SECTION_FORM].REGION).then((data) => {
```
miclettej, line 102: I think we should also catch the error state when executing this REST API call and display an alert to the user

PR #5104, opened 2022-07-21
Use tooltip

ChiChiHeHeZhou:

Previously, user should wrap the input text into `'<div>'` or `'<span>' `in the Component ToolContent in order to use Tooltip.
This PR is supposed to improve the functionality so user can wrap the text or just type in the pure text.
Also modify the location of the arrow in the tooltip

Fixes: #

### DIFF
--- a/cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/management-cluster-settings-step/ManagementClusterSettings.tsx
+++ b/cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/management-cluster-settings-step/ManagementClusterSettings.tsx
@@ -69,10 +69,7 @@ function ManagementClusterSettings<T>(props: Partial<MCSettings<T>>) {
     const [images, setImages] = useState<T[]>([]);
     const setImageParameters = (image) => {
         if (handleValueChange) {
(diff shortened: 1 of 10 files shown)


cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/management-cluster-settings-step/ManagementClusterSettings.tsx
```diff
@@ -53,20 +58,27 @@ const nodeProfiles = [
...
+        }
+    };
     const handleNodeProfileChange = (event: ChangeEvent<HTMLSelectElement>) => {
         setSelectedProfile(event.target.value);
         if (handleValueChange) {
             handleValueChange(INPUT_CHANGE, 'NODE_PROFILE', event.target.value, currentStep, errors);
         }
     };
-
+    const { awsState } = useContext(AwsStore);
```
miclettej, line 81: He, it looks like your changes to add the os image step got mixed into this PR. You may want to remove these changes to keep the tooltip PR clean.

ChiChiHeHeZhou: Yes sir. My bad. Gonna create a new PR then

PR #5143, opened 2022-07-26
Doc update for aws ebs csi driver

jeffwubj:
Doc update for aws ebs csi driver

Fixes: #


xiaoping8385:
I assume the installation would be automatic except upgrade scenario? and even for upgrade scenario, you may want to add a link for the detailed steps?

jeffwubj:
> 

hmm...  let me modify it to
```
The AWS EBS CSI Driver package is installed automatically during cluster creation.
```

as there are no upgrade description in other core package, and our tkg 1.6 doc is not published yet, I'd prefer to leave it not included, hope this makes sense.

xiaoping8385 (approved):

Issue #4771, opened 2022-06-13
UI Plugin: Fix Clarity CSS issues on Safari


miclettej:
## Feature Request

Several clarity form components do not render correctly on Safari. There are some issues with form fields visually overlapping each other.

Ticket owner should scan the UI in Safari to identify these areas and develop some common CSS overrides to resolve the issues

## Describe alternatives you've considered

none

## Additional context

none

PR #5163, opened 2022-07-29
UI Plugin: modify css difference between Safari and Chrome

ChiChiHeHeZhou:

For some input without setting `layout=vertical` , those input fields will be rendered differently between Safari and Chrome.
In Safari, those input fields will be `horizontal ` like below
image

image

Fixed those difference

Fixes: #4771


miclettej (approved):
very nice, thanks for finding the solution

PR #5154, opened 2022-07-27
Fix ako operator package YTT template error when specifying NSX-T T1 router path

XudongLiuHarold:
Signed-off-by: Xudong Liu <xudongl@vmware.com>

## What this PR does / why we need it

This PR fixes `ako-operator` package ytt template error when specifying NSX-T T1 router

## Which issue(s) this PR fixes

Fixes: #

## Describe testing done for PR

- Locally tested YTT template
- manually tested on e2e env

PR #5173, opened 2022-07-31
Fix ako-operator unit test and add assertions for non-empty avi_nsxt_…

HanFa:
…t1_lr.

Fix unit test so that the fix  https://github.com/vmware-tanzu/community-edition/pull/5154 can be validated.  

Fixes: #


XudongLiuHarold (approved):
The PR looks great, approved.
Just one more question: have we ever tested it locally?

PR #5185, opened 2022-08-02
UI plugin: Adds review step for vSphere MC

swalner-vmware:


miclettej (approved):
Looks great. Happy to help with the css and theming when if you'd like

PR #5191, opened 2022-08-03
UI plugin: Adds Azure MC Basic config review page

swalner-vmware:


cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/azure/azure-mc-basic/AzureBasicReview.config.tsx
```diff
@@ -0,0 +1,53 @@
...
+    ],
+};
+
+const configGroupClusterSettings: ConfigGroup = {
+    label: 'Cluster settings',
+    pairs: [
+        { label: 'Name', field: AZURE_FIELDS.CLUSTER_NAME },
+        { label: 'Node Type', field: AZURE_FIELDS.NODE_PROFILE },
+        {
+            label: 'Image (AMI)',
```
miclettej, line 25: For this label I think we should go with OS Image. AMI is an AWS-specific term. Azure doesn't seem to have a clear term and vsphere OS image could be vSphere OS Image Template if we wanted to be cute. I think simply OS Image is the safe term to use across providers. I should have thought through this on the OS image component label usage in previous PRs.

miclettej (approved):

PR #5209, opened 2022-08-08
Add avi controller version annotation to ako statefulset

XudongLiuHarold:

- Add avi controller version annotation to ako statefulset so that the …

Fixes: #


chenlin07 (approved):
lgtm

PR #5229, opened 2022-08-11
fix harbor using with pspNames

MinerYang:
Signed-off-by: yminer <yminer@vmware.com>

 we found a issue need to be fixed when we are using PSP which should switch origin Role/Role Binding into ClusterRole / ClusterRoleBinding instead. 

Fixes: #


addons/packages/harbor/2.5.3/bundle/config/globals.star
```diff
@@ -71,5 +71,9 @@ def get_resource_names_for_role(names):
   end
 end
 
+def setClusterRoleName():
```
zyyw, line 74: the implementation of this function doesn't actually `set` any value but return a value. Maybe it should be `getXXX()`?

MinerYang: I actually using get at the beginning... How about we use `assembleClusterRoleName` or `concatClusterRoleName`.
@heww

addons/packages/harbor/2.5.3/bundle/config/overlay/add-rbac.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - podsecuritypolicies
+  verbs:
+  - use
+
+#@ if/end values.pspNames:
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: #@ setClusterRoleName()
```
heww, line 25: `setClusterRoleName` => getClusterRoleBindingName

addons/packages/harbor/2.5.3/bundle/config/globals.star
heww, line 74: setClusterRoleName => getClusterRoleName

addons/packages/harbor/2.5.3/bundle/config/overlay/add-rbac.yaml
heww: We can add two functions in the globals.star.

One is getClusterRoleName, another is getClusterRoleBindingName

MinerYang: I am using this function to concat both `ClusterRole`  name and `ClusterRoleBinding` name

heww (approved):

zyyw (approved):

PR #5243, opened 2022-08-16
Clear previous credentials

ChiChiHeHeZhou:

**Bug Description**
**1.**
Assume user chooses to use _One-time credential_ and input some value already as following:

image

**2.**
Then user switch to _AWS credential profile_ , and input some value.

image

**But the previous input info from _One-time credential_ will still exist.**

This PR fix this issue.
1.Add reset to clean the previous input info
2.Remove the `awsDispatch` while change input. As substituion, call `awsDispatch` when click _**Next**_ button.
3.Resolved minor issue from previous PR which is changing all field names to AWS_FIELDS element.

Fixes: #

### DIFF
--- a/cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/aws-mc-common/management-credential-step/ManagementCredentials.tsx
+++ b/cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/aws-mc-common/management-credential-step/ManagementCredentials.tsx
@@ -54,7 +54,7 @@ function ManagementCredentials(props: Partial<StepProps>) {
             [AWS_FIELDS.ACCESS_KEY_ID]: '',
             [AWS_FIELDS.SESSION_TOKEN]: '',
             [AWS_FIELDS.PROFILE]: '',
-            [AWS_FIELDS.REGION]: '',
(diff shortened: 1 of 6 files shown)


cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/aws-mc-common/management-credential-step/ManagementCredentials.tsx
```diff
@@ -95,6 +103,16 @@ function ManagementCredentials(props: Partial<StepProps>) {
     };
 
     const onSubmit: SubmitHandler<FormInputs> = (data) => {
+        const valueList = getValues();
+        for (const i in valueList) {
+            console.log(i);
```
miclettej, line 108: Sorry just noticed this. Since these are account credentials, we should probably not log them to the browser console.

ChiChiHeHeZhou: Yes sir. My bad. This is supposed to be printed for test. Removing them. 😄

miclettej (approved):

PR #5273, opened 2022-08-22
UI plugin refactors:

swalner-vmware:
* DefaultOrchestrator.initResources uses generics
* fxnOnSelectArrayObject returns an event-handling function useful for selecting array objects
* objects are put in the store (rather than just the key value)
* putting data in the store is preferred to local storage in a component, which means onSubmit does not need tos tore values
* Array utils: first, last, middle


cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/aws-mc-common/aws-orchestrator/AwsOrchestrator.service.tsx
```diff
@@ -46,29 +40,26 @@ export const nodeProfiles: NodeProfileType[] = [
...
+            resourceName: AWS_FIELDS.EC2_KEY_PAIR,
+            dispatch: awsDispatch,
+            errorObject,
+            setErrorObject,
+            fetcher: () => AwsService.getAwsKeyPairs(),
+            fxnSelectDefault: AwsDefaults.selectDefaultEC2KeyPairs,
+        });
     }
 
     static async initNodeProfile(props: AwsOrchestratorProps) {
```
ChiChiHeHeZhou, line 65: Do we want to use `DefaultOrchestrator` for `initNodeProfile` ?

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/AwsClusterSettingsStep.tsx
```diff
@@ -78,28 +78,27 @@ function AwsClusterSettingsStep(props: Partial<StepProps>) {
...
+    // NOTE: we assume that the osImages were set in the store during the credentials step
+    const osImages = getResource<AWSVirtualMachine[]>(AWS_FIELDS.OS_IMAGE, awsState) || [];
+    const initialSelectedNodeProfileId = awsState[STORE_SECTION_FORM][AWS_FIELDS.NODE_PROFILE] || nodeProfiles[0].id;
+    const [selectedNodeProfileId, setSelectedNodeProfileId] = useState(initialSelectedNodeProfileId);
 
-    if (!initialSelectedNodeProfileId) {
-        initialSelectedNodeProfileId = nodeProfiles[0].id;
-        setValue(AWS_FIELDS.NODE_PROFILE, initialSelectedNodeProfileId);
-    }
+    const onNodeProfileChange = (instanceType: string) => {
```
miclettej, line 86: I think this is one of the areas where we are using 'instanceType' but the value is actually the node profile type

cli/cmd/plugin/ui/web/tanzu-ui/src/views/management-cluster/aws/aws-mc-common/aws-orchestrator/AwsOrchestrator.service.tsx
swalner-vmware: I think AWS will have a different strategy than vSphere, so I'd suggest we leave it here for now.

miclettej (approved):

PR #5278, opened 2022-08-23
Enables ExternalDNS observability via pod labels

### LINKED ISSUE #5277: Allow configuration for external-dns to publish metrics
## Feature Request

External-dns can expose metrics.  However, in order to configure it to do so, you must set a deployment arg specifying the metrics address, and then add labels to the deployment pods to set appropriate prometheus metrics.  The current implementation of the values schema provides ability to set the deployment args, but you can not set the deployment pod labels.

```yaml
deployment:
  args:
    - --metrics-address=:7979 #! ensures external dns exposes metrics
```

I have created an overlay to apply the deployment pod labels, but would like to be able to configure observability directly using the values file.

```yaml
#@ load("@ytt:overlay", "overlay")
#@overlay/match by=overlay.subset({"metadata":{"name":"external-dns"}, "kind": "Deployment"})
---
spec:
  template:
    #@overlay/match missing_ok=True
    metadata:
      #@overlay/match missing_ok=True
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/path: "/metrics"
        prometheus.io/port: "7979"
```

doddatpivotal:
Observability is key to day 2 operations. 

Fixes: #5277

Updated and executed unit tests for ExternalDNS

### DIFF
--- a/addons/packages/external-dns/0.11.0/package.yaml
+++ b/addons/packages/external-dns/0.11.0/package.yaml
@@ -106,7 +106,7 @@ spec:
                 nullable: true
                 default: null
               default: null
-            podVolumes:
+            podLabels:
               nullable: true
               description: 'Labels to be added to all deployment pods'
               default: null
--- a/addons/packages/external-dns/0.11.0/package.yaml
+++ b/addons/packages/external-dns/0.11.0/package.yaml
@@ -106,6 +106,10 @@ spec:
                 nullable: true
                 default: null
               default: null
+            podVolumes:
+              nullable: true
+              description: 'Labels to be added to all deployment pods'
+              default: null
         serviceaccount:
           type: object
           additionalProperties: false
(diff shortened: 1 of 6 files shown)


addons/packages/external-dns/0.11.0/package.yaml
```diff
@@ -106,6 +106,10 @@ spec:
                 nullable: true
                 default: null
               default: null
+            podVolumes:
```
tylerschultz, line 109: This needs to be changed to podLabels?

tylerschultz (commented):
Thanks for your contribution. We'll merge once the package yaml is fixed up.

addons/packages/external-dns/0.11.0/package.yaml
doddatpivotal: Great catch.

tylerschultz (approved):

PR #5289, opened 2022-08-24
fixed wizard tab issue

raymondz1:

fixed the issue: the user can't go back to the previous step if the current step is invalid.

Fixes: #


cli/cmd/plugin/ui/web/tanzu-ui/src/shared/components/wizard/StepNav.tsx
```diff
@@ -28,9 +27,12 @@ function StepNav(props: ChildProps | any) {
                         key={index}
                         style={{ width: `${100 / totalSteps}%` }}
                         onClick={() => {
-                            if (!hasInvalidStep) {
```
swalner-vmware, line 31: nit: to make the code more readable to a new developer, what do you think of

1. creating a method `existsPreviousInvalidStep(index)` and 
2. using lines 30-34  in the implementation
3. using the original logic but instead of `!hasInvalidStep` use `!existsPreviousInvalidStep(index)`

swalner-vmware (approved):

miclettej (approved):

PR #5342, opened 2022-09-06
UI Plugin: Set vsphere useThumbprint to true by default

miclettej:

Update vsphere credentials step to set useThumbprint to true if value is undefined

Fixes: #


raymondz1 (approved):
LGTM

PR #5401, opened 2022-09-30
Add kube vip cloud provider package

lubronzhan:

Tanzu is considering supporting kube-vip as loadbalancer type service. To do that, it needs the kube-vip-cloud-provider

Fixes: #


HanFa (approved):
lgtm

PR #5412, opened 2022-10-04
Contour: Envoy service overlay allows nodeport configuration w/ LoadBalancer Service

sunjayBhatia:
## What this PR does / why we need it
Allow NodePort configuration of the Envoy Service when using LoadBalancer type as well as NodePort type

## Which issue(s) this PR fixes
N/A

## Describe testing done for PR
- unit testing
- applied generated YAML to cluster w/ configuratio exercised

## Special notes for your reviewer
N/A

PR #5504, opened 2022-10-31
allow envoy to be run as deployment

KauzClay:
This change allows users to configure Contour so that they can deploy Envoy as a DaemonSet OR a Deployment, and choose the number of replicas when Deployment is used.

Fixes: # N/A

* unit tests
* I've deployed Contour package with these changes and have seen Envoy successfully run as a Deployment.

### DIFF
--- a/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
+++ b/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
@@ -4,7 +4,19 @@
 
 #@overlay/match by=overlay.subset({"kind": "DaemonSet", "metadata": {"name": "envoy"}})
 ---
+kind: #@ data.values.envoy.workload.type
 spec:
+  #@ if data.values.envoy.workload.type == "Deployment":
+  #@overlay/match missing_ok=True
+  replicas: #@ data.values.envoy.workload.replicas
+  #@overlay/remove
+  updateStrategy:
+  #@overlay/match missing_ok=True
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
+  #@ end
   template:
     spec:
       containers:
--- a/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
+++ b/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
@@ -15,7 +15,9 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 10%
+      #! This value of maxSurge means that during a rolling update
+      #! the new ReplicaSet will be created first.
+      maxSurge: 10%
   #@ end
   template:
     spec:
--- a/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
+++ b/addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
@@ -19,6 +19,19 @@ spec:
   #@ end
   template:
     spec:
+      #@ if data.values.envoy.workload.type == "Deployment":
+      #@overlay/match missing_ok=True
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                      - envoy
+              topologyKey: "kubernetes.io/hostname"
+      #@ end
       containers:
       #@overlay/match by=overlay.subset({"name": "envoy"})
       - args:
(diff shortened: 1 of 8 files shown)


addons/packages/contour/1.22.0/test/unittest/contour_test.go
```diff
@@ -597,6 +599,61 @@ var _ = Describe("Contour Ytt Templates", func() {
...
+				},
+				TopologyKey: "kubernetes.io/hostname",
+			}
+			Expect(deployment.Spec.Template.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution).To(ContainElement(affinity))
+
+		})
+
+		Context("Envoy replicas specified", func() {
+			BeforeEach(func() {
+				values = envoyServiceAnnotationsAWSNLB
```
sunjayBhatia, line 640: can remove this line!

sunjayBhatia (commented):
changes to schema/overlays/tests lgtm, looks like you've captured the relevant differences between the two kinds

next steps I think are to update the README for this version with the new config fields and package.yaml by running (from the repo root):
```
make generate-openapischema-package PACKAGE=contour VERSION=1.22.0
```

and then if you have permissions:
```
make push-package PACKAGE=contour VERSION=1.22.0 TAG=1.22.0_build.N
```
(where N is a build number, probably 3 given https://github.com/vmware-tanzu/community-edition/pull/5412#discussion_r987196102, but the push will fail if you choose something that exists, just increment from there!)

and updating the package bundle sha like [here](https://github.com/vmware-tanzu/community-edition/pull/5412/files/a5798d3c1eca459924de6c2a05af10ac3fdecd7b#diff-351d0143093a2d3dc396480fbae617cf7626a2bc1f55ba9ad7e29b7aa00a3ddc)

if you don't have permission to push the package, let me know, we can push it for you (youll need to do `docker login projects.registry.vmware.com`

sunjayBhatia:
cc @skriss in case i missed anything in the overlay review

addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
```diff
@@ -4,9 +4,34 @@
...
+  #@ if data.values.envoy.workload.type == "Deployment":
+  #@overlay/match missing_ok=True
+  replicas: #@ data.values.envoy.workload.replicas
+  #@overlay/remove
+  updateStrategy:
+  #@overlay/match missing_ok=True
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
```
skriss, line 18: Our example Deployment uses maxSurge here instead of maxUnavailable

addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
```diff
@@ -4,9 +4,34 @@
 
 #@overlay/match by=overlay.subset({"kind": "DaemonSet", "metadata": {"name": "envoy"}})
 ---
+kind: #@ data.values.envoy.workload.type
```
skriss, line 7: Would be good to add some validation on this field in https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/contour/1.22.0/bundle/config/contour.star#L117, to only allow `Deployment` or `DaemonSet`. The overlay looks safe but still better UX to explicitly validate rather than ignore invalid values.

skriss (changes requested):
Thanks for doing this @KauzClay! Just a couple small comments

addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
KauzClay: ah good point, I added some now. What do you think?

addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
KauzClay: noted, just changed it

addons/packages/contour/1.22.0/bundle/config/overlays/update-envoy-daemonset.yaml
skriss: :+1: lgtm

KauzClay:
Hey @sunjayBhatia , I updated the package yaml with that make command. Unfortunately I don't have permissions to push `make push-package PACKAGE=contour VERSION=1.22.0 TAG=1.22.0_build.3`

Can you please do that?

sunjayBhatia:
> Hey @sunjayBhatia , I updated the package yaml with that make command. Unfortunately I don't have permissions to push `make push-package PACKAGE=contour VERSION=1.22.0 TAG=1.22.0_build.3`
> 
> Can you please do that?

checked out your branch and pushed the package, sha: `b68ad8ec3012db7d2a2e84f8544685012e2dca09d28d54dce8735fb60f0d05bf1`

sunjayBhatia:
I think you might need to modify the unit tests for the `maxSurge` change as well

sunjayBhatia (approved):

skriss (approved):

PR #5529, opened 2022-11-11
Remove the default empty value from configMap of kube-vip-cloud-provider

lubronzhan:

kube-vip-cloud-provider can't handle empty string

Fixes: #


HanFa (approved):
lgtm

PR #5645, opened 2023-02-03
WIP: enable user to configure net permissions

nikhilbarge:

Fixes: #


addons/packages/vsphere-csi/2.6.2/bundle/config/values.yaml
```diff
@@ -32,3 +32,25 @@ vsphereCSI:
   no_proxy: ""
   deployment_replicas: 3
   windows_support: false
+  netpermissions:
+    netpermission_A:
+      ip: "*"
+      permissions: READ_WRITE
+      rootsquash: false
+    netpermission_B:
+      ip: "10.20.20.0/24"
```
divyenpatel, line 41: we should not be putting the default values for net permissions in the package.
I think we should have default no net permission and if the users want they can fill up these values.

PR #5652, opened 2023-02-09
update image for vsphere-csi

nikhilbarge:
Enable TKGm users to configure the NetPermission when create the workload clusters with vSphere CSI NFS volumes.

Build new Package image and PackageRepository.
verified after installing package that netpermissions are configured.

Enable TKGm users to configure the NetPermission when create the workload clusters with vSphere CSI NFS volumes.


chethanv28 (approved):
/approve
/lgtm

